PR #1, opened 2013-10-19
Basic compile time fixes for mac builds

pdehne:
These changes are needed to make beast_core compile on the mac


vinniefalco:
Library is completely different now

PR #5, opened 2016-04-28
Port cmake scripts to linux

seelabs:
Modified root CMakeLists.txt to find the boost libraries. Every executable will be linked against the boost libs. I also downgraded the required CMake to 3.2 (that's what's on my system).

CXX_FLAGS should be the same as those in Jamroot.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Part of Beast
 
-cmake_minimum_required (VERSION 3.5)
+cmake_minimum_required (VERSION 3.2)
 
 project (Beast)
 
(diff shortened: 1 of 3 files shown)


seelabs:
Rebased against master. I had to make some changes to support the examples directory. In particular:

1) I had to use `target_link_libraries` for each executable instead of one command in the root CMakeLists.txt

2) Each executable links in the superset of boost libraries needed by all the exes, even if it doesn't need it (filesystem, for example).

I'd like to resolve these issues in a future PR

vinniefalco:
Merged

PR #6, opened 2016-05-02
http::prepare and more

vinniefalco:
A lot of refactoring, tests, docs, and fine tuning of the interfaces.

Documentation preview: http://vinniefalco.github.io/stage/beast/


examples/file_body.hpp
```diff
@@ -79,7 +79,11 @@ struct file_body
         boost::tribool
         operator()(resume_context&&, error_code&, Write&& write)
         {
-            buf_len_ = std::min(size_ - offset_, sizeof(buf_));
+            if(size_ - offset_ < sizeof(buf_))
+                buf_len_ = static_cast<std::size_t>(
+                    size_ - offset_);
+            else
+                buf_len_ = sizeof(buf_);
```
ximinez: Would `buf_len_ = static_cast<std::size_t>(std::min(size_ - offset_, sizeof(buf_)));` be more efficient, or does upcasting `sizeof(buf_)` only to downcast it back make it not worth it?

vinniefalco: The answer is revealed when you try to make the code work for both 32-bit and 64-bit architectures :-)

include/beast/http/impl/message.ipp
```diff
@@ -310,9 +195,176 @@ is_upgrade(message<isRequest, Body, Headers> const& msg)
...
+    message<isRequest, Body, Headers>& msg)
+{
+    if(msg.version >= 11)
+    {
+        if(! msg.headers.exists("Content-Length") &&
+            ! rfc2616::token_in_list(
+                msg.headers["Transfer-Encoding"], "chunked"))
+            if(! rfc2616::token_in_list(
+                    msg.headers["Connection"], "close"))
+                msg.headers.insert("Connection", "close");
```
ximinez: Formatting aside, this is functionally equivalent to
```
       if(! msg.headers.exists("Content-Length") &&
            ! rfc2616::token_in_list(
                msg.headers["Transfer-Encoding"], "chunked") &&
                    ! rfc2616::token_in_list(
                        msg.headers["Connection"], "close"))
            msg.headers.insert("Connection", "close");
```
Yes?

examples/file_body.hpp
```diff
@@ -36,8 +36,8 @@ struct file_body
 
     class writer
     {
-        std::size_t size_;
-        std::size_t offset_ = 0;
+        std::uint64_t size_;
+        std::uint64_t offset_ = 0;
         std::string const& path_;
         FILE* file_ = nullptr;
         char buf_[4096];
```
HowardHinnant: From `~writer()` it looks like you intend for `writer` to be non-copyable.

include/beast/http/impl/message.ipp
vinniefalco: I think so

examples/file_body.hpp
vinniefalco: I hadn't given it any thought. The implementation never makes copies or tries to move a writer, its an ephemeral object.

examples/http_crawl.cpp
```diff
@@ -46,10 +46,11 @@ int main(int, char const*[])
...
             auto ep = hs.lowest_layer().remote_endpoint();
-            request<empty_body> req({method_t::http_get, "/", 11});
+            request<empty_body> req({"GET", "/", 11});
             req.headers.insert("Host", host +
                 std::string(":") + std::to_string(ep.port()));
             req.headers.insert("User-Agent", "beast/http");
+            prepare(req);
             hs.write(req);
             response<string_body> resp;
             hs.read(resp);
```
HowardHinnant: I got this example to run.  I have no idea what it is trying to do, but it sure spent quite a while trying to do it!

examples/file_body.hpp
HowardHinnant: If your customers copy this example, and then copy `writer` in their code.  You either want it to work, or not compile.

examples/http_example.cpp
```diff
@@ -23,9 +23,10 @@ int main()
...
 
     // Send HTTP request using beast
-    request<empty_body> req({method_t::http_get, "/", 11});
+    request<empty_body> req({"GET", "/", 11});
     req.headers.replace("Host", host + ":" + std::to_string(sock.remote_endpoint().port()));
     req.headers.replace("User-Agent", "Beast");
+    prepare(req);
     write(sock, req);
 
     // Receive and print HTTP response using beast
```
HowardHinnant: Nice clear simple example.  Well commented!

examples/http_crawl.cpp
vinniefalco: It fetches the home page of the 10,000 top visited domains ranked by Alexa.

examples/http_example.cpp
vinniefalco: Heh thanks! The example program was your idea

examples/file_body.hpp
ximinez: Aha. Of course, I've gotten so used to 64-bit arch, that I often forget to consider 32-bit. I need to stop doing that.

include/beast/detail/write_streambuf.hpp
```diff
@@ -133,11 +133,12 @@ write_streambuf(Streambuf& streambuf, T const& t)
...
+write_streambuf(Streambuf& streambuf,
+    T0 const& t0, T1 const& t1, TN const&... tn)
 {
-    write_streambuf(streambuf, std::forward<T0>(t0));
-    write_streambuf(streambuf, std::forward<T1>(t1));
-    write_streambuf(streambuf, std::forward<TN>(tn)...);
+    write_streambuf(streambuf, t0);
+    write_streambuf(streambuf, t1);
+    write_streambuf(streambuf, tn...);
 }
```
HowardHinnant: This is very optional.  Actually in the obsessive polishing category...  

You could eliminate this overload: `write_streambuf(Streambuf&)` by changing the above code from:
````
    write_streambuf(streambuf, t0);
    write_streambuf(streambuf, t1);
    write_streambuf(streambuf, tn...);
````
to
````
    write_streambuf(streambuf, t0);
    write_streambuf(streambuf, t1, tn...);
````

File: include/beast/websocket/detail/endian.hpp [hunk 7]
```diff
@@ -53,6 +53,17 @@ little_uint32_to_native(void const* buf)
...
+inline
+void
+native_to_little_uint32(std::uint32_t v, void* buf)
+{
+    auto p = reinterpret_cast<std::uint8_t*>(buf);
+    p[0] =  v        & 0xff;
+    p[1] = (v >>  8) & 0xff;
+    p[2] = (v >> 16) & 0xff;
+    p[3] = (v >> 24) & 0xff;
+}
```
ximinez: Is this assuming a particular native data layout? Is that cross-platform safe?

vinniefalco: it *should* work on all platforms - please verify!

HowardHinnant: This looks right (and portable) to me.

ximinez: Sounds good. I wasn't sure, so I figured it would be better to double-check.

HowardHinnant:
👍 Looks good to me.

include/beast/static_string.hpp
```diff
@@ -0,0 +1,696 @@
...
+template<std::size_t M>
+auto
+static_string<N, CharT, Traits>::
+operator=(const CharT (&s)[M]) ->
+    static_string&
+{
+    static_assert(M-1 <= N,
+        "static_string overflow");
+    n_ = M-1;
+    std::copy(&s[0], &s[M], &s_[0]);
```
ximinez: You use `Traits::copy(&s_[0], &s[0], M);` everywhere in the file except here. Is there a reason, or should this one be changed for consistency.

include/beast/static_string.hpp
```diff
@@ -0,0 +1,696 @@
...
+    }
+    else if(M-1 > rhs.size())
+    {
+        auto const v = Traits::compare(
+            &s[0], rhs.data(), rhs.size());
+        if(v == 0)
+            return 1;
+        return v;
+    }
+    return Traits::compare(&s[0], rhs.data(), M-1);
```
ximinez: Silly question: why not implement this as `return -compare(rhs, s);` to call the other override, and reduce code duplication (particularly if you declare this override as `inline`)?

include/beast/websocket/impl/stream.ipp
```diff
@@ -437,15 +529,16 @@ read_frame(frame_info& fi, Streambuf& streambuf, error_code& ec)
...
+                    if(error_)
                         return;
                     continue;
                 }
                 else if(rd_fh_.op == opcode::pong)
                 {
                     ping_payload_type data;
                     detail::read(data, fb.data(), code);
-                    if((error_ = ec != 0))
+                    if(code != close_code::none)
```
ximinez: Oooh. Close call there.

include/beast/websocket/impl/stream.ipp
```diff
@@ -170,18 +170,29 @@ template<class NextLayer>
...
-        "Stream requirements not met");
+}
+
+template<class NextLayer>
+void
+stream<NextLayer>::
+accept()
+{
+    static_assert(is_SyncStream<next_layer_type>::value,
+        "SyncStream requirements not met");
```
ximinez: Are the `static_assert`s redundant in the overloads that throw, since they call overloads that don't throw, which have duplicate checks? I suppose it doesn't hurt since they are static after all, but I figured it's worth asking.

test/static_string.cpp
```diff
@@ -0,0 +1,191 @@
...
+            expect(s2.front() == 'x');
+            expect(s2.back() == 'x');
+            str1 const s3(s2);
+            expect(s3 == "x");
+            expect(s3[0] == 'x');
+            expect(s3.at(0) == 'x');
+            expect(s3.front() == 'x');
+            expect(s3.back() == 'x');
+            s2 = "y";
+            expect(s2 == "y");
```
ximinez: Test that values were copied, and not pointers:
```expect(s3 == "x");```

File: test/static_string.cpp [hunk 13]
```diff
@@ -0,0 +1,191 @@
...
+        {
+            str1 s1("x");
+            str2 s2("x");
+            expect(s1 == s2);
+            expect(s1 <= s2);
+            expect(s1 >= s2);
+            expect(! (s1 < s2));
+            expect(! (s1 > s2));
+            expect(! (s1 != s2));
+        }
```
ximinez: {
            str1 s1("x");
            str2 s2("y");
            expect(! (s1 == s2));
            expect(s1 <= s2);
            expect(! (s1 >= s2));
            expect(s1 < s2);
            expect(! (s1 > s2));
            expect(s1 != s2);
        }

ximinez: Good equality tests there, but need some inequality / ordering tests, too.

test/static_string.cpp
```diff
@@ -0,0 +1,191 @@
...
+            expect(! (s < "x"));
+            expect(! (s > "x"));
+            expect(! (s != "x"));
+            expect("x" == s);
+            expect("x" <= s);
+            expect("x" >= s);
+            expect(! ("x" < s));
+            expect(! ("x" > s));
+            expect(! ("x" != s));
+        }
```
ximinez: Same here:
```         {
            str2 s("x");
            expect(! (s == "y"));
            expect(s <= "y");
            expect(! (s >= "y"));
            expect(s < "x");
            expect(! (s > "x"));
            expect(s != "x");
            expect(! ("y" == s));
            expect(! ("y" <= s));
            expect("y" >= s);
            expect(! ("y" < s));
            expect("y" > s);
            expect("y" != s);
        }
```

test/streambuf_readstream.cpp
```diff
@@ -7,3 +7,40 @@
...
+        boost::asio::io_service ios;
+        {
+            streambuf_readstream<socket_type, streambuf> srs(ios);
+            streambuf_readstream<socket_type, streambuf> srs2(std::move(srs));
+            srs = std::move(srs2);
+        }
+        {
+            socket_type sock(ios);
+            streambuf_readstream<socket_type&, streambuf> srs(sock);
+            streambuf_readstream<socket_type&, streambuf> srs2(std::move(srs));
```
ximinez: Move it back here, too? `srs = std::move(srs2);`

.travis.yml
```diff
@@ -66,9 +69,9 @@ matrix:
       env: GCC_VER=5 VARIANT=release ADDRESS_MODEL=64
       addons: *ao_gcc5
 
-    # - compiler: gcc
-    #   env: GCC_VER=5 VARIANT=release ADDRESS_MODEL=32
-    #   addons: *ao_gcc5
+    # # - compiler: gcc
+    # #   env: GCC_VER=5 VARIANT=release ADDRESS_MODEL=32
+    # #   addons: *ao_gcc5
```
ximinez: _Really_ commented out!

include/beast/static_string.hpp
vinniefalco: That's a bug. It should use `Traits::copy`

include/beast/static_string.hpp
vinniefalco: That's a good idea ^_^

If you're really asking me *why* it was done this way, probably because I was trying to write it as quickly as possible.

test/streambuf_readstream.cpp
vinniefalco: Think about what you're saying :-) study the declaration.

include/beast/websocket/impl/stream.ipp
vinniefalco: Yeah whoops

include/beast/websocket/impl/stream.ipp
vinniefalco: Great question. They are redundant. The thinking is that if they go off, the user will get an error message that is very close in the source file to the signature of the function they are trying to call, instead of a function with a different signature.

include/beast/static_string.hpp
vinniefalco: This will be addressed in a subsequent PR that also increases coverage

test/static_string.cpp
vinniefalco: Will fix, thanks for the code paste

test/static_string.cpp
vinniefalco: This fails:
```
expect(s < "x");
```

include/beast/http/impl/write.ipp
```diff
@@ -29,6 +30,114 @@ namespace http {
...
+        write(streambuf, " HTTP/1.0\r\n");
+        break;
+    case 11:
+        write(streambuf, " HTTP/1.1\r\n");
+        break;
+    default:
+        write(streambuf, " HTTP/");
+        write(streambuf, msg.version / 10);
+        write(streambuf, ".");
+        write(streambuf, msg.version % 10);
```
ximinez: Bad things will happen here if HTTP ever gets to a 3 digit (`1.25`) or subdivided (`1.1.5`) version number.

vinniefalco: As HTTP/2 is already out, I have confidence that we are okay. Anyway, I"m going to refactor this code because its not getting coverage (and likely never will, because we will probably never see anything new below HTTP/2 in our lifetime).

doc/http.qbk
```diff
@@ -25,20 +25,43 @@ compliant with the Hypertext Transfer Protocol and the supplements that
...
+
+
+[section:scope Scope]
+
+This library is designed to be a building block for creating higher level
+libraries. It is not designed to be end-user facing. There is no convenient
+class that implements the core of a web server, nor is there a convenient
+class to quickly perform common operations such as fetching a file or
+connecting and retrieving a document from a secure connection. These
+use-cases are important. But this library does not try to do that. Instead,
```
ximinez: IMHO, this works better as one sentence. "These use-cases are important, but this library does not try to do that."

include/beast/http/write.hpp
```diff
@@ -18,25 +18,59 @@
...
+
+    @li An error occurs.
+
+    This operation is implemented in terms of one or more calls
+    to the stream's `write_some` function.
+
+    The implementation will automatically perform chunk encoding if
+    the contents of the message indicate that chunk encoding is required.
+    If the semantics of the message indicate that the connection should
+    be closed after the message is sent, the error returns from this
```
ximinez: "the error *thrown* from this function".

include/beast/http/write.hpp
```diff
@@ -18,25 +18,59 @@
...
```
ximinez: "the error return*ed* from this function"

ximinez: Moving this comment to the correct line: "the error return*ed* from this function"

doc/http.qbk
vinniefalco: Thanks for the edits. They will go in the next PR

include/beast/websocket/stream.hpp
```diff
@@ -527,23 +598,35 @@ class stream : public detail::stream_base
...
+        @li An error occurs on the stream.
+
+        This operation is implemented in terms of one or more calls to the
+        next layer's `async_write_some` functions, and is known as a
+        <em>composed operation</em>. The program must ensure that the
+        stream performs no other operations until this operation completes.
+
+        If the passed HTTP request is a valid HTTP WebSocket Upgrade
+        request, a HTTP response is sent back indicating a successful
+        upgrade. When this asynchronous operaiton completes, the stream is
```
ximinez: Typo: operation

test/http/body_writer.cpp
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright (c) 2013-2016 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+// Test that header file is self-contained.
+#include <beast/http/body_type.hpp>
```
ximinez: This file can be deleted. It is a duplicate of `test/http/body_type.cpp`, but does not appear to be referenced anywhere.

include/beast/websocket/impl/stream.ipp
ximinez: Good thinking.

test/http/body_writer.cpp
vinniefalco: Yep, will be removed in b4.

test/static_string.cpp
ximinez: Copy/paste error. All the comparisons should be against `"y"`.

```
         {
            str2 s("x");
            expect(! (s == "y"));
            expect(s <= "y");
            expect(! (s >= "y"));
            expect(s < "y");
            expect(! (s > "y"));
            expect(s != "y");
            expect(! ("y" == s));
            expect(! ("y" <= s));
            expect("y" >= s);
            expect(! ("y" < s));
            expect("y" > s);
            expect("y" != s);
        }
```

test/streambuf_readstream.cpp
ximinez: Ah. I was pattern matching with the previous block, and missed that this declared with a reference.

include/beast/http/impl/write.ipp
ximinez: That's fair.

PR #8, opened 2016-05-09
[WIP] Travis CI updates:

sublimator:
* Use clang 3.8
* Add msan/usan variants
* Build only gcc.coverage and clang.{asan,usan,msan}
* Make sure (correct) llvm-symbolizer is on PATH


vinniefalco:
I don't understand what I'm looking at here. Are your changes in the commit entitled "Travis CI updates?" If so, has that commit changed from the branch you cloned? If the answer is yes, why is the version commit still marked "Verified?" Rebasing would invalidate the signature. I'm not sure what's being submitted here. If you changed the .travis.yml file it would be best if the change was in a single commit marked "[FOLD]...".

sublimator:
It's [WIP] and based off your 'next' branch, which has some things disabled
for speed (autobahn, valgrind)

sublimator:
Wah, I see. I answered from mobile before. I dunno what happened here, and why it's showing all those commits of yours.

vinniefalco:
I think I have it sorted, I believe this is the individual commit. Did you intend to keep the number of items in the matrix down to 4 for speed (it seems like a good idea)?
https://github.com/vinniefalco/Beast/commits/niq-travis

sublimator:
Yes, per your idea of only building the full matrix for non feature
branches. Not sure best way to do that yet.

vinniefalco:
This branch has your change:
https://github.com/vinniefalco/Beast/commits/dstream

I turned off memory sanitizer, there's no way to get it working without building boost and the stdlib with memory sanitization turned on.

sublimator:
I *think* boost *was* getting built with it turned, but will need to look
into the stdlib

PR #9, opened 2016-05-09
1.0.0-b4

vinniefalco:


seelabs:
The warning level CMakeLists is too strict (the -Wconversion warnings are particularly painful to deal with). I turned down the warning level. I also removed some stray semi-colons that `-Wpedantic` didn't like. This commit has the changes I have in mind: https://github.com/seelabs/Beast/commit/8783827d93075eee21eed398e17c6e663f233ada

(Note: removing the semi-colon at the end of `BEAST_DEFINE_TESTSUITE_INSERT` _might_ cause some code to fail to compile if they did not include the semi themselves. If rippled does this we should fix rippled).

seelabs:
Nit: There is some stray whitespace in various files. This commit fixes them: https://github.com/seelabs/Beast/commit/ff734155033590c55d65f750c5de1b5b6f70a390

seelabs:
Nit: There are some spelling errors in some of the comments. This commit fixes the ones I noticed: https://github.com/seelabs/Beast/commit/fae3db6fb8ba8b71d3304b6483fcba816ca05d7a

seelabs:
Nit: `explicit` was used in some places where it wasn't needed, and could have been used in one place where it was absent. This commit fixes the ones I noticed: https://github.com/seelabs/Beast/commit/cf3f43b2fac8718647e943c610f9b45525273206

include/beast/http/basic_parser_v1.hpp
```diff
@@ -412,6 +431,24 @@ class basic_parser_v1
     needs_eof(std::false_type) const;
 
     template<class C>
+    class has_on_start_t
+    {
+        template<class T, class R =
+            decltype(std::declval<T>().on_start(
+                std::declval<error_code&>()),
+                    std::true_type{})>
+        static R check(int);
```
seelabs: We know `R` must be `std::true_type` here. We could simplify this trait by making `R` an unnamed template param and removing the `std::true_type` from the template and moving it to the function return type. Commit showing what I have in mind: https://github.com/seelabs/Beast/commit/dca9d2098380ff4c10d019597b0f3eafa4d25390

No change needed (and what you have here matches what you're done in other places. More of a comment for future traits).

test/http/detail/chunk_encode.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <beast/unit_test/suite.hpp>
+
+namespace beast {
+namespace http {
+namespace detail {
+
+class chunk_encode_test : public beast::unit_test::suite
+{
+public:
+    struct final_chunk
```
seelabs: I don't think we need the `final_chunk` class. We could add an overload to `encode` that will let us know when we've reached the last arg. Commit showing what I have in mind: https://github.com/seelabs/Beast/commit/8d9db13782aa8655c71ac8c69daf2a5b096bc67d

include/beast/http/impl/write.ipp
```diff
@@ -170,13 +148,13 @@ class write_op
         }
     };
 
-    class writef0
+    class writef0_lambda
```
seelabs: I don't see where the equivalent lambda would use any C++-14 features. Why the work-around class? (same comment for `writef_lambda`)

include/beast/http/impl/write.ipp
```diff
@@ -500,45 +478,43 @@ write(SyncWriteStream& stream,
             cv.notify_one();
         }};
     auto copy = resume;
-    for(;;)
+    boost::tribool result;
+    result = wp.w(std::move(copy), ec,
```
seelabs: Nit: initialize `result` in the decl above.

seelabs:
Not in this changeset, but I noticed a cast to `uint64_t` was probably a meant to be a `uint32_t`. Commit showing code in question: https://github.com/seelabs/Beast/commit/ea0fda48a67b7eedc5ab9a8e7e0beda167b2a638

include/beast/core/static_string.hpp
```diff
@@ -512,75 +495,67 @@ int compare(
 }
 
 template<std::size_t N, std::size_t M, class CharT, class Traits>
-int compare(
+inline
```
seelabs: I don't think `inline` is needed here. No harm tho.

include/beast/http/impl/basic_parser_v1.ipp
```diff
@@ -8,7 +8,8 @@
 #ifndef BEAST_HTTP_IMPL_BASIC_PARSER_V1_IPP
 #define BEAST_HTTP_IMPL_BASIC_PARSER_V1_IPP
```
seelabs: Would it make sense to have a `http::v1` namespace instead of a `_v1` suffix?

include/beast/http/impl/write.ipp
vinniefalco: `auto` lambda parameters are C++14

include/beast/http/impl/write.ipp
vinniefalco: its intentional, because the variable is set more than once

include/beast/http/basic_parser_v1.hpp
vinniefalco: Have you tried this on all platforms?

test/http/basic_parser_v1.cpp
```diff
@@ -236,13 +189,22 @@ class basic_parser_v1_test : public beast::detail::unit_test::suite
     parse(boost::string_ref const& m, F&& f)
     {
         using boost::asio::buffer;
+        for(;;)
```
seelabs: `do {...} while(false);` is clearer to me. We could remove one `break` and it makes it clearer that it's not really a loop.

test/http/detail/chunk_encode.cpp
vinniefalco: I don't know that I care enough to want to "simplify" test code but even so, `final_chunk` makes it clear to readers whereas `std:string{}` is obfuscated.

include/beast/http/impl/write.ipp
vinniefalco: Although I could be convinced to change it for the slight performance edge.

include/beast/core/static_string.hpp
vinniefalco: I want it to be inlined

include/beast/http/impl/basic_parser_v1.ipp
vinniefalco: Its an interesting idea. Somewhat cosmetic. Note, however, that `v1`, `v2`, ... named namespaces are often used to version APIs, commonly in conjunction with `inline` namespace declarations:
http://blog.aaronballman.com/2011/07/inline-namespaces/

test/http/basic_parser_v1.cpp
```diff
@@ -481,6 +529,87 @@ class basic_parser_v1_test : public beast::detail::unit_test::suite
         parse_ev<true>("GET / HTTP/1.1\r\nf :", parse_error::bad_field);
     }
 
+    void testInvalidMatrix()
+    {
+        using boost::asio::buffer;
+        std::string s;
+        std::size_t n;
+
+        for(n = 0;; ++n)
```
seelabs: Nit: Move the decl of `n` into the loop and remove the one at the outer scope: `for(size_t n=0;....)`.

include/beast/core/detail/sha1.hpp
```diff
@@ -275,15 +275,14 @@ finish(sha1_context& ctx, void* digest) noexcept
     std::uint64_t total_bits =
         (ctx.blocks*64 + ctx.buflen) * 8;
     // pad
-    auto const buflen = ctx.buflen;
     ctx.buf[ctx.buflen++] = 0x80;
+    auto const buflen = ctx.buflen;
```
seelabs: Instead of saving the old buflen here, how about a bool that saves if we should do the extra transform or not?: `auto const extra_transform = (ctx.buflen > BLOCK_BYTES - 8);`

include/beast/core/detail/sha1.hpp
```diff
@@ -275,15 +275,14 @@ finish(sha1_context& ctx, void* digest) noexcept
...
     while(ctx.buflen < 64)
         ctx.buf[ctx.buflen++] = 0x00;
     std::uint32_t block[BLOCK_INTS];
     sha1::make_block(ctx.buf, block);
     if (buflen > BLOCK_BYTES - 8)
     {
-        ++ctx.blocks;
         sha1::transform(ctx.digest, block);
         for (size_t i = 0; i < BLOCK_INTS - 2; i++)
             block[i] = 0;
```
seelabs: Replace this loop with `memset`?

test/http/write.cpp
```diff
@@ -87,22 +124,333 @@ class write_test : public beast::detail::unit_test::suite
...
+                return *this;
+            }
+        };
+
+        class writer
+        {
+            std::size_t n_ = 0;
+            value_type const& body_;
+            bool suspend_ = false;
+            enable_yield_to yt_;
```
seelabs: Note that `enable_yield_to` is advertised to be used as a mix-in, not a member var. When I think mix-in I think inheritance. Fine as-is.

include/beast/http/basic_parser_v1.hpp
seelabs: Linux only. Yeah, getting traits to work on windows must have been painful. Might be better to leave well enough alone.

include/beast/http/basic_parser_v1.hpp
```diff
@@ -412,6 +431,24 @@ class basic_parser_v1
...
+        template <class>
+        static std::false_type check(...);
+        using type = decltype(check<C>(0));
+    public:
+        static bool const value = type::value;
+    };
+    template<class C>
+    using has_on_start =
```
seelabs: Style note: I've seen the standard tack a `_v` suffix onto things like `has_on_start` (and a `_v` for values, `_t` for types). No change needed.

include/beast/core/impl/basic_streambuf.ipp
```diff
@@ -582,7 +582,7 @@ basic_streambuf<Allocator>::prepare(size_type n) ->
         list_.push_back(e);
         if(out_ == list_.end())
             out_ = list_.iterator_to(e);
-        if(n > e.size())
+        if(n >= e.size())
```
seelabs: I don't see where this fixes anything. It's correct as written, but as far as I can tell the old code was also correct. Why the change?

include/beast/core/detail/sha1.hpp
vinniefalco: I am very reluctant to change this code since it is based on another public work and it would be nice to make as few changes as possible (except those necessary for C++).

test/http/write.cpp
vinniefalco: Hey, I provide the hammer. You're free to hold it any way that you want. Including using your feet.

include/beast/core/impl/basic_streambuf.ipp
vinniefalco: It increases coverage

include/beast/http/basic_parser_v1.hpp
vinniefalco: Yeah unfortunately I'm limited to C++11 here so we can't do the `_v` or the modern `_t`.

seelabs:
:+1:

nbougalis:
:+1:

PR #14, opened 2016-05-18
WebSocket, test coverage

vinniefalco:
Tons of fixes including several crashes. Major coverage increase.


seelabs:
gcc 6.1 issues the following warning: 
`/home/swd/projs/common/boost_1_61_0/boost/function/function_template.hpp:572:11: warning: placement new constructing an object of type ‘boost::_bi::bind_t<std::vector<boost::program_options::basic_option<char> >, boost::_mfi::mf1<std::vector<boost::program_options::basic_option<char> >, boost::program_options::detail::cmdline, std::vector<std::__cxx11::basic_string<char> >&>, boost::_bi::list2<boost::_bi::value<boost::program_options::detail::cmdline*>, boost::arg<1> > >’ and size ‘24’ in a region of type ‘char’ and size ‘1’ [-Wplacement-new=]
           new (reinterpret_cast<void*>(&functor.data)) FunctionObj(f);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~`
Both clang 3.8 nor gcc 5.1 compile clean. This looks more like a boost issue than a beast issue. Just a heads up and something to look at later.
I

extras/beast/unit_test/suite_info.hpp
```diff
@@ -52,61 +63,58 @@ class suite_info
...
-
 inline
 bool
-operator< (suite_info const& lhs, suite_info const& rhs)
+operator<(suite_info const& lhs, suite_info const& rhs)
 {
-    return lhs.full_name() < rhs.full_name();
+    auto const& ls = lhs.full_name();
+    auto const& rs = rhs.full_name();
+    return std::strcmp(ls.c_str(), rs.c_str()) < 0;
```
seelabs: `full_name` does string concatenation, which may alloc. I don't like that in a compare. We could use `std::tuple` to do the comparison without allocating:
```   
    friend
    bool
    operator<(suite_info const& lhs, suite_info const& rhs)
    {
        return std::tie (lhs.name_, lhs.module_, lhs.library_) <
            std::tie (rhs.name_, rhs.module_, rhs.library_);
    }
```

extras/beast/unit_test/dstream.hpp
```diff
@@ -0,0 +1,144 @@
...
+    ~dstream_buf()
+    {
+        sync();
+    }
+
+    int
+    sync() override
+    {
+        write(str().c_str());
+        str("");
```
nbougalis: This would likely error in `-DUNICODE` builds. Why not this instead:

    str({});

If that (for some reason) doesn't work, the more explicit alternative will:

    str(std::basic_string<CharT, Traits, Allocator>{});

extras/beast/unit_test/dstream.hpp
```diff
@@ -130,7 +130,7 @@ class basic_dstream
 {
 public:
     basic_dstream()
-        : std::basic_ostream<CharT, Traits>(&member)
+        : std::basic_ostream<CharT, Traits>(&this->member)
```
nbougalis: This does do the right thing - `member` is from `boost::base_from_member` which has already been initialized at this point.

extras/beast/unit_test/suite.hpp
```diff
@@ -44,95 +45,100 @@ class suite
...
+            sync();
+        }
 
-        /** Returns the raw stream used for output. */
-        abstract_ostream&
-        stream();
+        int
+        sync() override
+        {
+            auto const& s = this->str();
```
nbougalis: Why is `this->` needed here and a couple of lines below?

extras/beast/unit_test/dstream.hpp
```diff
@@ -0,0 +1,144 @@
...
+#ifdef _MSC_VER
+
+template<class CharT, class Traits, class Allocator>
+class dstream_buf
+    : public std::basic_stringbuf<CharT, Traits, Allocator>
+{
+    bool dbg_;
+
+    template<class T>
+    void write(T const*) = delete;
```
seelabs: I'm not sure what this is doing. Preventing automatic conversions to `char const*` maybe?

doc/types/Field.qbk
```diff
@@ -0,0 +1,41 @@
...
+
+[table Field requirements
+
+[[operation][type][semantics, pre/post-conditions]]
+[
+    [`a.name()`]
+    [`boost::string_ref`]
+    [
+        This function returns a value implicitly convertible to
+        `boost::string_ref` containing the case-insensitve field
```
seelabs: insensitve -> insensitive

extras/beast/unit_test/suite_list.hpp
```diff
@@ -18,34 +18,41 @@
...
-/** A container of test suites. */
+/// A container of test suites.
 class suite_list
     : public detail::const_container <std::set <suite_info>>
 {
 private:
-#ifndef NDEBUG
-    std::unordered_set <std::string> names_;
-    std::unordered_set <std::type_index> classes_;
+#if 0//#ifndef NDEBUG
```
nbougalis: Why? Should this just be removed?

extras/beast/unit_test/suite_list.hpp
```diff
@@ -18,34 +18,41 @@
...
 void
-suite_list::insert (char const* name, char const* module, char const* library,
+suite_list::insert(
+    char const* name,
+    char const* module,
+    char const* library,
     bool manual)
 {
-#ifndef NDEBUG
+#if 0//#ifndef NDEBUG
```
nbougalis: Why? Should this just be removed?

extras/beast/unit_test/suite.hpp
seelabs: @nbougalis I had the same question. gcc give a nice explanation (along with a fix!):

`extras/beast/unit_test/suite.hpp:76:16: error: ‘str’ was not declared in this scope, and no declarations were found by argument-dependent lookup at the point of instantiation [-fpermissive]
             str("");
                ^
extras/beast/unit_test/suite.hpp:76:16: note: declarations in dependent base ‘std::__cxx11::basic_stringbuf<char>’ are not found by unqualified lookup
extras/beast/unit_test/suite.hpp:76:16: note: use ‘this->str’ instead`

doc/types/Parser.qbk
```diff
@@ -0,0 +1,58 @@
...
+]
+[
+    [`a.write(b, ec)`]
+    [`std::size_t`]
+    [
+        Parses the octets in the specified input buffer sequentially until
+        an error occurs, the end of the buffer is reached, or a complete
+        HTTP/1 message has been parsed. If an error occurs, `ec` is set
+        to the error code and parsing stops. This function returns the
+        number of bytes consumed in the input buffer.
```
nbougalis: "consumed in the input buffer" should be "consumed from the input buffer"

include/beast/http/impl/basic_parser_v1.ipp
```diff
@@ -463,8 +463,9 @@ write(boost::asio::const_buffer const& buffer, error_code& ec)
```
nbougalis: The commit message needs a bit of a cleanup. The first paragraph of the second sentence says that this fixes a bug, repeating what the first paragraph says:

> Fix parser off by one accessing buffer bug:
> This fixed a bug where in some cases the parser could dereference past the end of the caller provided buffer.

Consider instead:

> Fix off-by-one buffer accessing bug in parser:
> In some case the parser would dereference past the end of the caller provided buffer.

extras/beast/unit_test/suite.hpp
nbougalis: Nice!

test/core/buffer_cat.cpp
```diff
@@ -66,6 +66,12 @@ class buffer_cat_test : public unit_test::suite
             expect(buffer_size(buffer_cat(
                 sb1.data(), sb2.data())) == 12);
         }
+        for(auto it = bs.begin(); it != bs.end(); ++it)
+        {
+            decltype(bs)::const_iterator copy;
+            copy = it;
+            copy = copy;
```
nbougalis: What?

include/beast/websocket/stream.hpp
```diff
@@ -896,6 +903,84 @@ class stream : public detail::stream_base
     #endif
     async_close(close_reason const& cr, CloseHandler&& handler);
 
+    /** Send a WebSocket ping frame.
+
+        This function is used to sycnhronously send a ping frame on
```
seelabs: A couple spelling errors in this file (sycnhronously here). Cherry pick this to fix what I found: https://github.com/seelabs/Beast/commit/db564b94bc06410548a833c3f16a0c9dfe5e712d

test/websocket/stream.cpp
```diff
@@ -477,59 +491,8 @@ class stream_test
...
-    };
-
-    template<class... Vn>
-    cbuf_helper<sizeof...(Vn)>
-    cbuf(Vn... vn)
-    {
-        return cbuf_helper<sizeof...(Vn)>(vn...);
-    }
-
     void testClose(endpoint_type const& ep, yield_context do_yield)
```
seelabs: How is `testClose` called? Is this dead code?

include/beast/http/concepts.hpp
```diff
@@ -0,0 +1,156 @@
...
+namespace beast {
+namespace http {
+
+namespace detail {
+
+template<class T>
+class has_value_type
+{
+    template<class U, class R =
+        typename U::value_type>
```
seelabs: Nit: We don't need `R` `class =` should work.

test/websocket/stream.cpp
```diff
@@ -29,395 +31,407 @@ class stream_test
...
-        expect(detail::clamp(
-            std::numeric_limits<std::uint64_t>::max()) ==
-                std::numeric_limits<std::size_t>::max());
-    }
+        stream<socket_type> ws;
 
-    void testSpecialMembers()
-    {
-        stream<socket_type> ws(ios_);
+        explicit
```
seelabs: Nit: `explicit` not needed.

test/core/buffer_cat.cpp
vinniefalco: test of self-assignment

extras/beast/unit_test/suite_list.hpp
vinniefalco: It should be put back, I commented it out to see if it fixed a crashing bug and forgot to put it back

extras/beast/unit_test/dstream.hpp
vinniefalco: Yes

test/websocket/stream.cpp
vinniefalco: Its currently not called but I plan on circling back at some point and trying to get those last few lines that don't have coverage.

include/beast/websocket/stream.hpp
vinniefalco: FYI, I use UK spelling sometimes (e.g. initialise)

vinniefalco:
Incorporated all review feedback and rewrote the branch

include/beast/websocket/impl/read_frame_op.ipp
```diff
@@ -129,373 +130,415 @@ template<class NextLayer>
...
+template<class Buffers, class Handler>
+void
+stream<NextLayer>::read_frame_op<Buffers, Handler>::
 operator()(error_code ec,std::size_t bytes_transferred, bool again)
 {
+    enum
+    {
+        do_start = 0,
+        do_read_payload = 1,
+        do_frame_done = 3,
```
seelabs: I like how the state machine turned out with the enum and the `case state+1:` idiom. One idea is to reserve some low-order bits for this purpose. i.e all enums are numbered sequentially with the pattern `do_start = 0 << 2, do_read_payload = 1 << 2,` etc. That makes it easy to audit code (if I ever see `state+4` I know it's an error).

vinniefalco:
I added a commit to update the documentation. The HTML preview is here:
http://vinniefalco.github.io/stage/beast/index.html

seelabs:
:+1: Code looks good & tests look good.I see the travis coverage build is failing with the "negative counts" issue, I wish I knew what was going on with that.

vinniefalco:
"negative counts" is a spurious failure that is often solved merely by restarting the build.

extras/beast/unit_test/main.cpp
```diff
@@ -38,32 +38,34 @@ prefix(suite_info const& s)
...
+static
 void
-print(Log& log)
+print(std::ostream& os)
 {
-    log << "------------------------------------------";
-    print(log, global_suites());
-    log << "------------------------------------------";
+    os << "------------------------------------------\n";
+    print(os, global_suites());
```
nbougalis: I'd inline the body of `print(std::ostream&, suite_list const&)` in here.

doc/websocket.qbk
```diff
@@ -67,43 +69,47 @@ both Boost.Asio and the WebSocket protocol specification described in
...
 
-ssl::context ctx(ssl::context::sslv23);
-websocket::stream<ssl::stream<ip::tcp::socket>> wss(ios, ctx);
+boost::asio::ssl::context ctx(boost::asio::ssl::context::sslv23);
+beast::websocket::stream<
+    boost::asio::ssl::stream<boost::asio::ip::tcp::socket>> wss(ios, ctx);
 ```
 
 For servers that can handshake in multiple protocols, it may be desired
 to wrap an object that already exists. This socket can be moved in:
```
nbougalis: You use "object" and "socket" to refer to, seemingly, the same thing. This can be a bit confusing. Might make more sense if you did this:

> For servers that can handshake in multiple protocols, it may be desired to wrap an object that already exists. This can be achieved by moving the object in. For example, a socket could be moved in:

doc/websocket.qbk
```diff
@@ -152,23 +159,23 @@ been upgraded. The HTTP Upgrade request must include the Host HTTP field,
 and the URI of the resource to request. `hanshake` is used to send the
 request with the required host and resource strings.
 ```
-    websocket::stream<ip::tcp::socket> ws(ios);
+    beast::websocket::stream<boost::asio::ip::tcp::socket> ws(ios);
     ...
-    ws.set_option(websocket::keep_alive(true));
+    ws.set_option(beast::websocket::keep_alive(true));
     ws.handshake("ws.mywebapp.com:80", "/cgi-bin/bitcoin-prices");
```
nbougalis: Consider using the canonical `example.com` domain in your examples and documentation - see [RFC 2606](https://tools.ietf.org/html/rfc2606)).

doc/websocket.qbk
```diff
@@ -152,23 +159,23 @@ been upgraded. The HTTP Upgrade request must include the Host HTTP field,
...
 
-The `websocket::stream` automatically handles receiving and processing
-the HTTP response to the handshake request. The call to handshake is
-successful if a HTTP response is received with the 101 "Switching Protocols"
-status code. On failure, an error is returned or an exception is thrown.
-Depending on the keep alive setting, the socket may remain open for a
-subsequent handshake attempt
+The [link beast.ref.websocket__stream `beast::websocket::stream`] automatically
+handles receiving and processing the HTTP response to the handshake request.
+The call to handshake is successful if a HTTP response is received with the
```
nbougalis: This is a super-minor nit, and I get that reasonable people can view this topic differently - just pointing it out: "an HTTP response" reads better than "a HTTP response".

doc/websocket.qbk
vinniefalco: The problem is that `boost::asio::ssl::stream` is not a socket, but yet it is suitable as a `NextLayer` type in `beast::websocket::stream`.

doc/websocket.qbk
vinniefalco: Good idea

doc/websocket.qbk
vinniefalco: What reads better:
"an hypertext transfer protocol response"
                   -or-
"a hypertext transfer protocol response"
                    ?

Both are correct, as long as usage is consistent.

nbougalis: I suspect that the "correct" usage is:

> a hypertext transfer protocol response 

and 

> an HTTP response

According to http://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html "trick here is to use your ears (how the acronym is pronounced), not your eyes (how it's spelled)." which isn't perfect, but it tends to work.

test/core/buffer_cat.cpp
nbougalis: If you're going to test self-assignment, then you should ensure that the result is sane and not just successful. For example, you could `expect (copy == it);` after each of the assignments.

nbougalis:
:+1:

doc/websocket.qbk
```diff
@@ -152,23 +159,23 @@ been upgraded. The HTTP Upgrade request must include the Host HTTP field,
 and the URI of the resource to request. `hanshake` is used to send the
```
seelabs: `hanshake` -> `handshake`

seelabs:
Still :+1: (though you should fix the minor typo I found).

PR #15, opened 2016-05-23
Do not run coverage until all jobs have finished

seelabs:
Sometimes lcov would report negative counts. This may be related to lcov
running while the job it was collecting counts for was also running.
This patch makes two changes: Do not run the coverage build under
gdb, and wait for all jobs to finish before running lcov.

@ximinez @sublimator

### DIFF
--- a/scripts/build-and-test.sh
+++ b/scripts/build-and-test.sh
@@ -19,10 +19,25 @@ function run_tests_with_gdb {
   for x in bin/**/*-tests; do scripts/run-with-gdb.sh "$x"; done
 }
 
+function run_tests {
+    for x in bin/**/*-tests; do "$x"; done
+}
+
+num_procs=1
+if [[ $(uname) == "Darwin" ]]; then
+    num_procs=$(sysctl -n hw.ncpu)
+elif [[ $(uname -s) == "Linux" ]]; then
+    num_procs=$(lscpu -p | grep -v '^#' | sort -u -t, -k 2,4 | wc -l) # physical cores
+    virt_num_procs=$(nproc) # CircleCI returns 32 phys procs, but 1 virt proc
+    if (("$virt_num_procs" < "$num_procs")); then
+        num_procs=$virt_num_procs
+    fi
+fi
+
 function build_beast {
   $BOOST_ROOT/bjam toolset=$CC \
                variant=$VARIANT \
-               address-model=$ADDRESS_MODEL
+               address-model=$ADDRESS_MODEL -j${num_procs}
 }
 
 ##################################### BUILD ####################################
@@ -38,11 +53,11 @@ if [[ $VARIANT == "coverage" ]]; then
   lcov --no-external -c -i -d . -o baseline.info > /dev/null
 
   # Perform test
-  run_tests_with_gdb
+  run_tests
 
   # Run autobahn tests
-  export SERVER=`find . -name "websocket-echo"`
-  nohup scripts/run-with-gdb.sh $SERVER&
+  export SERVER=$(find . -name "websocket-echo")
+  nohup $SERVER&
 
   # We need to wait a while so wstest can connect!
   sleep 5
@@ -52,10 +67,10 @@ if [[ $VARIANT == "coverage" ]]; then
   cat nohup.out
   rm nohup.out
   jobs
+  sleep 5
   # Kill it gracefully
   kill -INT %1
-  sleep 1
-  kill -INT %1 || echo "Dead already"
+  wait
 
   # Create test coverage data file
   lcov --no-external -c -d . -o testrun.info > /dev/null


vinniefalco:
Nice looking changes especially the number of processors

sublimator:
good call on the wait!

Gdb messes with coverage results?

scripts/build-and-test.sh
```diff
@@ -19,10 +19,16 @@ function run_tests_with_gdb {
   for x in bin/**/*-tests; do scripts/run-with-gdb.sh "$x"; done
 }
 
+function run_tests {
+    for x in bin/**/*-tests; do "$x"; done
+}
+
+num_procs=$(lscpu -p | grep -v '^#' | sort -u -t, -k 2,4 | wc -l) # pysical cores
```
sublimator: This doesn't work on OSX, nor docker

scripts/build-and-test.sh
```diff
@@ -38,11 +44,11 @@ if [[ $VARIANT == "coverage" ]]; then
   lcov --no-external -c -i -d . -o baseline.info > /dev/null
 
   # Perform test
-  run_tests_with_gdb
+  run_tests
 
   # Run autobahn tests
-  export SERVER=`find . -name "websocket-echo"`
-  nohup scripts/run-with-gdb.sh $SERVER&
```
sublimator: I've noticed some stalls some time, which were the reason for the (hamfisted) extra kill invocation (one possible cause of the issue).

scripts/build-and-test.sh
```diff
@@ -54,8 +60,7 @@ if [[ $VARIANT == "coverage" ]]; then
   jobs
   # Kill it gracefully
   kill -INT %1
-  sleep 1
-  kill -INT %1 || echo "Dead already"
+  wait
```
sublimator:
```

[1]+  Running                 nohup $wsecho &
+kill -INT %1
+wait
+lcov --no-external -c -d . -o testrun.info
+lcov -a baseline.info -a testrun.info -o lcov-all.info
lcov: WARNING: negative counts found in tracefile testrun.info
lcov: ERROR: cannot write to lcov-all.info!
```

hrmmm ???  You might have gotten lucky here

sublimator: One thing to note, which may or may not be pertinent, is that the travis run for this pull request didn't actually run `wstest`, even though it started the echo server.

scripts/build-and-test.sh
sublimator:
```
++lscpu -p
++wc -l
++sort -u -t, -k 2,4
++grep -v '^#'
+num_procs=1
+build_beast
+/home/travis/boost_1_60_0/bjam toolset=gcc variant=asan address-model=64 -j1
```

Seems like Travis is just -j1 as per the usual implicit default. I'm using this in an updated branch:
```
if [[ `uname` == "Darwin" ]]; then
  num_jobs=`sysctl -n hw.ncpu`
else
  num_jobs=`nproc`
fi
```

This works on circle CI Docker and locally and on Travis:
```
+[[ Linux == \D\a\r\w\i\n ]]
++nproc
+num_jobs=2
```

scripts/build-and-test.sh
sublimator: Noting that, in an updated version, I added a sleep of 5 seconds to let the server `wind down` before issuing a kill. Not sure if it's needed or it will help ?

scripts/build-and-test.sh
seelabs: Thanks, pushed a fix for osx

scripts/build-and-test.sh
seelabs: Rats. OK, thanks. We'll have to think of something else.

scripts/build-and-test.sh
```diff
@@ -19,10 +19,21 @@ function run_tests_with_gdb {
...
 
+function run_tests {
+    for x in bin/**/*-tests; do "$x"; done
+}
+
+num_procs=1
+if [[ "$(uname)" == "Darwin" ]]; then
+    num_procs=$(sysctl -n hw.ncpu)
+elif [[ "$(expr substr $(uname -s) 1 5)" == "Linux" ]]; then
+    num_procs=$(lscpu -p | grep -v '^#' | sort -u -t, -k 2,4 | wc -l) # pysical cores
```
sublimator: This only gets 1 on Travis

sublimator: `nproc` works in Docker / linux

Above method shows 32 on CircleCI, when really the container only has  2 `available`

vinniefalco: Spelling: `pysical cores`

scripts/build-and-test.sh
```diff
@@ -23,7 +23,12 @@ function run_tests {
     for x in bin/**/*-tests; do "$x"; done
 }
 
-num_procs=$(lscpu -p | grep -v '^#' | sort -u -t, -k 2,4 | wc -l) # pysical cores
+num_procs=1
+if [[ "$(uname)" == "Darwin" ]]; then
+    num_procs=$(sysctl -n hw.ncpu)
+elif [[ "$(expr substr $(uname -s) 1 5)" == "Linux" ]]; then
```
sublimator: What's the `-s` arg to uname do?

scripts/build-and-test.sh
seelabs: Fixed

scripts/build-and-test.sh
```diff
@@ -19,10 +19,21 @@ function run_tests_with_gdb {
...
 
+function run_tests {
+    for x in bin/**/*-tests; do "$x"; done
+}
+
+num_procs=1
+if [[ "$(uname)" == "Darwin" ]]; then
+    num_procs=$(sysctl -n hw.ncpu)
+elif [[ "$(expr substr $(uname -s) 1 5)" == "Linux" ]]; then
+    num_procs=$(lscpu -p | grep -v '^#' | sort -u -t, -k 2,4 | wc -l) # physical cores
```
sublimator: Why not threads? e.g. my quad core i7 has 8 threads and I typically use -j8 for scons etc. Have I been "doing it wrong? ?

scripts/build-and-test.sh
seelabs: Returns the kernel name.

scripts/build-and-test.sh
seelabs: Yeah, you'll get better compilation times if you use physical cores rather than virtual cores (at least on the machines I've tested).

seelabs:
@sublimator I pushed a fix that takes the min of physical cores and virtual cores. That should work well everywhere.

scripts/build-and-test.sh
sublimator: Lesson learned! Cheers

scripts/build-and-test.sh
sublimator: fwiw, since I added in the 5 second sleep after wstest, I haven't noticed any further issues.

scripts/build-and-test.sh
```diff
@@ -19,10 +19,25 @@ function run_tests_with_gdb {
   for x in bin/**/*-tests; do scripts/run-with-gdb.sh "$x"; done
 }
 
+function run_tests {
+    for x in bin/**/*-tests; do "$x"; done
+}
+
+num_procs=1
+if [[ "$(uname)" == "Darwin" ]]; then
+    num_procs=$(sysctl -n hw.ncpu)
```
sublimator: We'll want `ysctl -n hw.physicalcpu` to get the `physical cores` on OSX it seems :)

vinniefalco:
This is merged in b5 but we still get "negative counts" from lcov.

PR #23, opened 2016-06-03
1.0.0-b6

vinniefalco:
New features include HTTP field parsers, strict parsing of rfc7230, and 100% code coverage in `basic_parser_v1`.


CHANGELOG
```diff
@@ -1,3 +1,6 @@
 1.0.0-b6
 
+* Use SFINAE on return values
+* Use beast::error_code instead of nested types
```
scottschurr: I'm not sure about your intention for squashing but, at least for this iteration, it appears that using `beast::error_code` doesn't happen until the next commit.  If they get squashed it really doesn't matter.  Just FYI.

vinniefalco: Edits to text files like TODO.txt and CHANGELOG are not necessarily temporally consistent with edits made to source code :-) 

For CHANGELOG what matters is the accuracy at the time of the commit that sets the version.

include/beast/core/streambuf_readstream.hpp
```diff
@@ -93,8 +93,6 @@ class streambuf_readstream
     static_assert(is_Streambuf<Streambuf>::value,
         "Streambuf requirements not met");
 
-    using error_code = boost::system::error_code;
-
     template<class Buffers, class Handler>
     class read_some_op;
```
HowardHinnant: If not squashing this commit message out, would be better if it read:  Use boost::error_code (if I'm understanding correctly).

vinniefalco: I think there's a misunderstanding. What I have done is create a `beast` level identifier `error_code` (which happens to be an alias for `boost::system::error_code`). The nature of the refactoring is that all code now uses this `beast` level namespace identifier instead of having every class nest its own alias for `boost::system::error_code` out of convenience.

HowardHinnant:
There's a growing danger of deriving from `detail::types`.  When considering ADL the compiler makes a list of associated namespaces within which it will search for overloads.  One of the associated namespaces is that of a base class (even if privately derived).  So when you have `beast::X` derive (even privately) from `beast::detail::Y`, `beast::detail` becomes an associated namespace of `beast::X` (an implementation detail leakage).

This has the potential to cause either compile-time errors, or in extreme cases, even run time errors.  Here's a contrived demo:
```
namespace beast {
namespace detail {

template <class T>
void
tolower(T);

struct base {};

}  // detail

struct derived : private detail::base
{
    operator long() const {return 0;}
};

}  // beast

// client code

#include <cctype>

int
main()
{
    beast::derived d;
    using namespace std;
    tolower(d);
}
```
The client intends to call `std::tolower()` from `<cctype>`, but ends up calling `beast::detail::tolower` because `beast::detail` is an associated namespace of `beast::derived`, and `beast::detail::tolower`  turns out to be a better match than `std::tolower`.

An alternative design using private data members instead of private inheritance avoids this issue:
```
struct derived
{
private:
    detail::base data_;
public:
    operator long() const {return 0;}
};
```

vinniefalco:
@HowardHinnant ack... which Beast code were you talking about?

HowardHinnant:
It was `tolower` that started me looking.  I found `basic_headers` as such an example.  Other places are deriving from `ci_less`, but I'm as yet unsure if any of those places are actual examples of this type of leakage.

test/http/nodejs_parser.hpp
```diff
@@ -611,7 +611,7 @@ nodejs_basic_parser<Derived>::check_header()
 {
     if (! value_.empty())
     {
-        rfc2616::trim_right_in_place(value_);
+        //detail::trim(value_);
```
scottschurr: I'm guessing you want to remove this commented out line?

vinniefalco: Nope, I want to leave it in. I got rid of the `trim` function and couldn't be bothered to fix this line of code because its part of the `nodejs_parser`, which is only for performance benchmark comparisons and not a public API. If we ever decide to use this code for anything else, we'd probably want to fix the commented out line.

include/beast/http/basic_dynabuf_body.hpp
```diff
@@ -0,0 +1,95 @@
...
+            using boost::asio::buffer;
+            using boost::asio::buffer_copy;
+            sb_.commit(buffer_copy(
+                sb_.prepare(size), buffer(data, size)));
+        }
+    };
+
+    class writer
+    {
+        DynamicBuffer const& body_;
```
scottschurr: `DynamicBuffer` -> `value_type`?

HowardHinnant:
👍

scottschurr:
Looks good to me. :+1:

PR #24, opened 2016-06-08
Medley

vinniefalco:
This is a response to compilation problems encountered by a user


include/beast/websocket/teardown.hpp
```diff
@@ -56,12 +77,63 @@ teardown(Socket& socket, error_code& ec) = delete;
     manner equivalent to using boost::asio::io_service::post().
 
 */
+/*
+    If you are trying to use OpenSSL and get an error about this function
+    being deleted, you need to include <beast/websocket/ssl.hpp> in the
+    corresponding source file!
+*/
```
ximinez: This comment can be removed. The function is not deleted anymore, and you have a similar comment for the `static_assert`.

ximinez:
Looks good. :+1: I assume we'll have to wait for #29 to get the CI to pass.

include/beast/http/parser_v1.hpp
```diff
@@ -62,6 +93,7 @@ class parser_v1
     std::string value_;
     message_type m_;
     typename message_type::body_type::reader r_;
+    std::uint8_t skip_body_ = 0;
```
seelabs: Why not `bool`?

seelabs:
:+1:

include/beast/http/parser_v1.hpp
vinniefalco: It can take on 3 values:

0 - Don't skip
1 - Skip
2 - Is HTTP Upgrade

include/beast/websocket/teardown.hpp
vinniefalco: True

include/beast/http/parser_v1.hpp
seelabs: Got it. And case (2) Isn't implemented right now? I only see `skip_body_` taking on values on 1 and 0.

vinniefalco: The base class handles the return value of `on_headers`, and it takes care of values 0, 1, and 2:
https://github.com/vinniefalco/Beast/blob/master/include/beast/http/impl/basic_parser_v1.ipp#L902

include/beast/websocket/teardown.hpp
vinniefalco: Updated

PR #27, opened 2016-06-16
cmake: add missing -pthread to core-tests and websocket-tests

cbodley:


vinniefalco:
Reviewers: @seelabs

seelabs:
:+1: Using the imported target seems to be the preferred over ${CMAKE_THREAD_LIBS_INIT}. On my system (ubuntu 16.04) they seem to do the same thing, though looking at https://github.com/Kitware/CMake/blob/master/Modules/FindThreads.cmake there are differences on other systems.

Good change.

vinniefalco:
@seelabs Is there an easy way to address #28 in this PR?

vinniefalco:
This is a clear improvement and it passes review.
:+1:

vinniefalco:
@cbodley Thanks!

vinniefalco:
This is now part of #29

PR #29, opened 2016-06-17
Add cmake build to travis

seelabs:


seelabs:
@vinniefalco Please note I forced travis to use trusty (ubuntu 14.04) to get a later version of cmake. Is that change OK with you?

vinniefalco:
@seelabs Uhhh... I don't know? What are the pros and cons?

seelabs:
@vinniefalco 

Pros: We can use more recent packages, in this case I care about using a relatively modern cmake. Ubuntu 12.04 is four years old at this point.

Cons: Not all travis features are supported in 14.04 (containers, for example). I don't think we care. I think we should go with 14.04.

seelabs:
@ximinez I'm making some changes to this PR (adding clang and installing a recent cmake on 12.04). Please hold off reviewing until that's done.

seelabs:
@ximinez This is ready to be reviewed. Two notes: 

1) The top two commits are from another PR and don't need to be reviewed there.
2) I left in a hack where a cached directory was found, but the executables expected under that directory were not found (that shouldn't happen). If I figure out what's going on there I'll take out the hacks.

ximinez:
Looks good, and with a little coaxing, Travis succeeds. I'll note that it looks like the second commit overwrites the changes from the first commit (the ones I was supposed to ignore, but didn't). Those two commits should probably be squashed, or just remove the first one. :+1:

vinniefalco:
Part of #39

Issue #30, opened 2016-06-20
http-server did not handle  the connections which are need to be closed.


avdbg:
When I use Apache Bench to test the performance of beast, however i found that 
the connection did not closed, this leading the test failed.
how to repair the problem?

vinniefalco:
I'm looking into it

vinniefalco:
I believe the problem has been resolved in *1.0.0-b7* can you please try it?

PR #31, opened 2016-06-20
Examples, HTTP fixes

vinniefalco:
Addresses #30 

* Fix prepare by calling init. prepare() can throw depending on the
  implementation of Writer. Publicly provided beast::http writers never throw.
* Fixes to example HTTP server
* Fully qualify ambiguous calls to read and parse
* Remove deprecated http::stream wrapper
* Example HTTP server now calculates the MIME-type


examples/http_async_server.hpp
```diff
@@ -67,13 +71,138 @@ class http_async_server
...
     }
 
+    void
+    log_args()
+    {
+    }
+
+    template<class Arg, class... Args>
+    void
+    log_args(Arg const& arg, Args const&... args)
```
scottschurr: Double checking that you really want `log_args()` to be `public`.  It side-steps the thread safety that you so nicely put into `log()`.  Your call.

vinniefalco: I didn't really pay it any mind since its example code and not a public interface, but I can change it.

examples/http_async_server.hpp
```diff
@@ -67,13 +71,138 @@ class http_async_server
...
+        {
+            return op->d_->cont;
+        }
+
+        template <class Function>
+        friend
+        void asio_handler_invoke(Function&& f, write_op* op)
+        {
+            return boost_asio_handler_invoke_helpers::
+                invoke(f, op->d_->h);
```
scottschurr: Should the argument be `std::forward<Function>(f)`?

vinniefalco: Definitely not. The intent here is that `f` can be passed as either `Function const&` or `Function&`.

examples/http_async_server.hpp
```diff
@@ -83,59 +212,81 @@ class http_async_server
         peer& operator=(peer const&) = delete;
 
         explicit
-        peer(socket_type&& sock, std::string const& root)
-            : stream_(std::move(sock))
-            , strand_(stream_.get_io_service())
-            , root_(root)
+        peer(socket_type&& sock, http_async_server& server)
```
scottschurr: I don't believe this multi-argument constructor needs to be marked `explicit`, but there's no harm to it.

examples/http_sync_server.hpp
```diff
@@ -66,22 +69,44 @@ class http_sync_server
...
+    log_args()
     {
-        std::cerr <<
-            what << ": " << ec.message() << std::endl;
     }
 
+    template<class Arg, class... Args>
     void
-    maybe_throw(error_code ec, std::string what)
+    log_args(Arg const& arg, Args const&... args)
```
scottschurr: If you do choose to make `http_async_server::log_args()` private, consider making this `log_args()` private as well.

examples/http_crawl.cpp
```diff
@@ -31,9 +32,10 @@ int main(int, char const*[])
             ip::tcp::resolver r(ios);
             auto it = r.resolve(
                 ip::tcp::resolver::query{host, "http"});
-            stream<ip::tcp::socket> hs(ios);
-            connect(hs.lowest_layer(), it);
-            auto ep = hs.lowest_layer().remote_endpoint();
+            streambuf sb;
```
scottschurr: The tiniest of nits: I'd move the `sb` declaration down close to the first use on line 49.  Your call.

scottschurr:
:+1:

examples/http_crawl.cpp
vinniefalco: Willfix

examples/http_sync_server.hpp
vinniefalco: Willfix

examples/http_async_server.hpp
vinniefalco: Willfix

examples/http_sync_server.hpp
```diff
@@ -65,21 +68,44 @@ class http_sync_server
...
 
+    template<class... Args>
+    void
+    log(Args const&... args)
+    {
+        if(log_)
+        {
+            std::lock_guard<std::mutex> lock(m_);
+            log_args(args...);
+            std::cerr.flush();
```
HowardHinnant: This is wasteful.  `cerr` already flushes after each output operation.

HowardHinnant:
👍 Only one nitpicky comment.  Tests pass on OS X with `-std=c++11`.

examples/http_sync_server.hpp
vinniefalco: Willfix

PR #34, opened 2016-06-24
Fix to_string.hpp include path in example code

JackBondPreston:


vinniefalco:
Ahh, good catch! Thanks!

vinniefalco:
Part of #35 now

PR #35, opened 2016-06-24
Fix to_string.hpp include path in example code 

vinniefalco:
Based on #34 

Reviewers: @miguelportilla


vinniefalco:
Part of #35 now

miguelportilla:
:+1:

PR #37, opened 2016-07-05
Fix rfc2616 Section 4.2 compliance:

vinniefalco:
`basic_headers` no longer combines fields with the same name by appending a comma and concatenating the two values together. This was breaking certain header fields which expect each value to be distinct, such as the "Set-Cookie" header.

Now the container behaves more like a multi set with respect to insertion of multiple values with the same field name. Additional member functions are provided to provide extra functionality.

Reviewers: @HowardHinnant @scottschurr @willwray

### DIFF
--- a/include/beast/http/basic_headers.hpp
+++ b/include/beast/http/basic_headers.hpp
@@ -105,7 +105,7 @@ protected:
     using list_t = typename boost::intrusive::make_list<
         element, boost::intrusive::constant_time_size<false>>::type;
 
-    using set_t = typename boost::intrusive::make_set<
+    using set_t = typename boost::intrusive::make_multiset<
         element, boost::intrusive::constant_time_size<true>,
             boost::intrusive::compare<less>>::type;
 
@@ -117,7 +117,6 @@ protected:
         : set_(std::move(set))
         , list_(std::move(list))
     {
-
     }
 
 public:
@@ -244,9 +243,10 @@ public:
     value.
 
     Field names are stored as-is, but comparison are case-insensitive.
-    The container preserves the order of insertion of fields with
-    different names. For fields with the same name, the implementation
-    concatenates values inserted with duplicate names as per rfc7230.
+    When the container is iterated, the fields are presented in the order
+    of insertion. For fields with the same name, the container behaves
+    as a std::multiset; there will be a separate value for each occurrence
+    of the field name.
 
     @note Meets the requirements of @b `FieldSequence`.
 */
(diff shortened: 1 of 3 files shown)


include/beast/http/impl/basic_headers.ipp
```diff
@@ -205,6 +206,18 @@ basic_headers(FwdIt first, FwdIt last)
 }
 
 template<class Allocator>
+std::size_t
+basic_headers<Allocator>::
+count(boost::string_ref const& name) const
```
willwray: return set_.count(name);  ?

vinniefalco: Whoops!!! Why use existing functions when you can write them again yourself!

willwray: Hmm, the combination of *list* and *set* reminds me of a certain coding
test...

include/beast/http/impl/basic_headers.ipp
```diff
@@ -205,6 +206,18 @@ basic_headers(FwdIt first, FwdIt last)
...
+    if(it == set_.end())
+        return 0;
+    auto const last = set_.upper_bound(name, less{});
+    return static_cast<std::size_t>(std::distance(it, last));
+}
+
+template<class Allocator>
 auto
 basic_headers<Allocator>::
 find(boost::string_ref const& name) const ->
```
scottschurr: Since we're now set up for duplicate entries, I just wanted to make sure we're happy that `find()` always returns the first entry (inserted in time sequence).  I think that's the behavior we'll see (since `insert()` always appends), and I think it's a reasonable behavior.  I just thought I'd mention it...

include/beast/http/basic_headers.hpp
```diff
@@ -244,9 +243,10 @@ class basic_headers_base::const_iterator
     value.
 
     Field names are stored as-is, but comparison are case-insensitive.
-    The container preserves the order of insertion of fields with
-    different names. For fields with the same name, the implementation
-    concatenates values inserted with duplicate names as per rfc7230.
+    When the container is iterated, the fields are presented in the order
+    of insertion. For fields with the same name, the container behaves
+    as a std::multiset; there will be a separate value for each occurrence
```
scottschurr: Doesn't it behave more like a `std::list`, at least if iterators are used?  `begin()` and `end()` return a `list_t::iterator`.

include/beast/http/impl/basic_headers.ipp
scottschurr: Sorry, yup, it's clearly good and expected.  The documentation in the *.hpp file says so.

scottschurr:
:+1: Looks good to me.

include/beast/http/basic_headers.hpp
vinniefalco: Yeah you're right, the comment is incorrect. Will fix

HowardHinnant:
👍 Looks good here.  Passes tests on macOS.

PR #50, opened 2016-08-15
Some build script tuning

vinniefalco:
This tidies up support for building both 32 and 64 bit targets, especially on Windows.


vinniefalco:
Part of #66 now

PR #67, opened 2016-08-29
B12

vinniefalco:
Improvements to the unit test framework, needed by NuDB

Reviewers: @mellery451 @willwray

### DIFF
--- a/extras/beast/unit_test/amount.hpp
+++ b/extras/beast/unit_test/amount.hpp
@@ -23,29 +23,29 @@ private:
     std::string const& what_;
 
 public:
-    amount (amount const&) = default;
-    amount& operator= (amount const&) = delete;
+    amount(amount const&) = default;
+    amount& operator=(amount const&) = delete;
 
     template<class = void>
-    amount (std::size_t n, std::string const& what);
+    amount(std::size_t n, std::string const& what);
 
     friend
     std::ostream&
-    operator<< (std::ostream& s, amount const& t);
+    operator<<(std::ostream& s, amount const& t);
 };
 
 template<class>
-amount::amount (std::size_t n, std::string const& what)
-    : n_ (n)
-    , what_ (what)
+amount::amount(std::size_t n, std::string const& what)
+    : n_(n)
+    , what_(what)
 {
 }
 
 inline
 std::ostream&
-operator<< (std::ostream& s, amount const& t)
+operator<<(std::ostream& s, amount const& t)
 {
-    s << t.n_ << " " << t.what_ << ((t.n_ != 1) ? "s" : "");
+    s << t.n_ << " " << t.what_ <<((t.n_ != 1) ? "s" : "");
     return s;
 }
 
--- a/extras/beast/unit_test/main.cpp
+++ b/extras/beast/unit_test/main.cpp
@@ -45,7 +45,7 @@ print(std::ostream& os, suite_list const& c)
     std::size_t manual = 0;
     for(auto const& s : c)
     {
-        os << prefix (s) << s.full_name() << '\n';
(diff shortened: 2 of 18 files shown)


extras/beast/unit_test/main.cpp
```diff
@@ -80,18 +80,18 @@ int main(int ac, char const* av[])
...
 #endif
 
     namespace po = boost::program_options;
     po::options_description desc("Options");
     desc.add_options()
-        ("help,h",  "Produce a help message")
-        ("print,r", "Print the list of available test suites")
-        ("suites,s", po::value<string>(), "suites to run")
+       ("help,h",  "Produce a help message")
+       ("print,r", "Print the list of available test suites")
```
mellery451: the shortcut arg for print is "r"? Is "-p" already taken?

vinniefalco: No idea why the shortcut is `r` instead of `p`

vinniefalco: This was a whitespace-only change

extras/beast/unit_test/amount.hpp
```diff
@@ -23,29 +23,29 @@ class amount
...
+    operator<<(std::ostream& s, amount const& t);
 };
 
 template<class>
-amount::amount (std::size_t n, std::string const& what)
-    : n_ (n)
-    , what_ (what)
+amount::amount(std::size_t n, std::string const& what)
+    : n_(n)
+    , what_(what)
```
willwray: Are we switching to curly-braces in constructor initializer lists?
(except for the awkward-squad)

vinniefalco: I don't want to be on the cutting edge of that, as compiler support is largely untested.

HowardHinnant:
<nit-picky>  The sub-commit message:
`Also fixes a bug where suite::log was incorrectly flushed.`
Seems a little confusing since `flush` was added, not removed.  However I don't like the double negative in this obvious alternative: `Also fixes a bug where suite::log was incorrectly not flushed.`

Maybe:
`Added flush() as needed to suite::log`
?

vinniefalco:
@HowardHinnant Well there's much more to it than that, the entire algorithm for how the log is flushed is different (see the change calling `log` instead of `on_log`). I could go into more depth..not sure its worth it since this is a change to the unit test code. My thinking, is that from the users perspective the log was flushed incorrectly before, and now its fixed.

extras/beast/unit_test/amount.hpp
HowardHinnant: "awkward-squad"  Going to have to put that one on a bumper sticker! :-)

Fwiw, I got hit with being too-cutting-edge with `{}` just today on my date lib: https://github.com/HowardHinnant/date/issues/78

extras/beast/unit_test/suite.hpp
```diff
@@ -158,74 +161,103 @@ class suite
...
-    template<class F, class String>
+    template<class Condition>
     bool
-    except (F&& f, String const& reason);
+    expect(Condition const& shouldBeTrue,
+        std::string const& reason, char const* file, int line);
+    /** @} */
 
+    //
+    // DEPRECATED
```
mellery451: can template methods be marked `[[deprecated]]`...and would we want to do that here?

mellery451:
👍

extras/beast/unit_test/suite.hpp
vinniefalco: If this was an official API then yeah I would consider it but its in the extras/ directory which means its not official (but still available for use if you accept the risks).

willwray:
👍

nbougalis:
:+1:

PR #86, opened 2016-09-23
Mingw build

e-fominov:
I was unable to build Beast as-is with MinGW and MSCV. This PR makes it possible
Tested with MINGW 6.2 x64, and MSVC 2015 x64, Windows 7

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ project (Beast)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 
-if (WIN32)
+if (MSVC)
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /W4 /wd4100 /bigobj /D _WIN32_WINNT=0x0601 /D_SCL_SECURE_NO_WARNINGS=1 /D_CRT_SECURE_NO_WARNINGS=1")
     set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /SAFESEH:NO")
 else()
@@ -21,6 +21,10 @@ else()
 
     set(CMAKE_CXX_FLAGS
       "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wpedantic")
+endif()
+
+if (MINGW)
+    link_libraries(ws2_32 mswsock ${Boost_LIBRARIES})
 endif()
 
 if ("${VARIANT}" STREQUAL "coverage")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,16 +6,17 @@ project (Beast)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 
+set(Boost_USE_STATIC_LIBS ON)
+set(Boost_USE_MULTITHREADED ON)
+find_package(Boost REQUIRED COMPONENTS coroutine context thread filesystem program_options system)
+include_directories(${Boost_INCLUDE_DIRS})
+link_directories(${Boost_LIBRARY_DIRS})
+link_libraries(${Boost_LIBRARIES})
+
 if (MSVC)
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /W4 /wd4100 /bigobj /D _WIN32_WINNT=0x0601 /D_SCL_SECURE_NO_WARNINGS=1 /D_CRT_SECURE_NO_WARNINGS=1")
     set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /SAFESEH:NO")
 else()
-    set(Boost_USE_STATIC_LIBS ON)
-    set(Boost_USE_MULTITHREADED ON)
-    find_package(Boost REQUIRED COMPONENTS coroutine context thread filesystem program_options system)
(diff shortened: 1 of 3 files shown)


vinniefalco:
Thanks! This looks very promising, although realistically I have no way to test it since I don't use MinGW. Is there some way to have MinGW as a target on Travis?

e-fominov:
I dont have much experience with Travis. MinGW itself is GCC-based and there will be no syntax differences, except the libc is missing on MinGW.

MinGW has cross-compiler that can be installed into Linux platform and also I found one stackoverflow post where it seems possible to use MinGW with Travis: http://stackoverflow.com/questions/28155308/how-to-cross-compile-with-mingw-on-linux-for-windows

vinniefalco:
Hmm...unfortunately these changes break 64-bit MSVC builds and I'm not sure why. This is the command line I used to build the solution and project files after cherry-picking your commits:
```
cd bin64
cmake -G"Visual Studio 14 2015 Win64" ..
```

vinniefalco:
I think I see what's happening. These are the lines causing problems:
```
find_package(Boost REQUIRED COMPONENTS coroutine context thread filesystem program_options system)
link_directories(${Boost_LIBRARY_DIRS})
link_libraries(${Boost_LIBRARIES})
```

Unfortunately with MSVC, we can't use the package finder to set the library directories, because there is no way to distinguish between 32 versus 64 bit builds. What I do is add a path to the Boost 32 bit libraries in the Microsoft.Cpp.Win32.user property sheet, and add a different path to the Boost 64 bit libraries in the Microsoft.Cpp.x64.user property sheet. If we were to use find_package and link_directories for this, then targeting both 32 and 64 bit builds in separate solutions would no longer be possible.

These are the steps I use to generate the separate solutions, please try them and see if they work with your changes:
```
cd bin
cmake ..
cd ../bin64
cmake -G"Visual Studio 14 2015 Win64" ..
```

e-fominov:
Do you have boost compiled for x64?   I have tested this PR with MSVCx64 - compiles well. But my boost is compiled for x64
Looks like D:\lib\boost_1_61_0\stage\lib\ libraries on your machine are x86

This problem is because your original CMakeLists.txt file had no find_module(Boost) for Win32 compilation. Looks like you have something in your System Environment variables that makes it find boost in incorrect place. Normally CMake's findboost module searches it in c:\boost

Now it tries to find boost and finds it in D:\lib\boost_1_61_0\stage\lib\, which contain x86 libraries
To make it use Boost from other folder, its possible to pass additional params to cmake:

```
cd bin64
cmake -G"Visual Studio 14 2015 Win64" .. -DBOOST_INCLUDEDIR=d:\boost64\include -DBOOST_LIBRARYDIR=d:\boost64\lib
```

vinniefalco:
I have 32-bit boost libraries in `D:\lib\boost_1_61_0\stage` and 64-bit boost libraries in `d:\lib\boost_1_61_0\stage64`. I also have `$BOOST_ROOT` set to `/d/lib/boost_1_61_0/`.

You're telling me that I should manually specify the header include and library include paths every time I want to rebuild the visual studio project files?

vinniefalco:
I tried this line with your changes:
```
cmake -G"Visual Studio 14 2015 Win64" .. -DBOOST_INCLUDEDIR=D:\lib\bo
ost_1_61_0 -DBOOST_LIBRARYDIR=d:\lib\boost_1_61_0\stage64
```

But the Additional Library Directories under Linker properties shows
```
D:/lib/boost_1_61_0/stage/lib;D:/lib/boost_1_61_0/stage/lib/$(Configuration);%(AdditionalLibraryDirectories)
```

e-fominov:
Try adding -DBoost_NO_SYSTEM_PATHS=ON

What is the version of CMake you are using? If your version < 3.6.2, it may not know about boost version 1.61. You can tell him that this version exists:

-DBoost_ADDITIONAL_VERSIONS=1.61

e-fominov:
I have just checked out master branch CMakeFiles.txt, and MSVC told me that he cant compile:

```
C:\work\Beast\include\beast/core/detail/ci_char_traits.hpp(11): fatal error C1083: Unable to open include file: boost/range/algorithm/equal.hpp: No such file or directory (compiling C:\work\Beast\test\http\parser_bench.cpp) [C:\work\Beast\b15\test\http\bench-tests.vcxproj]
...
```

Are you sure, it was compiling with previous version of CMakeFiles.txt?

vinniefalco:
I'm on CMake 3.6.0. I will try updating it.

Yes, I am sure that the tip of *master* compiles with CMake. I do all of my development in Visual Studio and I use CMake to generate the project files. You need to add the include path and library path to your Microsoft.Cpp.x64.user property sheet (View -> Property Manager).

vinniefalco:
Updating to CMake 3.6.2 had no effect on the result, I get the same errors. This is the command line I used:
```
cd bin64
rm -rf *
cmake -G"Visual Studio 14 2015 Win64" .. -DBOOST_INCLUDEDIR=D:\lib\boost_1_61_0 -DBOOST_LIBRARYDIR=d:\lib\boost_1_61_0\stage64
```
This is what the linker properties on the resulting project look like:
image

e-fominov:
I have restored original code for MSVC compilation process. MinGW compiles well on my pc
To compile MSVC I need to open .vcproj in Studio and set directories manually

vinniefalco:
This works for me now:
https://github.com/vinniefalco/Beast/commits/mingw

It will go into b14, so I'll close this. Thanks for your contribution!

PR #98, opened 2016-09-27
Added CMAKE instructions to find boost for MSVC in stage/stage64

e-fominov:
Tested with MinGW 6.2 and MSVC 2015 - both x64 and x86. Boost installed into stage and stage64 folders. Envoronment variable BOOST_ROOT set to boost sources folder
Boost wat built with commands:
```
b2 address-model=32 --build-type=minimal stage --stagedir=stage
b2 address-model=64 --build-type=minimal stage --stagedir=stage64
```
To build project simply run:
```
md build
cd build
cmake -G"Visual Studio 14 2015"
cmake --build .
```
same for Win64. No additional settings needed inside .vcproj files

Boost for MinGW was installed in default c:\boost folder
If not using environment variable BOOST_ROOT, its possible to add -DBOOST_ROOT=*boost_src_dir*

Also I have removed specific settings for NOT WIN32 in all examples/test and moved them into root CMakeLists.txt


e-fominov:
Looks like main branch is much ahead this PR. I can re-open PR for actual master branch. Do you need it?
And do you want to have "install" target for cmake?

vinniefalco:
I think we should re-open this issue, as a reminder that I need to revisit it later. I can't get to it this second but I agree that this needs to be looked at. I don't have an answers right now.

PR #105, opened 2016-09-30
b15

vinniefalco:
* rfc7230 section 3.3.2 compliance
* Add HTTPS example
* Add Secure WebSocket example
* Fix message_v1 constructor
* Tidy up DynamicBuffer requirements
* Tidy up error types and headers
* Fix handling empty HTTP headers in parser_v1


doc/makeqbk.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/sh
```
seelabs: This script should have execute attributes. Cherry pick this to change: https://github.com/seelabs/Beast/commit/a4d840607f7bcadd52f183c69891b7c50cdd8c1e

include/beast/http/impl/message_v1.ipp
```diff
@@ -139,9 +140,35 @@ prepare(message_v1<isRequest, Body, Headers>& msg,
...
+                    if((msg.status / 100 ) != 1 &&
+                        msg.status != 204 &&
+                        msg.status != 304)
+                    {
+                        msg.headers.insert("Content-Length",
+                            std::to_string(*pi.content_length));
+                    }
+                }
+            };
+            set_field{}(msg, pi);
```
seelabs: I like this idiom. It's an interesting way to give us overloaded local functions when we don't have `if constexpr` or overloaded lambdas. What do you think about renaming the struct to `overloads` and having static overloads called `set_field` to make it clear what the idiom is doing? Too verbose?
```            
            struct overloads
            {
                static
                void
                set_field(message_v1<true, Body, Headers>& msg,
                    detail::prepare_info const& pi)
                {
                    using beast::detail::ci_equal;
                    if(*pi.content_length > 0 ||
                        ci_equal(msg.method, "POST"))
                    {
                        msg.headers.insert("Content-Length",
                            std::to_string(*pi.content_length));
                    }
                }

                static
                void
                set_field(message_v1<false, Body, Headers>& msg,
                    detail::prepare_info const& pi)
                {
                    if((msg.status / 100 ) != 1 &&
                        msg.status != 204 &&
                        msg.status != 304)
                    {
                        msg.headers.insert("Content-Length",
                            std::to_string(*pi.content_length));
                    }
                }
            };
            overloads::set_field(msg, pi);
```

examples/ssl/Jamfile.v2
```diff
@@ -0,0 +1,54 @@
...
+else if [ os.name ] = NT
+{
+  lib ws2_32 ;
+  lib mswsock ;
+}
+else if [ os.name ] = HPUX
+{
+  lib ipv6 ;
+}
+else if [ os.name ] = HAIKU
```
seelabs: HAIKU seeps pretty obscure, but no harm.

vinniefalco: LOL... yeah... copied that from Asio

seelabs:
This commit adds cmake support for the ssl examples. I only tested on linux, and Scott S tested on Mac ( @scottschurr there have been changes, it needs to be re-tested): https://github.com/seelabs/Beast/commit/397507e0348ecd45821d101353bbe1b9dc8cf263

seelabs:
I am not able to build the ssl examples using `b2` on linux. If we can change to shared linking, the problem goes away.

I'm still learning bjam; I'll try to come up with a solution that let's us keep static linking, in the meantime, this commit allows me to compile with bjam: https://github.com/seelabs/Beast/commit/2c8bb10d81f66c82dc5efab7da3791396cac41c8

seelabs:
Reviewing my cmake commit, this is needed for it to work on windows: https://github.com/seelabs/Beast/commit/f761ec080cfebc815364f54475d93b9b8ec102cd

include/beast/http/impl/message_v1.ipp
vinniefalco: Actually this goes beyond generic lambdas. The implementation of the lambda is different depending on the type of argument. This technique allows the code to stay inline with the function. The alternative is to use a set of overloaded functions, resulting in non-local code.

examples/http_crawl.cpp
```diff
@@ -49,7 +49,7 @@ int main(int, char const*[])
             beast::http::read(sock, sb, res);
             std::cout << res;
         }
-        catch(boost::system::system_error const& ec)
+        catch(beast::system_error const& ec)
```
vinniefalco: The reason that beast defines these aliases is to support namespace injection. i.e. choose between `std::error_code` and `boost::system::error_code` at compile time.

doc/makeqbk.sh
vinniefalco: willfix, thanks

include/beast/http/impl/parse_error.ipp
```diff
@@ -0,0 +1,107 @@
...
+    message(int ev) const override
+    {
+        switch(static_cast<parse_error>(ev))
+        {
+        case parse_error::connection_closed: return "data after Connection close";
+        case parse_error::bad_method: return "bad method";
+        case parse_error::bad_uri: return "bad request-target";
+        case parse_error::bad_version: return "bad HTTP-Version";
+        case parse_error::bad_crlf: return "missing CRLF";
+        case parse_error::bad_request: return "bad reason-phrase";
```
vinniefalco: `bad_request` is unused. willfix

include/beast/http/read.hpp
```diff
@@ -44,7 +44,7 @@ namespace http {
     @param parser An object meeting the requirements of Parser
     which will receive the data.
 
-    @throws boost::system::system_error on failure.
+    @throws system_error on failure.
```
vinniefalco: willfix

include/beast/websocket/impl/error.ipp
```diff
@@ -5,20 +5,92 @@
...
+
+} // detail
 
 inline
 error_code
 make_error_code(error e)
 {
-    return error_code(
-        static_cast<int>(e), detail::get_error_category());
+    return error_code(static_cast<int>(e),
```
vinniefalco: NOICE!!

vinniefalco:
I'd like to see individual commits for CMake and bjam changes

vinniefalco:
SSL on CMake is looking good. bjam needs work though

seelabs:
Here's a cmake commit with better mac support: https://github.com/seelabs/Beast/commit/05abef04d0e22972588ca19bf1b9c65a5dbe0092

( I asked @scottschurr to this this; we should wait for a thumbs up from him before cherry picking this)

scottschurr:
I tested @seelabs cmake fix on OS X and it worked great.

scottschurr:
Looks good to me. :+1:

seelabs:
I spent a long time trying to get static linking working with ssl examples without success. My plan is to post a message to the boost developers list and ask for help. If we remove the `<link>static` from the Jamroot, then :+1: from me. Without that we'll need to wait for an answer on static linking before we merge this.

vinniefalco:
The SSL examples are not supposed to build when `b2` is invoked from the root

seelabs:
Unfortunately, no. Same kinds of error as when we just link `crypto` and `dl` as shared" `# error Must not define both BOOST_SYSTEM_DYN_LINK and BOOST_SYSTEM_STATIC_LINK`

vinniefalco:
Yeah. I removed `<link>static`

PR #127, opened 2016-10-13
Fix unused parameter warnings

awhatson:
Fixes warnings from gcc 5.4 (ubuntu xenial) with -Wall -Wextra.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ else()
     find_package(Threads)
 
     set(CMAKE_CXX_FLAGS
-      "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wpedantic")
+      "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Wpedantic")
 endif()
 
 if (APPLE AND NOT DEFINED ENV{OPENSSL_ROOT_DIR})
--- a/examples/http_sync_server.hpp
+++ b/examples/http_sync_server.hpp
@@ -11,6 +11,8 @@
 #include "file_body.hpp"
 #include "mime_type.hpp"
 
+#include <beast/http.hpp>
+#include <beast/core/placeholders.hpp>
 #include <beast/core/streambuf.hpp>
 #include <boost/asio.hpp>
 #include <cstdint>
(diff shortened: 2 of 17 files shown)


vinniefalco:
Thanks for submitting this! The reason that I leave parameters in, even if they are unused, is because the names show up in the documentation. And for private functions, the name provides a clue as to what the variable does. I will consider further increasing the warning levels though, and see what happens.

awhatson:
No worries, thanks for making this great library!

If you'd rather leave them in the signature, there are still a couple of options:
  a) "use" the parameter with a noop like (void)(someUnusedParam);
  b) mark them unused with boost::ignore_unused()

Unfortunately it looks like boost::ignore_unused was only added in boost 1.60, so you'd need to increase the required boost version, which I'd prefer you didn't do as ubuntu xenial ships with boost 1.58, and that's going to be around for a while.

Let me know what you think, I'm happy to amend the pull request.

vinniefalco:
Ah, `boost::ignore_unused` looks quite handy! There's no need to require 1.60, we can just borrow the "implementation" and make it a beast function (in a detail namespace):
```
template <typename... Ts>
BOOST_FORCEINLINE BOOST_CXX14_CONSTEXPR void ignore_unused(Ts const& ...)
{}

template <typename... Ts>
BOOST_FORCEINLINE BOOST_CXX14_CONSTEXPR void ignore_unused()
{}
```

Beast requires C++11 so we can keep it simple.

vinniefalco:
Should we add `/Wall` and `/Wextra` to the compilation flags?

awhatson:
I've added -Wextra for gcc only, and caught a few more missing includes from tests and examples.  I'm not in a position to test additional warnings on MSVC or OSX so those haven't been added, sorry.

I've also taken the liberty of fixing some missing includes, these weren't a problem in normal usage but popped up when building each header separately (eg. clang-complete or similar IDE functionality).

To find these I used the following:
`find -name '*.hpp' | xargs -n1 -P4 g++ -std=c++11 -Wall -Wextra -Wpedantic -Iinclude -Iextras`

To clean up all the pre-compiled headers this produces:
`find -name '*.gch' | xargs rm`

The same should be done for .ipp files as these are also producing errors when built on their own, however those are trickier to solve as it involves shuffling type definitions around - a bit too much for me at this stage.

awhatson:
Rebased on b16.

vinniefalco:
I made some adjustments - we don't want examples source code to include detail/ headers so I'm tweaking it.

PR #137, opened 2016-10-15
Improve websocket example in README.md

seeekr:
change call: to_string --> beast::to_string

My IDE at least was complaining about a plain `to_string` in the code, even though the compilation works fine. I don't understand C++ well enough yet to understand how the compiler manages to resolve `to_string` as `beast::to_string` all by itself, but at least my IDE (CLion) agrees with me (wrongly), so it might have merit to clean up that part of the sample!

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ int main()
     beast::websocket::opcode op;
     ws.read(op, sb);
     ws.close(beast::websocket::close_code::normal);
-    std::cout << to_string(sb.data()) << "\n";
+    std::cout << beast::to_string(sb.data()) << "\n";
 }
 ```
 
--- a/examples/websocket_example.cpp
+++ b/examples/websocket_example.cpp
@@ -31,5 +31,5 @@ int main()
     beast::websocket::opcode op;
     ws.read(op, sb);
     ws.close(beast::websocket::close_code::normal);
-    std::cout << to_string(sb.data()) << "\n";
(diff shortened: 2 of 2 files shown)


vinniefalco:
Hmm...that is quite interesting! I'm not sure how it compiles myself... @HowardHinnant ?

Anyway, I agree it should change. Can you please also make the change to the corresponding source file? This one:
https://github.com/vinniefalco/Beast/blob/master/examples/websocket_example.cpp

Thanks!

seeekr:
Amended commit, both README and example cpp should be in sync now!

seelabs:
@seeekr `beast::to_string` is found through ADL (see: http://en.cppreference.com/w/cpp/language/adl). Since `sb.data()` returns a type in the `beast` namespace the compiler knows to look for an overload in that namespace.

seeekr:
@seelabs Ha, that's pretty awesome. Though I imagine it's not really a good idea to depend on it for the situation that we have here, at least not in code targeted at non-expert C++ devs, potentially. But excellent to know how that works! Thanks!

seeekr:
closing as we're merging the two PRs as one

vinniefalco:
@seeekr I agree, the example code should be clear and concise.

HowardHinnant:
Fwiw, `to_string(sb.data())` compiles because `sb.data()` is a type defined in namespace `beast` and so ADL (Argument Dependent Lookup) searches namespace `beast` for a matching `to_string` signature.

PR #138, opened 2016-10-15
Improve websocket example

seeekr:
This one I'm not expecting to be merged, but I wanted to bring this up anyway:

Again, using CLion IDE, the IDE seems to get confused for whatever reason about the call to

`boost::asio::buffer("Hello, world!")`

and at least when we're talking about an example it might be a good thing to not trigger any error highlighting in popular IDE choices when we're expecting relative newcomers to jump in and try to load up and run a sample.

I'm not comfortable enough with C++ to tell if wrapping the `const char*` in a `std::string()` is adding any overhead (and as such might teach someone out there looking at this code a bad trick), and in any case more code is worse if less code will produce the same result, so I'm not sure if this should be merged.

Also, speaking of equal results: I noticed that while the compiler (clang++-3.8) compiles the example just fine as is, I'm getting a different output from running the program with vs without the `std::string()` wrapping -- some null character or so seems to be attached in the output in the example as is, while I'm getting no funny output in the wrapped version.

Again, I'm a relative newcomer (actually "back-comer") to C++ development, so maybe this is a complete non-issue to every normal C++ dev out there.

Cheers,
Denis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Relax ForwardIterator requirements in FieldSequence
 * Fix websocket failure testing
 * Refine Writer concept and fix exemplar in documentation
+* Improve websocket example
 
 API Changes:
 
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ int main()
     beast::websocket::opcode op;
     ws.read(op, sb);
     ws.close(beast::websocket::close_code::normal);
-    std::cout << to_string(sb.data()) << "\n";
+    std::cout << beast::to_string(sb.data()) << "\n";
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ int main()
     // WebSocket connect and send message using beast
     beast::websocket::stream<boost::asio::ip::tcp::socket&> ws{sock};
     ws.handshake(host, "/");
-    ws.write(boost::asio::buffer("Hello, world!"));
+    ws.write(boost::asio::buffer(std::string("Hello, world!")));
 
     // Receive WebSocket message, print and close using beast
     beast::streambuf sb;
--- a/examples/websocket_example.cpp
+++ b/examples/websocket_example.cpp
@@ -31,5 +31,5 @@ int main()
     beast::websocket::opcode op;
     ws.read(op, sb);
     ws.close(beast::websocket::close_code::normal);
-    std::cout << to_string(sb.data()) << "\n";
+    std::cout << beast::to_string(sb.data()) << "\n";
 }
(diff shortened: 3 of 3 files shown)


vinniefalco:
Oh, that is very interesting! `"Hello, world!"` is getting interpreted as an array (char[14]) and the call to `boost::asio::buffer` is seeing the null terminator (since its the last element in the array). I agree with your change, and I did not see the null output during development probably because my IDE and console windows simply ignore it.

But we also need to apply the change to the corresponding source file, its here:
https://github.com/vinniefalco/Beast/blob/master/examples/websocket_example.cpp

Thanks!

seeekr:
@vinniefalco Great! Happy to be of service then :)

I've amended the commit, so both files are in sync now.

vinniefalco:
Almost there... just a couple more things. Please merge the change from #137 into this commit, and update CHANGELOG.md. Or I can do it, up to you.

seeekr:
Rebased on top of #137, closed the other one.

seeekr:
Updated changelog as well. I think you can squash the commits as you're merging the PR, right? If not I can also squash them for you if you prefer, which I think you do, judging from the commit history and how CHANGELOG.md is managed!

vinniefalco:
Thank you for your contribution! It will go in to b17, and look like this:
https://github.com/vinniefalco/Beast/commit/010eb8d9b5cbc91b17b4533d99eee920c8941ae2

vinniefalco:
Looks good 👍

seeekr:
Cool! Yeah, that looks nice!

Thought b16 was representing the version currently being worked on as is often the custom with the version in the master branch :)

Nice! Thanks for the work on & publishing of this library! Looking forward to putting it into production in the coming weeks! :)

vinniefalco:
Oh hmm...whenever I merge to master, I place a commit at the tip which sets a new version. This way, the tip of master always has a unique version number. Is it done differently in other projects?

I appreciate the kind words.

seelabs:
:+1: Good changes

seeekr:
Ah, I see -- you're using the master branch as a place to keep the "latest released version", iiuc.
The projects I'm frequenting most, as well as in my own projects, master is usually being used as the development version, where any feature branch or similar is merged into master as it is completed, and then at some point a release branch is cut based on master that goes into polishing mode with only fixes being applied to it and used as the base to cut further point releases which are continuously shipped to the customer. While at the same time work keeps going on master as usual.
If that makes sense. So master is then potentially unstable at times, and eventually always branches out into release branches where most of the stabilization and maintenance work happens, being merged back regularly into master, of course.

vinniefalco:
Ah, I see what you mean. In the workflow we use, the **master** branch is always stable and always has a version at the tip. The **develop** branch gets batches of approved pull requests merged, and also has a version at the tip (but the version has b1, b2, etc... to represent that they are beta versions).

Well, that's how it will be after Beast has an official 1.0.0 release. Since Beast is still in beta, we use treat the **master** branch as if it is the **develop** branch. But the tip of **master** is still relatively stable. Things only get merged to master after they pass review, tests, CI, and local inspection.

vinniefalco:
This will go into **1.0.0-b17**

PR #141, opened 2016-10-15
Changed implicit_value to default_value

e-fominov:
This PR changes the usage of boost::program_options in http_server example producing simplier code

Here is an explanation of the difference: http://stackoverflow.com/questions/2829806/what-is-the-difference-between-default-value-and-implicit-value-in-boostprogra


vinniefalco:
Very nice! I don't know program options that well.

vinniefalco:
Reviewers: @miguelportilla @wilsonianb

vinniefalco:
This is part of #146

Issue #132, opened 2016-10-14
Partial message reading


e-fominov:
When making a server side, I want have two features:
1. read message headers, analyze them and may be close connection before reading the body. This situation is similar to 100-continue (#123), but with different use-case
2. read message body partially block-by-block when it is arriving in socket

The use case is client sending very large file 2+ GB. If the file is too-large, I want to close connection with no reading body at all. Yes, client has many ways on how to split file, but If it does not split it - I want to server be alive when processing this request
If file is large, but not much, I dont want to allocate 2GB per connection, but I want to have small buffer, about 64 kb that will be used to read parts of message body

I can solve this situation partially, by making some custom message body class and use it in `request_v1<my_custom_body>` with custom `reader` class,  but I want to close connection before allocating a buffers if something is wrong in headers, but headers syntax is correct

vinniefalco:
I agree with everything you said. Another use-case is to read the headers, and then choose the type of body depending on the headers. I am working on a comprehensive solution.

For now, if you want to abort the connection on bad headers, a work-around is to derive your own parser (make a copy of `parser_v1` and then change `on_headers`), and use `beast::http::async_parse` instead of `async_read`. I can help with that since there aren't any examples. In your parser, you can perform your checks in `on_headers` and set the error code:
https://github.com/vinniefalco/Beast/blob/master/include/beast/http/parser_v1.hpp#L218

vinniefalco:
The **Reader** should not be constructed until the headers are known

vinniefalco:
@e-fominov How about something like this:
```
/** Read a message from the stream and process it
*/
template<bool isRequest, class Stream, class DynamicBuffer>
void process_message(Stream& stream, DynamicBuffer& rb)
{
  using namespace beast::http;

  // A header parser processes just the request/response line
  // and headers, and doesn't need to know the Body type.
  //
  header_parser_v1<isRequest, headers> p0;

  // Parse the request/response line and headers.
  // http://vinniefalco.github.io/beast/beast/ref/http__parse/overload1.html
  //
  parse(stream, rb, p0);

  // The header parser owns the message headers, you can read
  // them using p0.get() or you can take ownership by calling p0.release().
  //
  message_headers<isRequest, headers> const& m0 = p0.get();
  // do any processing on m0...

  // This function associates a Body type with the parser and
  // returns a new parser which inherits the state machine of p0
  //
  auto p = with_body<string_body>(std::move(p0), ...);

[...]

PR #146, opened 2016-10-17
1.0.0-b17

vinniefalco:
* Tidy up some declarations
* Fix basic_streambuf::capacity
* Add basic_streambuf::alloc_size
* Parser callbacks may not throw
* Fix Reader concept doc typo
* Add is_Reader trait
* Tidy up basic_headers for documentation
* Tidy up documentation
* Add basic_parser_v1::reset
* Fix handling of body_what::pause in basic_parser_v1

API Changes:

* Added init() to Reader requirements
* Reader must be nothrow constructible
* Reader is now constructed right before reading the body
  - The message passed on construction is filled in
* Rework HTTP concepts:
  - Writer uses write instead of operator()
  - Refactor traits to use void_t
  - Remove is_ReadableBody, is_WritableBody
  - Add has_reader, has_writer, is_Reader, is_Writer
  - More friendly compile errors on failed concept checks
* basic_parser_v1 requires all callbacks present
* on_headers parser callback now returns void
* on_body_what is a new required parser callback returning body_what


vinniefalco:
@e-fominov Please look at these changes, especially `headers_parser`. There's a unit test in **headers_parser.cpp** which shows how it is used. Its what we talked about in that other issue.

vinniefalco:
This is now possible with this branch:
```
/** Read a message from the stream and process it
*/
template<bool isRequest, class Stream, class DynamicBuffer>
void process_message(Stream& stream, DynamicBuffer& rb)
{
  using namespace beast::http;

  // A header parser processes just the request/response line
  // and headers, and doesn't need to know the Body type.
  //
  header_parser_v1<isRequest, headers> p0;

  // Parse the request/response line and headers.
  // http://vinniefalco.github.io/beast/beast/ref/http__parse/overload1.html
  //
  parse(stream, rb, p0);

  // The header parser owns the message headers, you can read
  // them using p0.get() or you can take ownership by calling p0.release().
  //
  message_headers<isRequest, headers> const& m0 = p0.get();

  // Do any processing on m0...

  // This function associates a Body type with the parser and
  // returns a new parser which inherits the state machine of p0
  //
  auto p = with_body<string_body>(std::move(p0), ...);

  // Continue parsing the message body into p,
  // now that we have chosen a the body type.
  parse(stream, p);

  // Complete message is available now, take ownership:
  message<isRequest, string_body, headers> m = p.release();

  // Do something with m
}
```

vinniefalco:
Reviewers: @JoelKatz @wilsonianb @bachase

vinniefalco:
Documentation preview:
http://vinniefalco.github.io/beast/develop

include/beast/http/concepts.hpp
```diff
@@ -18,63 +21,107 @@ namespace http {
...
+        std::declval<U>().init(std::declval<error_code&>()),
+            std::true_type{})>
+    static R check1(int);
+    template<class>
+    static std::false_type check1(...);
+    using type1 = decltype(check1<T>(0));
+
+    // VFALCO This is unfortunate, we have to provide the template
+    //        argument type because this is not a deduced context?
+    //
```
bachase: Anything in [Boost.TTI](http://www.boost.org/doc/libs/1_62_0/libs/tti/doc/html/index.html) that would help?

include/beast/http/impl/write.ipp
```diff
@@ -527,11 +538,16 @@ write(SyncWriteStream& stream,
     wp.sb.consume(wp.sb.size());
     if(! result)
     {
+        detail::writef_lambda<SyncWriteStream> wf{
```
bachase: How come you ended up not using this hoisted write lambda?

include/beast/http/message.hpp
```diff
@@ -18,41 +18,174 @@
...
+        of the headers member.
+    */
+    /** @{ */
+    request_headers() = default;
+
+    template<class Arg1, class... ArgN,
+        class = typename std::enable_if<
+            (sizeof...(ArgN) > 0) || ! std::is_convertible<
+                typename std::decay<Arg1>::type,
+                    request_headers>::value>::type>
```
bachase: cool!

test/http/message.cpp
```diff
@@ -147,9 +146,154 @@ class message_test : public beast::unit_test::suite
...
+            request_headers<MoveHeaders> r{std::move(h)};
+            BEAST_EXPECT(h.moved_from);
+            BEAST_EXPECT(r.headers.moved_to);
+            request<string_body, MoveHeaders> m{std::move(r)};
+            BEAST_EXPECT(r.headers.moved_from);
+            BEAST_EXPECT(m.headers.moved_to);
+        }
+
+        {
+            MoveHeaders h;
```
bachase: This scope looks identical to the one above starting in L185.  Am I missing something?

include/beast/http/concepts.hpp
vinniefalco: I don't think its possible in the current C++ language

include/beast/core/basic_streambuf.hpp
```diff
@@ -209,6 +209,36 @@ class basic_streambuf
...
+        The size of the allocation can influence capacity, which will
+        affect algorithms that use capacity to efficiently read from
+        streams.
+
+        @note This will not affect any already-existing allocations.
+
+        @param n The number of bytes.
+    */
+    void
+    alloc_size(std::size_t n)
```
vinniefalco: It should be in there, basic_streambuf.cpp

include/beast/http/concepts.hpp
vinniefalco: See boost-dev mailing list.
From "[boost] How can I detect if a template static member function exists?":

_It is not possible AFAIK. The tti library can detect whether a class template is a member of a type but it cannot detect whether a function template is a member of a type. The reason for this, within tti, is that class templates and function templates are not treated orthogonally in C++; a class template can be a template parameter but a function template cannot be a template parameter. If a function template could be a template parameter then tti could detect a function template as a member of a type, whether that function template were a static or a non-static member._

_I have felt that the lack of orthogonality in C++, where a class template can be a template parameter but a function template cannot be a template parameter, is a design weakness of C++. But I would have to convince the C++ standards committee of the practicality of allowing a function template to be a template parameter to possibly change this.  I doubt whether saying that the change would allow the Boost tti library to detect function templates in the same way it can detect class templates, would be a good enough practical reason <g>._

include/beast/core/basic_streambuf.hpp
bachase: D'oh.  For others, it is [here](https://github.com/vinniefalco/Beast/pull/146/files/ebebe52612fc9a5f754ab4caf5ddb45b97b8f599#diff-9f4842e13a767b52e0c1c9e0437e6f1bR462)

vinniefalco:
Top 4 commits are new

bachase:
New commits look good to me.

bachase:
Accidentally closed

vinniefalco:
New doc preview URL
http://vinniefalco.github.io/stage/beast/index.html

include/beast/http/message.hpp
vinniefalco: Couldn't have done it with Howard!

doc/quickref.xml
```diff
@@ -60,12 +59,33 @@
             <member><link linkend="beast.ref.http__async_parse">async_parse</link></member>
             <member><link linkend="beast.ref.http__async_read">async_read</link></member>
             <member><link linkend="beast.ref.http__async_write">async_write</link></member>
+            <member><link linkend="beast.ref.http__async_write">async_write</link></member>
```
vinniefalco: copypasta, will fix

doc/quickref.xml
```diff
@@ -60,12 +59,33 @@
...
             <member><link linkend="beast.ref.http__swap">swap</link></member>
+            <member><link linkend="beast.ref.http__with_body">with_body</link></member>
             <member><link linkend="beast.ref.http__write">write</link></member>
           </simplelist>
+          <bridgehead renderas="sect3">Type Traits</bridgehead>
+          <simplelist type="vert" columns="1">
+           <member><link linkend="beast.ref.http__is_Body">is_Body</link></member>
+           <member><link linkend="beast.ref.http__is_Parser">is_Parser</link></member>
+           <member><link linkend="beast.ref.http__is_Reader">is_Reader</link></member>
+           <member><link linkend="beast.ref.http__is_Writer">is_Reader</link></member>
```
vinniefalco: copypasta, will fix

test/http/read.cpp
```diff
@@ -96,6 +151,20 @@ class read_test
...
+                sb.prepare(len), buffer(s, len)));
+            test::fail_counter fc{n};
+            test::string_stream ss{ios_, s};
+            parser_v1<isRequest, fail_body, headers> p{fc};
+            error_code ec;
+            parse(ss, sb, p, ec);
+            if(! ec)
+                break;
+        }
+        BEAST_EXPECT(n < limit);
```
vinniefalco: You are proposing that if the parse completes successfully on the first iteration through the loop, that it should be considered a test failure?

The purpose of this loop is to generate an error on the n-1th I/O operation. And it keeps going until the parse is successful. This needs to happen in some bounded number of I/Os, and that's what the test is testing.

Thinking about your idea, a way to improve the test is to take the parsed result and convert it back to a string, then compare against the input to make sure that the result of the parse matches. I can open an issue for that. Note that the purpose of this code is to increase test coverage by exercising all the failure branches, which usually never get tested.

include/beast/http/concepts.hpp
```diff
@@ -116,31 +163,97 @@ class is_Parser
...
-using is_Body = typename detail::is_Body<T>::type;
+using is_Body = detail::has_value_type<T>;
 #endif
 
-/// Determine if `T` meets the requirements of `ReadableBody`.
-template<class T>
+/** Determine if a @b Body has a nested type `reader`.
+
+    @tparam T The type to check, which must meet the
+    requirements of @b Body.
```
vinniefalco: Yeah, that's unfortunately a bug in the xsl. See:
https://github.com/vinniefalco/docca/issues/8

include/beast/http/parser_v1.hpp
```diff
@@ -150,7 +152,7 @@ class parser_v1
 
     /** Returns the parsed message.
 
-        Only valid if `complete()` would return `true`.
+        Only valid if @ref complete would return `true`.
```
vinniefalco: Yeah, that's because `complete` is implemented in the base class `basic_parser_v1`. If anything, this is a bug or missing feature in the xsl transform. The information is there, in the Doxygen XML output, we just need to figure out a way to format it and insert into the docs something like "inherited from base class ____".

test/http/headers_parser_v1.cpp
```diff
@@ -0,0 +1,62 @@
...
+                "GET / HTTP/1.1\r\n"
+                "User-Agent: test\r\n"
+                "Content-Length: 5\r\n"
+                "\r\n"
+                "*****"
+                ), ec);
+            BEAST_EXPECT(n == 55);
+            BEAST_EXPECTS(! ec, ec.message());
+            BEAST_EXPECT(p.complete());
+        }
```
vinniefalco: will fix

vinniefalco:
@bachase @wilsonianb I'd like to hear your thoughts on the headers parser interface and if you feel it elegantly solves the problem of reading the message in two stages.

include/beast/http/basic_parser_v1.hpp
```diff
@@ -140,71 +140,81 @@ static std::uint64_t constexpr no_content_length =
...
+        void on_complete(error_code&) {}
+    };
+    @endcode
 
-    The parser uses traits to determine if the callback is possible.
-    If the @b Derived type omits one or more callbacks, they are simply
-    skipped with no compilation error. The default behavior of `on_body`
-    when the derived class does not provide the member, is to specify that
-    the body should not be skipped.
+    The return value of `on_body_what` is special, it controls
```
nbougalis: The comma after 'special' is a bit weird. A colon would read better.

> The return value of `on_body_what` is special: it controls whether or not the parser should expect a body.

vinniefalco: You are literally asking me to move one pixel a bit lower

include/beast/http/impl/basic_parser_v1.ipp
```diff
@@ -913,9 +989,9 @@ write(boost::asio::const_buffer const& buffer, error_code& ec)
                 flags_ |= parse_flag::skipbody;
                 break;
             case body_what::pause:
-                s_ = s_headers_done;
                 return used();
             }
+            --p;
```
wilsonianb: What's the point of this decrement and the prior increment in `case s_headers_almost_done:`?
Could they both be removed?

include/beast/http/concepts.hpp
vinniefalco: `with_body` is a function, while the other one is on a class. The xsl sheet has two different templates to handle that.

include/beast/http/impl/basic_parser_v1.ipp
vinniefalco: They are necessary. It has to do with how the parser transitions state and when it consumes a character (by executing `break`) versus when it changes state without advancing the input pointer (by executing `goto redo`). Worth studying to make sure I didn't screw it up...

include/beast/http/impl/write.ipp
bachase: Looking at the tip of develop, I still see the `#if 0` block. https://github.com/vinniefalco/Beast/blob/develop/include/beast/http/impl/write.ipp#L545-L550

bachase:
I'm unable to reply to your comment on the `header_parser` design.  I'm not as familiar with real world examples doing that separate parsing, but I think it address the case in #132 nicely.

vinniefalco:
All issues addressed, rebased

PR #152, opened 2016-10-24
Add ZLib module

vinniefalco:
The zlib module is a port of ZLib's raw deflate algorithm to header-only C++11.


vinniefalco:
Reviewers: @JoelKatz @seelabs

seelabs:
Great work on this. Lots of unneeded code went away, macros removed, reduced memory allocations, added enums & classes, and added error codes. All of those "little" changes add up to a vastly improved code base.

I reviewed this by reviewing the many commits on the on zlib-19 branch and confirming the transformations did not introduce errors. I could not find any problems. However, there were a _lot_ of changes and some of the changes were hard to review (things that moved lots of code around). My biggest recommendation is to add testing against the "standard" zlib library: for example: compress with that zlib library and decompress with the beast zlib.

I did not look for nits (it would take too long to comb through such a large PR for nits), however a few did pop out while I reviewed. Here's a commit with couple of nits I found: https://github.com/seelabs/Beast/commit/797778d962fe5a02a2cb68608f438b84992d2c67

Note: CI reports a warning from gcc (I don't see this warning locally):
`deflate_stream.hpp:1048:10: warning: ‘*((void*)& old_flush +4)’ may be used uninitialized in this function [-Wmaybe-uninitialized]` Warning looks like a false alarm to me, not sure what to do about that one.

:+1:

vinniefalco:
>add testing against the "standard" zlib library: for example: compress with that zlib library and decompress with the beast zlib.

You mean like this?
https://github.com/vinniefalco/Beast/blob/zlib/test/zlib/inflate_stream.cpp#L311

seelabs:
@vinniefalco Ug, yes <embarrassed>.

vinniefalco:
>Note: CI reports a warning from gcc (I don't see this warning locally):

This is something to do with `boost::optional`. I have shuffled it around a few times to make the warning go away, to no avail. I don't like the warning, I'd like to get rid of it. Maybe we can analyze it and use a bool + int instead of optional.

vinniefalco:
> I did not look for nits

Well the original code is like one giant nit

JoelKatz:
👍 This is awesome.

vinniefalco:
Part of #168

PR #161, opened 2016-10-29
Override incremental link flags:

mellery451:
Disable incremental linking for RelWithDebInfo builds on MSVC.


vinniefalco:
Part of #164

PR #168, opened 2016-11-04
B19

vinniefalco:
Optimize mask and utf8 checking

ZLib is already passed


HowardHinnant:
In `testOneByteSequence()` why is [194, 244] not checked for being invalid?

include/beast/websocket/detail/mask.hpp
```diff
@@ -105,73 +105,119 @@ ror(T t, unsigned n = 1)
...
+    // Mask 4 bytes at a time
     for(auto i = n / sizeof(key); i; --i)
     {
-        *p ^=  key      ; ++p;
-        *p ^= (key >> 8); ++p;
-        *p ^= (key >>16); ++p;
-        *p ^= (key >>24); ++p;
+        *reinterpret_cast<
+            std::uint32_t*>(p) ^= key;
+        p += sizeof(key);
```
miguelportilla: This may be faster:
```
    {
        // Mask 4 bytes at a time
        auto pp = reinterpret_cast<std::uint32_t*>(p);
        for(auto i = n / sizeof(key); i; --i)
            *pp++ ^= key;
    }
```

JoelKatz:
:neckbeard::+1:

HowardHinnant:
👍  Passes all tests on macOS.  Additionally I wrote an exhaustive hello world test for check_utf8:  I ran every single 1, 2, 3 and 4 byte sequence of bits through it, and compared it to the result returned from `std::codecvt_utf8<char32_t>` and it all checked out.  I don't recommend putting that test into beast as it takes about 15min to run.  But running it once gives pretty high confidence that the code is correct.

miguelportilla:
@HowardHinnant A unit test for the utf8 checker class that checks all possible code points is in the project: https://github.com/vinniefalco/Beast/blob/master/test/websocket/utf8_checker.cpp

miguelportilla:
👍

include/beast/websocket/detail/mask.hpp
miguelportilla: Turns out to be identical in speed. Nothing to see here folks, please move along.

vinniefalco:
Spoke too soon, there's a build failure on CI

PR #191, opened 2016-11-16
permessage-deflate extension for WebSocket

vinniefalco:
This adds support for the permessage-deflate extension, documented here:
https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-00

Documentation preview:
http://vinniefalco.github.io/stage/beast/b21/index.html

Autobahn|Testsuite results:
http://vinniefalco.github.io/stage/autobahn/

Reviewers: @HowardHinnant @ximinez


HowardHinnant:
I'm just now noticing a waring coming out of test/zlib/zlib-1.2.8/inflate.c:1507:
````
 warning: shifting a negative signed value is undefined [-Wshift-negative-value]
    if (strm == Z_NULL || strm->state == Z_NULL) return -1L << 16;
                                                        ~~~ ^
````
I think this should be taken seriously, and it is easily fixed:
````
    if (strm == Z_NULL || strm->state == Z_NULL) return static_cast<long>(~0UL << 16);
````
I've tested this on all three compilers (online) and I'm seeing no warnings, and the correct constant, both for 32bit and 64bit long.

vinniefalco:
The problem is then I am modifying the ZLib source code. If anything, that change needs to be submitted to the upstream. And note, the file is compiled with the C compiler.

HowardHinnant:
Understood.  I don't think the ZLib source code is trustworthy without this fix.  Today's C++ compilers aggressively optimize code into non-intuitive behaviors in situations like this.  There is a very real possibility that any part of this TU (and the TU is even bigger with unity builds) does not do what is intended.

That is, my recommendation is either to do whatever it takes to fix this, or not use the ZLib code.  A C-style cast to long instead of static_cast would be acceptable.

HowardHinnant:
Here's a simpler fix that doesn't require a cast:
````
 if (strm == Z_NULL || strm->state == Z_NULL) return ~0xFFFFL;
````

vinniefalco:
hmm... looks like it was fixed:
https://github.com/madler/zlib/issues/111

HowardHinnant:
Their fix looks like a rant (one which I agree with) :-).  At least they are aware of the issue.

I ran a godbolt test and was unable to find a compiler/version that optimized based on the theory that UB can not happen (which is good).

HowardHinnant:
`threading-multi/zlib-tests` has been running for 16min now...

vinniefalco:
Dealing with zlib test failures, will re-open

PR #198, opened 2016-11-23
Beast install target and debian package generator.

avalchev:
Additional cmake configuration to properly generate Beast package libbeast-dev on Debian (RPM should be easy too).

Example usage with cmake:

```
find_package(Beast REQUIRED)
include_directories(${Beast_INCLUDE_DIR})
```

Also cmake sets variable Beast_FOUND properly.

Please notice that package maintainer and contact are set to my info. Feel free to change these.


CMakeLists.txt
```diff
@@ -116,3 +124,77 @@ add_subdirectory (test/core)
...
+  set(CPACK_GENERATOR "DEB")
+  set(CPACK_SOURCE_GENERATOR "DEB")
+  set(CPACK_PACKAGE_VENDOR "syslandscape")
+  set(CPACK_PACKAGE_CONTACT "aleksandar.valchev@gmail.com")
+
+  set(CPACK_DEBIAN_ARCHITECTURE all)
+  set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Aleksandar Valchev")
+  set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
+  set(CPACK_DEBIAN_PACKAGE_DEPENDS
+    "libc6 (>= 2.3) , libssl-dev (>=1.0.2) , libboost-coroutine-dev , libboost-context-dev , libboost-thread-dev , libboost-filesystem-dev , libboost-program-options-dev "
```
djarek: Is there a way specify optional dependencies? I'm quite sure that libssl-dev and libboost-coroutine-dev aren't necessary.

avalchev: Actually boost dependencies are not optional: 
[find_package(Boost REQUIRED COMPONENTS coroutine context thread filesystem program_options system)
](https://github.com/vinniefalco/Beast/blob/master/CMakeLists.txt#L28)

As far as I know with Linux packaging systems there are not `optional` dependencies. But you can "suggest":

> CPACK_DEBIAN_PACKAGE_SUGGESTS -  Allows packages to declare a suggested package install grouping.

Two reasons I made `libssl-dev` mandatory:

- I never install suggested packages.  
- A lot of questions from new users may emerge about SSL support.
- `libssl-dev` is very common package for all Linux distributions, so there is no chance to be missing.

For me it's not a problem to move `libssl-dev` to "Suggested" packages.

Please advice how to proceed.

djarek: I guess then it's fine to keep these as dependencies.

vinniefalco:
I appreciate the contribution, but there's no way for me to maintain this since I don't build packages. And when Beast becomes part of Boost, well then this will be obsolete. I guess if Beast fails to make it into Boost then we could revisit it but for now I am regretfully closing this.

Issue #215, opened 2016-12-27
Implement the Asio deallocation-before-invocation guarantee


harrishancock:
### TL; DR

Great library! I love it. However, not deallocating all memory associated with a handler before making the upcall to that handler is problematic, as it prevents users from implementing composed operations on top of Beast's composed operations without knowing which invoked handlers are safe to move and which are not. [You already tripped over one such bug, it looks like](https://github.com/vinniefalco/Beast/blob/master/include/beast/websocket/impl/accept.ipp#L261).

There are a couple workarounds (make users copy their invoked handlers, or equivalently make copies of upcall handlers before invoking them, or just defer upcalls with `ios.post()`), but a full, performant fix is challenging. I have a few suggestions at the end.

### The deallocation-before-invocation guarantee

Beast supports the same custom memory allocation strategy as Asio, using the `asio_handler_allocate` and `asio_handler_deallocate` hooks. However, it does not implement the same [guarantee that Asio does](http://www.boost.org/doc/libs/1_62_0/doc/html/boost_asio/overview/core/allocation.html), namely:

> The implementation guarantees that the **deallocation will occur before the associated handler is invoked**, which means the memory is ready to be reused for any new asynchronous operations started by the handler.

And reiterated [here](http://www.boost.org/doc/libs/1_62_0/doc/html/boost_asio/reference/asynchronous_operations.html):

> If the implementation needs to allocate storage for an asynchronous operation, the implementation will ... **deallocate the storage prior to the invocation of the handler** via asio_handler_invoke.

I will refer to this guarantee as the "deallocation-before-invocation guarantee."

### Beast does not implement this guarantee

Beast's composed operations are implemented using stateful handlers (`accept_op`, `response_op`, `write_op`, etc.), which are lightweight wrappers around a `shared_ptr<data>`, where `data` is the type containing the operation's state. The `shared_ptr<data>` object is allocated using the Asio custom memory allocation strategy.

Since the handler `h` passed to the operation's initiating function (I will refer to this as its "upcall handler") lives in the `data` object, `h` is therefore stored in memory that was allocated using its own allocation strategy. In order to implement Asio's guarantee when invoking `h`, `h` would have to be moved/copied out of the `data` object to a new object `h'`, the `data` object destroyed and deallocated, then `h'` finally invoked.

[...]

PR #218, opened 2017-01-04
1.0.0-b22

vinniefalco:
* Fix broken Intellisense (MSVC)
* Add missing include
* Implement asio dealloc-before-invoke guarantee

Doc preview:
http://vinniefalco.github.io/stage/beast/b22/index.html

Reviewers @nbougalis @HowardHinnant @JoelKatz


include/beast/core/impl/handler_ptr.ipp
```diff
@@ -0,0 +1,126 @@
...
+void
+handler_ptr<T>::
+release()
+{
+    BOOST_ASSERT(p_);
+    if(--p_->n)
+        return;
+    if(p_->t)
+    {
+        // VFALCO This should be a move instead
```
harrishancock: I'm curious, why isn't it?

include/beast/core/handler_ptr.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+    This is a smart pointer that retains shared ownership of an
+    object through a pointer. Memory is managed using the allocation
+    and deallocation functions associated with a completion handler,
+    which is also stored in the object. The object is destroyed and
+    its memory deallocated when one of the following happens:
+
+    * The last remaining container owning the object is destroyed
+    * The last remaining container owning the object is assigned
+      another object via @ref operator= or @ref reset.
```
harrishancock: Remove this line? It looks like both `operator=` and `reset` functions were removed.

include/beast/core/impl/handler_ptr.ipp
vinniefalco: The destructor of `T` may need the handler. Example:
https://github.com/vinniefalco/Beast/blob/master/include/beast/websocket/impl/write.ipp#L152

This code is not compatible with Networking-TS since it requires **CopyConstructible**. Changing it to a move will require an overhaul and certainly a change to the `handler_ptr` interface. I'm experimenting with such changes presently.

include/beast/core/handler_ptr.hpp
vinniefalco: True

include/beast/core/impl/handler_ptr.ipp
harrishancock: Ouch, I see.

You're probably already aware, but the Net TS replaces the allocation hooks with a single `get_associated_allocator(handler)` function. This should have the benefit that the forwarding chain only has to be ascended a single time, at operation initiation time, rather than once for every allocation and once for every deallocation. This might mitigate the problem, because it ideally would return an allocator that won't depend on any individual handler participating in the forwarding chain, save perhaps the top-most.

vinniefalco: Didn't know that, reading up now

vinniefalco: Too much for me to think about right now! I think its best to just make sure that Beast works with Boost.Asio as written, and worry about Net TS when it is closer to becoming part of the standard library on the big 3 compilers.

vinniefalco:
>...if a handler's copy constructor throws, bad things happen: either a std::terminate in handler_ptr's destructor or a resource leak in its invoke function. I

This looks like a problem, not sure what the solution is

vinniefalco:
>...if a handler's copy constructor throws, bad things happen: either a std::terminate in handler_ptr's destructor or a resource leak in its invoke function.

Working on a fix for this. I have a solution, but it involves some change.

include/beast/core/impl/handler_ptr.ipp
harrishancock: I concur, that's why I didn't bring it up in the original #215 issue. Boost.Asio is the important target.

vinniefalco:
I *think* I fixed it all...I rewrote the branch so please review the code again. The implementation for `handler_ptr` has changed significantly.

include/beast/core/handler_ptr.hpp
```diff
@@ -0,0 +1,176 @@
...
+/** A smart pointer container.
+
+    This is a smart pointer that retains shared ownership of an
+    object through a pointer. Memory is managed using the allocation
+    and deallocation functions associated with a completion handler,
+    which is also stored in the object. The object is destroyed and
+    its memory deallocated when either of the following happens:
+
+    * The last remaining container owning the object is destroyed
+    * The function @ref invoke is called.
```
harrishancock: And also now the function `release_handler`.

include/beast/core/impl/handler_ptr.ipp
```diff
@@ -0,0 +1,130 @@
...
+invoke(Args&&... args)
+{
+    BOOST_ASSERT(p_);
+    BOOST_ASSERT(p_->t);
+    p_->t->~T();
+    boost_asio_handler_alloc_helpers::
+        deallocate(p_->t, sizeof(T), p_->handler);
+    p_->t = nullptr;
+    p_->handler(std::forward<Args>(args)...);
+}
```
harrishancock: Consider implementing this function as `release_handler()(std::forward<Args>(args)...)`. The only difference between the two expressions that I can think of is that `invoke` leaves the handler in place, while `release_handler()(...)` would zap the handler by moving it into its return value, but since both functions first destroy/deallocate `data`, I don't think that'd be a problem. I suppose the difference between the two could be removed by making `release_handler` return an rvalue reference, allowing the recipient of the ref to make the decision to zap or not, but I'm not sure how palatable/correct that is.

include/beast/core/handler_ptr.hpp
vinniefalco: lol...

vinniefalco: Will fix

HowardHinnant:
👍

include/beast/core/handler_ptr.hpp
```diff
@@ -0,0 +1,177 @@
...
+/** A smart pointer container.
+
+    This is a smart pointer that retains shared ownership of an
+    object through a pointer. Memory is managed using the allocation
+    and deallocation functions associated with a completion handler,
+    which is also stored in the object. The object is destroyed and
+    its memory deallocated when one of the following happens:
+
+    * The function @ref invoke is called.
+    * The function @release_handler is called
```
harrishancock: Sorry Vinnie, should be `@ref release_handler`.

The Markdown-style `*` bullets should be `@li` instead, I think. They're not rendered correctly in the docs: http://vinniefalco.github.io/stage/beast/b22/beast/ref/handler_ptr.html

vinniefalco: Will fix

vinniefalco:
>This is almost elegant.

Its got a couple of warts, none of the terminal. The biggest one is that it allocates `sizeof(std::std::uint16_t) + sizeof(Handler)` using the regular heap. We might be able to do better than this in [net.ts] which only requires that handlers are movable.

include/beast/core/impl/handler_ptr.ipp
```diff
@@ -0,0 +1,138 @@
...
+{
+}
+
+template<class T, class Handler>
+handler_ptr<T, Handler>::
+~handler_ptr()
+{
+    if(! p_)
+        return;
+    if(--p_->n)
```
nbougalis: Shouldn't that this be an `std::atomic`? Are we always guaranteed that this will only be invoked in a strand?

vinniefalco: `n` is `std::atomic<std::uint16_t>`

include/beast/core/handler_helpers.hpp
```diff
@@ -0,0 +1,104 @@
...
+    using boost::asio::asio_handler_invoke;
+    asio_handler_invoke(function, std::addressof(handler));
+#endif
+}
+
+/// Invoke function for handlers.
+template<class Function, class Handler>
+inline
+void
+invoke(Function const& function, Handler& handler)
```
vinniefalco: I'm not sure. I just copied the code from Boost.Asio:
https://github.com/boostorg/asio/blob/develop/include/boost/asio/detail/handler_invoke_helpers.hpp#L29

include/beast/core/impl/handler_ptr.ipp
nbougalis: Huh, you're right. I double-checked before posting this too... I was sure it wasn't.

PR #223, opened 2017-01-11
1.0.0-b23: Add permessage-deflate WebSocket extension

ximinez:
Subsumes #193, fixes CI build issue (by disabling autobahn testsuite), and rebased on 1.0.0-b22.

### DIFF
--- a/include/beast/websocket/detail/pmd_extension.hpp
+++ b/include/beast/websocket/detail/pmd_extension.hpp
@@ -1,0 +1,467 @@
+//
+// Copyright (c) 2013-2016 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef BEAST_WEBSOCKET_DETAIL_PMD_EXTENSION_HPP
+#define BEAST_WEBSOCKET_DETAIL_PMD_EXTENSION_HPP
+
+#include <beast/core/error.hpp>
+#include <beast/core/consuming_buffers.hpp>
+#include <beast/core/detail/ci_char_traits.hpp>
+#include <beast/zlib/deflate_stream.hpp>
+#include <beast/zlib/inflate_stream.hpp>
+#include <beast/websocket/option.hpp>
+#include <beast/http/rfc7230.hpp>
+#include <boost/asio/buffer.hpp>
+#include <utility>
+
+namespace beast {
+namespace websocket {
+namespace detail {
+
+// permessage-deflate offer parameters
+//
+// "context takeover" means:
+// preserve sliding window across messages
+//
+struct pmd_offer
+{
+    bool accept;
+
+    // 0 = absent, or 8..15
+    int server_max_window_bits;
+
+    // -1 = present, 0 = absent, or 8..15
+    int client_max_window_bits;
+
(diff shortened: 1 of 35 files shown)


vinniefalco:
You might as well put a signed version commit on it ^_^

include/beast/websocket/detail/pmd_extension.hpp
```diff
@@ -0,0 +1,467 @@
...
+}
+
+// Set permessage-deflate fields for a client offer
+//
+template<class Fields>
+void
+pmd_write(Fields& fields, pmd_offer const& offer)
+{
+    std::string s;
+    s = "permessage-deflate";
```
HowardHinnant: Why not construct `s` with this constant?

vinniefalco: Because Feng Shui

include/beast/websocket/detail/pmd_extension.hpp
```diff
@@ -0,0 +1,467 @@
...
+        dynabuf.commit(zs.total_out);
+        zs.total_out = 0;
+        if( ec == zlib::error::need_buffers ||
+            ec == zlib::error::end_of_stream)
+        {
+            ec = {};
+            break;
+        }
+        if(ec)
+            return;
```
HowardHinnant: The use of `break` and `return` do the same thing, but spelling them differently like this doesn't give me confidence that the code is doing what the author intended.  I recommend using `break` in both places or `return` in both places.

Also it might be nice to combine this like so:
````
       if (ec)
       {
             if( ec == zlib::error::need_buffers ||
                 ec == zlib::error::end_of_stream)
                 ec = {};
             return;
       }
````

Comment (vinniefalco) [inline include/beast/websocket/detail/pmd_extension.hpp]: The idiom used throughout the library is as follows:
1. Errors cause a function to terminate control flow immediately via `return`
2. For functions returning `void` which complete successfully, the flow of control always runs to the end of the function
3. For functions returning other than `void` which complete successfully, the flow of control always runs to the function epilogue where the return value is calculated.

I preferred applying this style consistently. If I was to change it in the manner suggested, it would stand out as the only function that is written differently.

**Comment by vinniefalco** [comment]
The decline in coverage is unfortunate. Its because of zlib.

File: include/beast/websocket/detail/stream_base.hpp [hunk 3]
```diff
@@ -321,79 +396,96 @@ read_fh2(DynamicBuffer& db, close_code::value& code)
...
-        if(rd_msg_max_ && rd_size_ > rd_msg_max_)
+#if 0
+        if(rd_msg_max_ && rd_.size > rd_msg_max_)
         {
             code = close_code::too_big;
             return;
         }
-        rd_need_ = rd_fh_.len;
-        rd_cont_ = ! rd_fh_.fin;
+#endif
```
HowardHinnant: Can we drop this `#if` section?  I don't feel strongly that we need to.  Just checking if it was left in by accident.

vinniefalco: No, quite the opposite we should generate a compiler warning, this is lost functionality that we need to soonish figure out how to get back.

HowardHinnant:
👍  Checks out on macOS.

include/beast/websocket/detail/stream_base.hpp
ximinez: I'll add a warning.

vinniefalco:
I'm not feeling this branch anymore, it blows up when it is integrated into rippled. Back to the drawing board...

PR #233, opened 2017-01-20
permessage-deflate

vinniefalco:
This branch fixes a number of bugs and issues with the earlier permessage-deflate implementation. 

These commits are new:

* Add optional yield_to arguments
* [FOLD] Refactoring and fixes

### DIFF
--- a/extras/beast/test/yield_to.hpp
+++ b/extras/beast/test/yield_to.hpp
@@ -21,9 +21,9 @@ namespace test {
 
 /** Mix-in to support tests using asio coroutines.
 
-    Derive from this class and use yield_to to launch test functions
-    inside coroutines. This is handy for testing asynchronous asio
-    code.
+    Derive from this class and use yield_to to launch test
+    functions inside coroutines. This is handy for testing
+    asynchronous asio code.
 */
 class enable_yield_to
 {
@@ -72,12 +72,32 @@ public:
         Function will be called with this signature:
 
         @code
-            void f(yield_context);
+            void f(args..., yield_context);
         @endcode
+
+        @param f The Callable object to invoke.
+
+        @param args Optional arguments forwarded to the callable object.
     */
-    template<class Function>
+#if GENERATING_DOCS
+    template<class F, class... Args>
     void
-    yield_to(Function&& f);
+    yield_to(F&& f, Args&&... args);
+#else
+    template<class F>
+    void
+    yield_to(F&& f);
+
+    template<class Function, class Arg, class... Args>
+    void
+    yield_to(Function&& f, Arg&& arg, Args&&... args)
+    {
+        yield_to(std::bind(f,
+            std::forward<Arg>(arg),
+                std::forward<Args>(args)...,
+                    std::placeholders::_1));
+    }
+#endif
 };
 
 template<class Function>
--- a/include/beast/websocket/detail/invokable.hpp
+++ b/include/beast/websocket/detail/invokable.hpp
@@ -125,7 +125,7 @@ public:
     void
     emplace(F&& f);
 
-    void
+    bool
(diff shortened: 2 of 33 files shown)


test/websocket/stream.cpp
```diff
@@ -1395,37 +1244,75 @@ class stream_test
...
+
+        pmd.client_enable = true;
+        pmd.client_max_window_bits = 10;
+        pmd.client_no_context_takeover = false;
+        doClientTests(pmd);
+
+        pmd.client_enable = true;
+        pmd.client_max_window_bits = 10;
+        pmd.client_no_context_takeover = true;
+        doClientTests(pmd);
```
ximinez: `server_enable` defaults to false, and I don't see anywhere that it's getting enabled on the server, though I may have missed it. That means that deflate is not being tested by any config.

The coverage report seems to support this hypothesis, because `inflate` and `deflate` are both red. https://codecov.io/gh/vinniefalco/Beast/src/83173f5578294b0e2c0bb03c2a6dd0328a1b2e8c/include/beast/websocket/detail/pmd_extension.hpp#L365...477

include/beast/websocket/detail/invokable.hpp
```diff
@@ -125,15 +125,17 @@ class invokable
     void
     emplace(F&& f);
 
-    void
+    bool
     maybe_invoke()
     {
         if(base_)
         {
             auto const basep = base_;
```
miguelportilla: Can this local be avoided?
`        if (! base_)
            return false;
        (*base_)();
        base_ = nullptr;
        return true;`

ximinez: As written, this ensures that `base_` is null _before_ the function gets called. This prevents recursion / callback problems.

vinniefalco: Nope, because a call to `(*base_)()` could cause `emplace()` to get called which would break invariants. `base_` has to be set to `nullptr` before the invocation takes place.

vinniefalco: What he said

extras/beast/test/yield_to.hpp
```diff
@@ -72,12 +72,32 @@ class enable_yield_to
...
+    template<class Function, class Arg, class... Args>
     void
-    yield_to(Function&& f);
+    yield_to(Function&& f, Arg&& arg, Args&&... args)
+    {
+        yield_to(std::bind(f,
+            std::forward<Arg>(arg),
+                std::forward<Args>(args)...,
+                    std::placeholders::_1));
+    }
```
vinniefalco: Isn't it called from stream.cpp?

test/websocket/stream.cpp
vinniefalco: *whoops* that explains the drop in coverage

extras/beast/test/yield_to.hpp
ximinez: The calls in stream.cpp both have one parameter - a lambda, so it calls the other override.

ximinez: PS. Thanks!

include/beast/websocket/detail/mask.hpp
```diff
@@ -60,11 +60,17 @@ void
 maskgen_t<_>::rekey()
 {
     std::random_device rng;
+#if 0
     std::array<std::uint32_t, 32> e;
```
miguelportilla: This might work:

    std::uint32_t e[32];
    std::generate(std::begin(e), std::end(e), std::ref(rng));
    std::seed_seq ss(std::begin(e), std::end(e));

miguelportilla:
👍 Looks sane.

PR #236, opened 2017-01-23
Simplify Travis package install specification

ximinez:
Because all build matrix entries have the same package dependencies, this moves that info to the global config. Allows the rest of the matrix to be edited with impunity.


vinniefalco:
Lipstick on a pig :)

vinniefalco:
Part of #233, thanks

PR #245, opened 2017-01-31
1.0.0-b25

vinniefalco:
* Fixes for WebSocket echo server
* Better handler_ptr
* Fix 32-bit arm7 warnings
* Remove unnecessary include
* WebSocket server examples and test tidying

Documentation preview:
http://vinniefalco.github.io/stage/beast/


examples/websocket_async_echo_server.hpp
```diff
@@ -0,0 +1,375 @@
...
+                return;
+            }
+        }
+
+    private:
+        void
+        fail(std::string what, error_code ec)
+        {
+            auto& d = *d_;
+            if(d.server.log_)
```
mellery451: does it make sense for this method the method in the server class to both being checking `log_` ?

vinniefalco: I think so, because otherwise we might construct a `std::string` for nothing.

examples/websocket_async_echo_server.hpp
```diff
@@ -0,0 +1,375 @@
...
+    /** Destructor.
+    */
+    ~async_echo_server()
+    {
+        work_ = boost::none;
+        error_code ec;
+        ios_.dispatch(
+            [&]{ acceptor_.close(ec); });
+        for(auto& t : thread_)
+            t.join();
```
vinniefalco: `join` can only throw if `joinable` returns `false`. At this point in the code, `joinable` should return `true`.

examples/websocket_async_echo_server.hpp
```diff
@@ -0,0 +1,375 @@
...
+
+        std::unordered_map<std::type_index,
+            std::unique_ptr<callable>> list_;
+
+    public:
+        template<class Opt>
+        void
+        set_option(Opt const& opt)
+        {
+            std::unique_ptr<callable> p;
```
vinniefalco: Beast can only use C++11. Are you suggesting an alternative piece of code? If so, I'd like to see it :)

examples/websocket_async_echo_server.hpp
```diff
@@ -0,0 +1,375 @@
+//
+// Copyright (c) 2013-2016 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef BEAST_WEBSOCKET_ASYNC_ECHO_PEER_H_INCLUDED
+#define BEAST_WEBSOCKET_ASYNC_ECHO_PEER_H_INCLUDED
```
vinniefalco: Yeah it does..

vinniefalco: willfix

examples/websocket_async_echo_server.hpp
mellery451: nope - I'm good with this.

examples/websocket_async_echo_server.hpp
mellery451: 👍

examples/websocket_async_echo_server.hpp
mellery451: 👍

PR #253, opened 2017-02-05
Use bitwise instead of boolean OR

nbougalis:


vinniefalco:
Suggested commit message:
```
Check ostream modifier correctly
```
Also CHANGELOG.md needs an updating

ximinez:
s/OR/AND/

nbougalis:
D'oh. `AND`, yes.

vinniefalco:
Just a reminder, we need to analyze the calls to `log.flush()` to determine if they are still necessary (since I added them to work around this `unitbuf` bug):
```
extras\beast\unit_test\reporter.hpp(286): os_.flush();
extras\beast\unit_test\suite.hpp(356): suite_.log.flush();
extras\beast\unit_test\suite.hpp(397): suite_.log.flush();
```

nbougalis:
With this change, the calls to `flush` are no longer needed.

vinniefalco:
Please remove the calls to `flush` then!

vinniefalco:
Fixed in **1.0.0-b29**

Issue #166, opened 2016-11-02
Add verbosity option for unit tests


bachase:
Consider adding a verbosity level to determine when `testcase` identifiers are printed.  Since `BEAST_EXPECT` reports the file/line number, those identifiers are not typically needed.  But if you have multiple pathways to `BEAST_EXPECT` for different test cases, the file/line number are not sufficient to disambiguate the test.

vinniefalco:
I believe this has been addressed - if not, please open a new issue!

PR #263, opened 2017-02-10
Print testcase on failure only

bachase:
Addresses #166 by only reporting the testcase name if the testcase has a failure.

I'm open to making this configurable instead if that is important.

### DIFF
--- a/extras/beast/unit_test/reporter.hpp
+++ b/extras/beast/unit_test/reporter.hpp
@@ -91,6 +91,7 @@ private:
     results results_;
     suite_results suite_results_;
     case_results case_results_;
+    bool printed_case_ = false;
 
 public:
     reporter(reporter const&) = delete;
@@ -244,9 +245,7 @@ reporter<_>::
 on_case_begin(std::string const& name)
 {
     case_results_ = case_results(name);
-    os_ << suite_results_.name <<
-        (case_results_.name.empty() ? "" :
-            (" " + case_results_.name)) << std::endl;
+    printed_case_ = false;
 }
 
 template<class _>
@@ -272,6 +271,13 @@ on_fail(std::string const& reason)
 {
     ++case_results_.failed;
     ++case_results_.total;
+    if(!printed_case_)
+    {
+        os_ << suite_results_.name <<
+            (case_results_.name.empty() ? "" :
+                (" " + case_results_.name)) << std::endl;
+        printed_case_ = true;
+    }
     os_ <<
         "#" << case_results_.total << " failed" <<
        (reason.empty() ? "" : ": ") << reason << std::endl;


vinniefalco:
IDK how I feel about this. Can you provide an example of before and after output for a suitable exemplar?

bachase:
I find this useful for cases when I've rebased or just want to see if I've broken anything. In those cases, the testcase names for successful cases are just noise.  

Below are from ripple with a contrived error in two cases:
[no.spam.txt](https://github.com/vinniefalco/Beast/files/767285/no.spam.txt)
[spam.txt](https://github.com/vinniefalco/Beast/files/767286/spam.txt)

Generally, I'm not sure when always printing the testcase name is useful.  When filtering the tests being run?

vinniefalco:
Can we at least print the name of the unit test? For example, given:
```
ripple.tx.Taker XRP Quantization: input
ripple.tx.Taker XRP Quantization: output
ripple.tx.Taker IOU to IOU
ripple.consensus.Consensus single peer
ripple.consensus.Consensus 5 connected peers
ripple.consensus.Consensus slow peer
ripple.consensus.Consensus close time disagree
#1 failed: Consensus_test.cpp(190)
#2 failed: Consensus_test.cpp(190)
#3 failed: Consensus_test.cpp(190)
#4 failed: Consensus_test.cpp(190)
#5 failed: Consensus_test.cpp(190)
#6 failed: Consensus_test.cpp(190)
ripple.consensus.Consensus test clique fork
#6 failed: Consensus_test.cpp(232)
#7 failed: Consensus_test.cpp(232)
#8 failed: Consensus_test.cpp(232)
#9 failed: Consensus_test.cpp(232)
#10 failed: Consensus_test.cpp(232)
#11 failed: Consensus_test.cpp(232)
```

We could instead print:
```
ripple.tx.Taker
ripple.consensus.Consensus close time disagree
#1 failed: Consensus_test.cpp(190)
#2 failed: Consensus_test.cpp(190)
#3 failed: Consensus_test.cpp(190)
#4 failed: Consensus_test.cpp(190)
#5 failed: Consensus_test.cpp(190)
#6 failed: Consensus_test.cpp(190)
ripple.consensus.Consensus test clique fork
#6 failed: Consensus_test.cpp(232)
#7 failed: Consensus_test.cpp(232)
#8 failed: Consensus_test.cpp(232)
#9 failed: Consensus_test.cpp(232)
#10 failed: Consensus_test.cpp(232)
#11 failed: Consensus_test.cpp(232)
```

?

bachase:
Sure, that seems like a fair compromise.  I'll force push that.

mellery451:
Should we also print the testcase name if/when `log` is written to during the test? Here is an example from a current test:
```
ripple.ripple_data.BuildInfo version packing
    Ripple Version: 0.60.0-b4
    Protocol Version: 1.2
```
...otherwise we get log messages without any context. I think @vinniefalco 's suggestion about keeping the test suite name alleviates this concern.

bachase:
@mellery451 I can do either. I've not had a reason to use log before, so I'll defer to others on what makes more sense.

vinniefalco:
If a unit test produces any output using `log` its important for the `reporter` to show the test case or unit test name, or else not only will there be no context but it will look like the output is coming from the _previous_ test that ran.

bachase:
This makes more sense as a configurable option, since listing all the testcases does provide value (to show what you are testing).  

I'm going to close for now and will eventually get around to an improved version.

PR #266, opened 2017-02-17
fixed inconsitency in writer documentation

s-brain:
Return values were stated as the opposite of reality


vinniefalco:
Fixed in **1.0.0-b29** - Thanks!

PR #272, opened 2017-02-24
1.0.0-b30

vinniefalco:
This fixes races when composed websocket operations suspend. It also allows pings and writes to happen simultaneously.

### DIFF
--- a/include/beast/websocket/impl/close.ipp
+++ b/include/beast/websocket/impl/close.ipp
@@ -190,7 +190,8 @@ operator()(error_code ec, bool again)
 upcall:
     if(d.ws.wr_block_ == &d)
         d.ws.wr_block_ = nullptr;
-    d.ws.rd_op_.maybe_invoke();
+    d.ws.rd_op_.maybe_invoke() ||
+        d.ws.ping_op_.maybe_invoke();
     d_.invoke(ec);
 }
 
--- a/include/beast/websocket/impl/close.ipp
+++ b/include/beast/websocket/impl/close.ipp
@@ -147,25 +147,33 @@ operator()(error_code ec, bool again)
                         boost::asio::error::operation_aborted));
                 return;
             }
-            // fall through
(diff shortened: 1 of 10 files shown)


include/beast/websocket/impl/close.ipp
```diff
@@ -147,25 +147,32 @@ operator()(error_code ec, bool again)
                         boost::asio::error::operation_aborted));
                 return;
             }
-            // fall through
+            d.ws.wr_block_ = &d;
+            // [[fallthrough]
```
nbougalis: Missing extra `]` at the end.

vinniefalco: willfix

include/beast/websocket/impl/ping.ipp
```diff
@@ -146,24 +146,31 @@ operator()(error_code ec, bool again)
...
             // send ping frame
             d.state = 99;
-            BOOST_ASSERT(! d.ws.wr_block_);
-            d.ws.wr_block_ = &d;
             boost::asio::async_write(d.ws.stream_,
                 d.fb.data(), std::move(*this));
             return;
 
         case 2:
+            BOOST_ASSERT(! d.ws.wr_block_);
```
seelabs: I'm lost. I see that `wr_block_` is cleared in the `upcall`. But when I look on line 132 of this commit, I see that if we have a `wr_block_`, then we set the state to `2`. The time time into this state machine I would expect this assert to fail. What I'm missing is how is `wr_block_` cleared between state `0` and state `2`?

vinniefalco: `upcall` is a final state, once we perform the upcall (the call to the final completion handler) the composed operation is done and gets destroyed. When `wr_block_` is set and we suspend, it is the other blocking operation which clears `wr_block_` before the upcall (or continuing the composed operation such as in the case of `write_op`).

seelabs:
:+1:

PR #277, opened 2017-03-04
Added missing const overload of dynabuf_readstream<>::next_layer().

florianbehrens:


vinniefalco:
This will go into **1.0.0-b31**:
https://github.com/vinniefalco/Beast/tree/b31

vinniefalco:
Part of b31 now

PR #293, opened 2017-03-31
msvc build fix.

lysevi:
build on msvc 2015.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 if (MSVC)
     # /wd4244 /wd4127
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4100 /wd4244 /wd4251 /MP /W4 /bigobj /D BOOST_COROUTINES_NO_DEPRECATION_WARNING=1 /D _WIN32_WINNT=0x0601 /D _SCL_SECURE_NO_WARNINGS=1 /D _CRT_SECURE_NO_WARNINGS=1")
-    set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MTd")
+    #set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MTd")
     set (CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /Ob2 /Oi /Ot /GL /MT")
     set (CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} /Oi /Ot /MT")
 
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -24,19 +24,21 @@ if (MSVC)
...
     find_package(Threads)
 
     set(CMAKE_CXX_FLAGS
       "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Wpedantic -Wno-unused-parameter -DBOOST_COROUTINES_NO_DEPRECATION_WARNING=1")
 endif()
 
+set(Boost_USE_STATIC_LIBS ON)
+set(Boost_USE_MULTITHREADED ON)
+
+find_package(Boost REQUIRED COMPONENTS regex coroutine context thread filesystem program_options system)
```
vinniefalco: why regex?

vinniefalco:
I like the idea, but I get these warnings when I run CMake on Windows now:
```
$ cmake ..
CMake Warning at C:/Program Files/CMake/share/cmake-3.6/Modules/FindBoost.cmake:743 (message):
  Imported targets not available for Boost version 106300
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.6/Modules/FindBoost.cmake:842 (_Boost_COMPONENT_DEPENDENCIES)
  C:/Program Files/CMake/share/cmake-3.6/Modules/FindBoost.cmake:1395 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:37 (find_package)
```

Any ideas?

CMakeLists.txt
lysevi: regex need to work boost::asio::read_until

lysevi:
may be need change to

```cmake
find_package(Boost COMPONENTS regex coroutine context thread filesystem program_options system)
```

vinniefalco:
Same problem:
```
$ cmake ..
CMake Warning at C:/Program Files/CMake/share/cmake-3.6/Modules/FindBoost.cmake:743 (message):
  Imported targets not available for Boost version 106300
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.6/Modules/FindBoost.cmake:842 (_Boost_COMPONENT_DEPENDENCIES)
  C:/Program Files/CMake/share/cmake-3.6/Modules/FindBoost.cmake:1395 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:37 (find_package)
```

harrishancock:
@vinniefalco try upgrading your CMake. According to this SO question, you need CMake 3.7 to find Boost 1.63.

http://stackoverflow.com/questions/42123509/cmake-finds-boost-but-the-imported-targets-not-available-for-boost-version

vinniefalco:
How will CMake find my Boost in Windows exactly? Does Boost have to be in a well-known location?

harrishancock:
I don't think there's any well-known location. The usual way I do it is:

1. unzip Boost to `/path/to/boost`
2. cd to `/path/to/boost` and run `bootstrap.bat` and `b2 stage` (with appropriate parameters)
3. `export BOOST_ROOT=/path/to/boost`

You could also `b2 install` Boost somewhere instead of staging it in its build directory, in which case that install dir would become `BOOST_ROOT`. I find that to be useful when building multiple types of Boost (e.g., 32-, 64-bit) from the same source. Lastly, I I'm sure there are precompiled Boost distributions available for MSVC, too, though I haven't used them.

I think most other ways to do it involve using a package manager. I've had good luck with Hunter, although it builds the entirety of Boost during the CMake configure process. Conan and Vcpkg both have ways to do it, but I haven't tried them yet.

lysevi:
Cmake search boost in -DBOOST_ROOT="path/to/boost/"

ras0219-msft:
I'd recommend using
```cmake
option(Boost_USE_STATIC_LIBS "Use static libraries for boost" ON)
```
This enables users to disable it from the command line if they wish. This is important for vcpkg users, because we install DLLs for Boost by default.

@vinniefalco CMake searches a _ton_ of locations for Boost. The most common way for Windows users is probably via the `%BOOST_ROOT%` environment variable, with `-DBOOST_ROOT` being a close second. For anyone using vcpkg (`-DCMAKE_TOOLCHAIN_FILE=..../vcpkg.cmake`) we set `CMAKE_PREFIX_PATH` which also gets searched.

CMake documentation on `option()`: https://cmake.org/cmake/help/v3.0/command/option.html

vinniefalco:
Okay, I upgraded to CMake 3.7.2, and I like these changes! I've incorporated the option as per @ras0219-msft suggestion (thanks for that). Question, should I use `REQUIRED` here or not:
```
find_package(Boost COMPONENTS regex coroutine context thread filesystem program_options system)
```
??

ras0219-msft:
Yep, `REQUIRED` is the best way to alert the user that the build cannot continue when they don't have Boost.

vinniefalco:
Should I put `REQUIRED` on **all** calls to `find_package` or just the Boost one?

vinniefalco:
err... OpenSSL is optional so I guess I'm asking about
```
find_package(REQUIRED Threads)
```

I think I need it because otherwise, gcc/clang builds get link errors

ras0219-msft:
Yeah, it probably should be (though honestly I don't think I've ever seen that particular `find_package()` fail...).

CMakeLists.txt
vinniefalco: None of the Beast code calls `boost::asio::read_until` - where do you see it?

vinniefalco:
Part of https://github.com/vinniefalco/Beast/pull/294

Thanks everyone!!!

vinniefalco:
These changes have messed up my Visual Studio RelWithDebInfo builds, investigating...

PR #304, opened 2017-04-15
1.0.0-b33

vinniefalco:
* Require Visual Studio 2015 Update 3 or later

HTTP

* Use fwrite return value in file_body

WebSocket

* Set internal state correctly when writing frames
* Add decorator unit test
* Add write_frames unit test


vinniefalco:
Attention:
@nctusdk
@matinko

PR #345, opened 2017-05-08
Version 39

vinniefalco:
Beast versions are now identified by a single integer which is incremented on each merge. The macro BEAST_VERSION identifies the version number, currently at 39. A version setting commit will always be at the tip of the master and develop branches.

* Use beast::string_view alias

HTTP

* Tidy up basic_parser javadocs

WebSocket:

* Add websocket async echo ssl server test:
* Fix eof error on ssl::stream shutdown

API Changes:

* Refactor http::header contents
* New ostream() returns dynamic buffer output stream
* New buffers() replaces to_string()
* Rename to multi_buffer, basic_multi_buffer
* Rename to flat_buffer, basic_flat_buffer
* Rename to static_buffer, static_buffer_n
* Rename to buffered_read_stream
* Harmonize concepts and identifiers with net-ts
* Tidy up HTTP reason_string


vchang-akamai:
I tried building with b39. (gcc 4.8.4 on linux)  After fixing some API changes, I still ran into some compilation problems which appear to be related to the ostream_buffer changes. 

Excerpts:
```
/usr/include/c++/4.8/streambuf:802:7: error: 'std::basic_streambuf<_CharT, _Traits>::basic_streambuf(const std::basic_streambuf<_CharT, _Traits>&) [with _CharT = char; _Traits = std::char_traits<char>]' is private
       basic_streambuf(const basic_streambuf& __sb)
       ^

[...]/include/beast/core/detail/ostream.hpp:72:5: error: within this context
     ostream_buffer(


[...]/include/beast/core/detail/ostream.hpp:203:33: note: synthesized method 'beast::detail::ostream_buffer<DynamicBuffer, CharT, Traits>::ostream_buffer(beast::detail::ostream_buffer<DynamicBuffer, CharT, Traits>&&) [with DynamicBuffer = beast::basic_multi_buffer<std::allocator<char> >; CharT = char; Traits = std::char_traits<char>]' first required here 
     , osb_(std::move(other.osb_))
                                 ^
```


/usr/include/c++/4.8/streambuf:802:

```
    private:
      // _GLIBCXX_RESOLVE_LIB_DEFECTS
      // Side effect of DR 50. 
      basic_streambuf(const basic_streambuf& __sb)
      : _M_in_beg(__sb._M_in_beg), _M_in_cur(__sb._M_in_cur), 
      _M_in_end(__sb._M_in_end), _M_out_beg(__sb._M_out_beg), 
      _M_out_cur(__sb._M_out_cur), _M_out_end(__sb._M_out_cur),
      _M_buf_locale(__sb._M_buf_locale) 
      { }
```
Earlier in the same file, there is a protected constructor, but it's not a copy-constructor:
```
    protected:
      /**
       *  @brief  Base constructor.
       *
       *  Only called from derived constructors, and sets up all the
       *  buffer data to zero, including the pointers described in the
       *  basic_streambuf class description.  Note that, as a result,
       *  - the class starts with no read nor write positions available,
       *  - this is not an error
      */
      basic_streambuf()
      : _M_in_beg(0), _M_in_cur(0), _M_in_end(0), 
      _M_out_beg(0), _M_out_cur(0), _M_out_end(0),
      _M_buf_locale(locale()) 
      { }
```

vinniefalco:
Is this relevant?
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54316

vinniefalco:
@vchang-akamai How can I repro this on CI? See: https://github.com/vinniefalco/Beast/commits/ostream

vinniefalco:
@vchang-akamai Does this reproduce the issue?
https://wandbox.org/permlink/50w8X8rKnNbuhf6K

vchang-akamai:
The build errors in your wandbox link appear to be pretty similar.

vinniefalco:
I have an idea for a fix but it requires a memory allocation. You might not like it, but I don't see any way around it because GCC4.8.4 does not have full support for C++11 (which requires that `std::basic_streambuf` is movable). If the memory allocation bothers you, you will have to use a more recent GCC.

vchang-akamai:
hmm.. ok. I may have to live with the memory allocation for now until we can update  our toolchain...

vinniefalco:
@vchang-akamai This branch might fix the problem
https://github.com/vinniefalco/Beast/commits/ostream

I have not tested it on CI yet. Also, I need help figuring out what macro to check to determine if the version of libstdc++ being used does not support `std::basic_stream`.

vchang-akamai:
I successfully built using the "ostream" branch.

vchang-akamai:
Re: Macros:

I'm not sure if this helps, but I found some GNU-specific libstdc++ macro info:
[https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html](url)

and a timeline of releases here:
[https://gcc.gnu.org/develop.html#timeline](url)

vinniefalco:
@vchang-akamai Please try the **ostream** branch again, it uses `std::is_move_constructible` to switch between implementations as needed. It should work for you.

vchang-akamai:
Building with new "ostream" branch does not work. Ran into same original problem with ostream_buffer and std::basic_streambuf.

vinniefalco:
@vchang-akamai Hmm... I'm not sure why I am getting different results. On wandbox I have gcc4.8.4 set, and I even put a static assert in there that the `ostream_buffer` is not move constructible. Why is this code giving you different results from https://wandbox.org/permlink/iT3GPEn5fjaY4yLu ?
I can't reproduce your compile error.

vchang-akamai:
My code isn't actually calling ostream() on a multi_buffer.  It was my call to beast::websocket::async_accept() that seemed to trigger the error (there were first a bunch of messages about "required from" template functions in write.ipp and accept.ipp before it got to the ostream and std::basic_streambuf errors).

I experimentally commented out my call to async_accept(), and then the compiler had a similar complaint for a call to beast::websocket::async_write()

not sure if that clarifies anything...

vinniefalco:
Right, your code doesn't have to call `ostream`, it already happens in the implementation of Beast. Have you tried compiling the tests? I'd like to know if they build for you in the **ostream** branch.

vinniefalco:
@vchang-akamai I need the smallest possible test case that reproduces the problem, could you please put that together on wandbox? I'd also like to know what version of stdlib you are using, and what happens when you compile this code in your project:
```
#include <beast/core/ostream.hpp>
#include <beast/core/multi_buffer.hpp>
    static_assert(! std::is_move_constructible<
        beast::detail::ostream_buffer<beast::multi_buffer, char, std::char_traits<char>>>::value, "");
```

vchang-akamai:
Trying to compile the beast tests:  My build environment's cmake is too old, so just compiling by hand from the "test" subdir.
` g++ -m64 -std=c++11   -I [...]include -I../extras *cpp ../extras/beast/unit_test/main.cpp -L [...]/lib/ -lboost_system -o testsuite`
(the "include" dir has both beast and boost. The "lib" dir has boost libs)

I'm getting this error:
```
../extras/beast/unit_test/main.cpp:102:26: error: invalid initialization of non-const reference of type 'beast::unit_test::dstream {aka std::basic_ostream<char>&}' from an rvalue of type '<brace-enclosed initializer list>'
     dstream log{std::cerr};
                          ^
```


Ok, I'll see what I can do about producing a small wandbox example...

vinniefalco:
@vchang-akamai It looks like you're using a version of gcc with a bug in C++11 conformance:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50025

vinniefalco:
@vchang-akamai I think this reproduces the issue, working on a fix:
https://travis-ci.org/vinniefalco/Beast/jobs/230171812#L9496

vchang-akamai:
Re: stdlibc++ version:
I think I'm using GLIBCXX_3.4.19, dated 20150426

Re: trying to compile the static_assert()
I added it to one of my files and the static assertion failed.

vinniefalco:
@vchang-akamai Phew, I think this is fixed in the pull request!!

vinniefalco:
@vchang-akamai I also fixed the compilation error with the unit test (`dstream dout{std::cerr}`) so you should be able to build the unit tests now as well. Please let me know if there are still errors, thanks!

vchang-akamai:
ok, I grabbed the latest b39 and it builds. Also tested the dstream error from the unit test and that's gone too.

Thanks!!

Q: Re: the memory allocation in your work around:   When does it occur?  Is this only if I use ostream on a multi_buffer?  I'm not currently doing that.  Or would it also happen if I use the various beast::websocket::async_*() calls?

vinniefalco:
>Q: Re: the memory allocation in your work around: When does it occur? Is this only if I use ostream on a multi_buffer? I'm not currently doing that. Or would it also happen if I use the various beast::websocket::async_*() calls?

It will happen any time Beast serializes the header of an HTTP message. This happens on `write` and `async_write`.

vinniefalco:
If it makes you feel any better, the previous version was doing multiple memory allocations, this does just two :)

PR #436, opened 2017-06-08
Fix file_body::get() not setting the more flag correctly

### LINKED ISSUE #434: The http-server example doesn't work for files > 4 KB
On master when serving files larger than 4 KB the http-server example only sends the first 4 KB of a file leaving the stream open but not sending any more.

The problem is an inversion of the check for more data in file_buffer. I'll have a Pull Request ready in a bit.

niklas88:
The check for more data actually checked if the end of data was reached
so it was logically inverted.

This fixes #434


vinniefalco:
Thanks! This will go into **v51**

vinniefalco:
Merged to **v51**

Issue #442, opened 2017-06-08
CMake tests/examples option


xsacha:
By default all tests/examples are compiled. Could these be made in to options?
See: https://github.com/xsacha/Beast/blob/hunter-b50/CMakeLists.txt
> option(Beast_BUILD_EXAMPLES "Build examples" OFF)
> option(Beast_BUILD_TESTS "Build tests" OFF)

Can submit this part as a pull if you want but you probably want them enabled by default?

vinniefalco:
Hmm...why wouldn't we want to build those?

xsacha:
If we want to include Beast in a submodule or another project, we will not be concerned with tests or examples but just want the linkage and includes.

vinniefalco:
If you're using Beast in another project, you don't need Beast's CMakeLists.txt at all, just add an `include_directories` line to your own CMake and that's all you need!

xsacha:
This is only because there is another issue in the CMakeLists: no target.
If you have a Beast target, it provides linkage and includes automatically. This is how CMake is designed.
If you can provide a Beast interface, all my projects can link to the same interface no matter where Beast is located and what it links to.

vinniefalco:
Your commit is intriguing. I am interested in simplifying the link targets for the tests and examples. However, when using Beast in another project you do not need program_options, coroutine, context, filesystem, or thread - Beast doesn't depend on those (just the tests).

I am also puzzled why you have removed `GroupSources` - that's a very important function because it helps to organize the generated Visual Studio project files which I use exclusively for development.

xsacha:
Misunderstanding on my part. I would change it to only Boost::system in this case.

I wasn't suggesting going so far yet. Baby steps :) Not all the changes in that commit are suggested for here. Some of them are changes due to unrelated package-management.

vinniefalco:
This is now in **develop**

Issue #443, opened 2017-06-08
CMake target


xsacha:
An interface target is missing. It is very easy to add, see code sample below.
The includes and linkage (Boost) can be added to this.

Adding this will greatly simplify your own CMakeLists. See: your usage of include paths and library linkage vs my commit.

```
### Library ###
add_library(${PROJECT_NAME} INTERFACE)
target_link_libraries(${PROJECT_NAME} INTERFACE Boost::coroutine Boost::context Boost::thread Boost::filesystem Boost::program_options Boost::system)

### Install ###
set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")

set(config_install_dir "lib/cmake/${PROJECT_NAME}")
set(include_install_dir "include")

set(version_config "${generated_dir}/${PROJECT_NAME}ConfigVersion.cmake")
set(project_config "${generated_dir}/${PROJECT_NAME}Config.cmake")
set(targets_export_name "${PROJECT_NAME}Targets")
set(namespace "${PROJECT_NAME}::")

include(CMakePackageConfigHelpers)
write_basic_package_version_file(
    "${version_config}" COMPATIBILITY SameMajorVersion
)

# Note: use 'targets_export_name'
configure_file("cmake/Config.cmake.in" "${project_config}" @ONLY)

install(
    TARGETS ${PROJECT_NAME}
    EXPORT "${targets_export_name}"
    INCLUDES DESTINATION "${include_install_dir}"
)

install(
    DIRECTORY "include/"
    DESTINATION "${include_install_dir}"
)

install(
    FILES "${project_config}" "${version_config}"
    DESTINATION "${config_install_dir}"
)

install(
    EXPORT "${targets_export_name}"
    NAMESPACE "${namespace}"
    DESTINATION "${config_install_dir}"
)
```

vinniefalco:
It looks like we have the interface target now? Is this issue still relevant?

xsacha:
If installing target isn't an option, then yes this can be closed.

xsacha:
@vinniefalco 
Just a note that it looks like Boost is officially moving to CMake:
http://boost.2283326.n4.nabble.com/CMake-Announcement-from-Boost-Steering-Committee-tt4696934.html

vinniefalco:
It is possible yes although I am sure I will have to change something

PR #460, opened 2017-06-10
Add options for tests and examples

xsacha:
#442

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@ cmake_minimum_required (VERSION 3.5.2)
 project (Beast)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
+option (Beast_BUILD_EXAMPLES "Build examples" ON)
+option (Beast_BUILD_TESTS "Build tests" ON)
 
 if (MSVC)
     set (CMAKE_VERBOSE_MAKEFILE FALSE)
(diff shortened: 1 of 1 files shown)


xsacha:
Added #443

vinniefalco:
I don't understand CMake well enough to evaluate these changes but an "install" is sort of out of scope, as this is being proposed as a Boost library and therefore, does not require an "install" target since it will adhere to the Boost directory structure.

xsacha:
If it were to be included in Boost, then it would not need a CMakeLists.txt, correct? Boost unfortunately doesn't use CMake. So I am not sure how it is related.

xsacha:
I have switched back to a single commit.

vinniefalco:
>If it were to be included in Boost, then it would not need a CMakeLists.txt

Its not required, but I need CMakeLists.txt so I can generate Visual Studio project files (.vcxproj). That's my primary development environment.

xsacha:
OK. So what's the problem? It will still make these .vcxproj :)

vinniefalco:
Let me make sure I understand this (I am no CMake expert) - this change only adds the option of not building the examples or tests?

xsacha:
Yes and they are enabled by default.

vinniefalco:
While these changes look like they do what is intended, this is why I think we should not merge them:

* They have little meaning when Beast becomes part of Boost

* The build options and install target are not part of my workflow, and not part of the continuous integration scripts, so they will not receive testing. I won't know if I break them. Which means, they will break.

xsacha:
I think this whole file will have little meaning when Beast becomes part of Boost unless it is still used for Travis CI (but I think Boost uses some other CI).

These options should not be hard to maintain but I will fix them up if needed. It should be as simple as adding new tests inside the if (Beast_BUILD_TESTS) and new examples inside the if (Beast_BUILD_EXAMPLES).

vinniefalco:
The formal review is July 1st, just 19 days!

xsacha:
Oh I see! But the project isn't even API complete yet?

vinniefalco:
>But the project isn't even API complete yet?

It is **very** close to being complete.

xsacha:
@ksergey has a newer version of this rebased on v61 so this can be closed.

PR #461, opened 2017-06-10
Integrate Beast interface target

xsacha:
Adds a target (partial of #443)

Does not install the target.


vinniefalco:
While these changes look like they do what is intended, this is why I think we should not merge them:

* They have little meaning when Beast becomes part of Boost

* The build options and install target are not part of my workflow, and not part of the continuous integration scripts, so they will not receive testing. I won't know if I break them. Which means, they will break.

xsacha:
The target can be used throughout your tests and examples. So you will know if it breaks.

xsacha:
Added a new commit that uses targets throughout.

vinniefalco:
I put together a branch based on your changes: https://github.com/vinniefalco/Beast/commits/cmake

However, I get errors when I run CMake on Windows:
```
$ cmake ..
-- Boost version: 1.64.0
-- Configuring done
CMake Error at examples/CMakeLists.txt:8 (add_executable):
  Target "echo-op" links to target "Boost::system" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


CMake Error at examples/CMakeLists.txt:8 (add_executable):
  Target "echo-op" links to target "Boost::system" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

vinniefalco:
Related failed project build output:
https://travis-ci.org/vinniefalco/Beast/jobs/241761226

xsacha:
Which version of CMake is this?
I tested using .travis.yml on CMake 3.7 (and script built boost 1.58).
Also your Windows/MSVC doesn't have COMPONENTS?

vinniefalco:
My cmake:
```
$ cmake --version
cmake version 3.8.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

Travis is using... I am not sure. Perhaps its in the log somewhere?

xsacha:
Looks like a bug in whichever CMake version Travis is using because it isn't automatically adding dependant targets (date_time and atomic). I can confirm 3.7 and 3.8 do the correct thing here on Linux.
As a workaround, these can be manually added to targets since they are dependencies since ~v1.5 anyway.

On Windows, you should probably use COMPONENTS (same as Linux). Is there a reason you changed this? Here: https://github.com/vinniefalco/Beast/commit/0128743ddadcc1ad3f7f9c5b78116ae2bbe0c68a

vinniefalco:
I tried to cherry pick the commit https://github.com/vinniefalco/Beast/pull/461/commits/e9492acc19eb21944df79ad926c17b2c77a1aa1a into my branch:
https://github.com/vinniefalco/Beast/commits/cmake
But it has conflicts. Can you please rebase this pull request off my branch?

xsacha:
I need to do some more work on it. It looks like it does use the thread library but just has not built date_time and atomic so I need to tell the interface not to use it.

xsacha:
OK. That's all working now. Sorry about spamming travis but I didn't want to recompile Boost 1.58 again, without date_time/atomic.

Unfortunately all the disable options for date_time and atomic weren't working so I just imported empty targets.

vinniefalco:
This is looking much better now!

vinniefalco:
This will go in to *v55*

vinniefalco:
Exactly what I was afraid of, has happened as a result of this merge. The changes break homebrew-installed Boost on Mac OS with apple's clang. And I didn't know because that is not a system that I test with, and it is not configured on the CI scripts. This has to be rolled back in v56, it has broken some users.

xsacha:
Is there a way to test this scenario?

vinniefalco:
I have simply removed these two lines in **v56**
```
set (Boost_NO_SYSTEM_PATHS ON)
set (Boost_USE_MULTITHREADED ON)
```

This is the problem with the changes you proposed:
image

I don't mind adding this INTERFACE thing, if we can do it without relying on imported targets.

xsacha:
Ok. Thanks for the info. That makes sense.

I can submit without Boost and SSL targets as original. I had added just to simplify the list.

vinniefalco:
In principle I like your change with the imported targets. But in practice, it suffers because of the issues between Boost and CMake.

vinniefalco:
Is removing these two lines the right thing to do?
```
set (Boost_NO_SYSTEM_PATHS ON)
set (Boost_USE_MULTITHREADED ON)
```

xsacha:
Not the multithreaded.
Added another commit.

vinniefalco:
`Boost_USE_MULTITHREADED` defaults to on:
https://lists.gnu.org/archive/html/discuss-gnuradio/2011-12/txta1SOQqu6kA.txt

What is the reason to manually set it to ON?

vinniefalco:
**v56** has your change, thanks

xsacha:
Ah, I didn't know multithreaded was on by default.
I set it on everywhere too.

vinniefalco:
Does **v56** look good to you? Its in the pull request queue

xsacha:
Looks fine to me but MSVC still grabs all of Boost instead of COMPONENTS. I hope that doesn't affect it findings Boost_SYSTEM_LIBRARY.
If it all compiles fine, can leave that to another day.

xsacha:
I have redone the part of the commit that didn't make it in without using Boost targets.

vinniefalco:
Its in **v57** now

vinniefalco:
@xsacha Something changes with the warnings and now CI fails
https://travis-ci.org/vinniefalco/Beast/jobs/243058513

vinniefalco:
I see it. You removed:
```
include_directories (SYSTEM ${Boost_INCLUDE_DIRS})
```

The "SYSTEM" keyword is special, it informs gcc and clang that headers in that directory are considered part of the system headers, and warnings should be suppressed for them.

xsacha:
Fixed. I didn't know that was intentional.

vinniefalco:
Still causes warnings and build failure, see the last item in the matrix:
https://travis-ci.org/vinniefalco/Beast/builds/243073581

xsacha:
This one tested fine on .travis.yml here.

vinniefalco:
Is that different?

xsacha:
Yeah apparently some ongoing bug in CMake. The SYSTEM keyword does not send it to -isystem as you would expect but manually setting the properties works.
See: https://github.com/MoMEMta/MoMEMta/blob/master/CMakeLists.txt#L83

vinniefalco:
I'm going to have to put this pull request on hold, because I am in the middle of a large refactoring of all the examples. The directories are moving around as are the CMakeLists.txt. So bringing this in would create a lot more work for me than I can handle right now. This is the branch:
https://github.com/vinniefalco/Beast/commits/v58

vinniefalco:
Version 58 is in PR now, please rebase
https://github.com/vinniefalco/Beast/pull/492

PR #502, opened 2017-06-16
Version 60

vinniefalco:
* String comparisons are public interfaces
* Fix response message type in async websocket accept
* New server-framework, full featured server example

Documentation preview:
http://vinniefalco.github.io/stage/beast/v60


vinniefalco:
Merged

PR #511, opened 2017-06-18
Cmake fix v61

ksergey:
#509 for v61 branch

PR #514, opened 2017-06-19
Cmake fix v61

ksergey:
Changes for v61 branch #511


xsacha:
@ksergey I thought OpenSSL isn't related to Beast. Why is it added to target if found?

vinniefalco:
Beast does not require OpenSSL but the tests and examples will use it if found. And if you want to use OpenSSL sockets with Beast you will of course have to configure OpenSSL includes and link libraries.

xsacha:
@vinniefalco Will some features of Beast (eg. https) not work without OpenSSL linked? Or is it only for things specific to the example?
I'm thinking if it is needed for certain features of Beast, it could be an option like Beast_USE_OPENSSL. Otherwise it shouldn't be linked to Beast.

vinniefalco:
Its for the example. If you don't want to use OpenSSL in your own program that uses Beast, you don't have to.

ksergey:
@vinniefalco INTERFACE target provides ability to link Beast with all necessary third party libraries without additional code in parent CMakeLists.txt.

I agree with @xsacha - it will be good to have BEAST_USE_OPENSSL option.

Example1:
```
option(BEAST_USE_OPENSSL "" ON)
add_executable(app main.cpp)
target_link_libraries(app Beast)
```

Example2:
```
find_package(OpenSSL REQUIRED)
include_directories(${OPENSSL_INCLUDE_DIR})
add_executable(app main.cpp)
target_link_libraries(app Beast ${OPENSSL_LIBRARIES})
```

For me example 1 looks much better than example 2.

vinniefalco:
I don't think it should be Beast's job to help users find OpenSSL, that responsibility lies with Asio if anything. Also note, I have already starting using `BEAST_USE_OPENSSL` as a macro to conditionally compile tests and examples if OpenSSL is found.

xsacha:
@ksergey,
Like vinnie said, the actual Beast project isn't dependent on OpenSSL so the target should not require it to be linked (even if Beast is using OpenSSL for some task, it is using it indirectly).
So this should probably be dropped. I know it means you probably need to link OpenSSL separately but it isn't that big of a deal hopefully :)

PR #519, opened 2017-06-21
Version 63

vinniefalco:
* Use std::to_string instead of lexical_cast
* Put num_jobs back up on Travis
* Fix BOOST_ROOT dir check when installing
* Only build and run tests in variant=coverage
* Move benchmarks to a separate project
* Only run the tests under ubasan
* Tidy up CMakeLists.txt
* Tidy up Jamfiles
* Control running with valgrind explicitly


vinniefalco:
Merged

PR #525, opened 2017-06-21
Version 64

vinniefalco:
* Simplify buffered_read_stream composed op
* Simplify ssl teardown composed op
* Simplify websocket write_op
* Exemplars are compiled code
* Better User-Agent in examples
* async_write requires a non-const message
* Doc tidying
* Add link_directories to cmake

API Changes:

* Remove make_serializer

Actions Required:

* Replace calls to make_serializer with variable declarations

Documentation preview:
http://vinniefalco.github.io/stage/beast/v64/


doc/3_4_composed.qbk
```diff
@@ -150,7 +143,85 @@ care must be taken to ensure that no access to data members are made after the
...
+  ties up a thread running on the __io_service__.
+
+* Forgetting to overload `asio_handler_invoke` for the composed
+  operation. This will cause undefined behavior if someone calls
+  the initiating function with a strand-wrapped function object,
+  and there is more than thread running on the `io_service`.
+
+* For operations which complete immediately (i.e. without calling an
+  intermediate initiating function), forgetting to use `io_service::post`
+  to invoke the final handler, breaking this guarantee of the initiating
```
harrishancock: Maybe rephrase:

> For operations which complete immediately (i.e. without calling an intermediate initiating function), forgetting to use `io_service::post` to invoke the final handler. **[full stop]** This breaks the following initiating function guarantee: ...

vinniefalco: willfix

example/echo-op/echo_op.cpp
```diff
@@ -217,18 +201,84 @@ operator()(beast::error_code ec, std::size_t bytes_transferred)
...
+//[example_core_echo_op_6
+
+// Handler hook forwarding. These free functions invoke the hooks
+// associated with the final completion handler. In effect, they
+// make the Asio implementation treat our composed operation the
+// same way it would treat the final completion handler for the
+// purpose of memory allocation and invocation.
+//
+// Our implementation just passes through the call to the hook
+// associated with the final handler.
```
harrishancock: Consider moving the comments from `asio_handler_is_continuation()` here.

> Our implementation just passes ~through~ the call **through** to the hook associated with the final handler. Note that we structure the calls to permit argument dependent lookup to take effect (`using boost::asio::asio_handler_*;`).
>
> Also, always use std::addressof to pass the pointer to the handler, otherwise an unwanted overload of operator& may be called instead.

vinniefalco: willfix

example/echo-op/echo_op.cpp
```diff
@@ -13,13 +13,24 @@
...
 //[example_core_echo_op_1
 
+template<
+    class AsyncStream,
+    class CompletionToken>
+auto
+async_echo(AsyncStream& stream, CompletionToken&& token)
+
+//]
+    -> beast::async_return_type<CompletionToken, void(beast::error_code)>;
```
harrishancock: I was originally going to say "missing semicolon," but now I see it. I presume you did the trailing return type because you're testing everything against a C++11 compiler?

Now that I see the deduced return type version live, I'm no longer sure it's less confusing, since the prototypes presented no longer match. What do you think?

vinniefalco: Right, this ensures that all snippets compile. I think `auto` looks better early in the section.

harrishancock: SGTM.

harrishancock: Haha, yes. We are bikeshedding over quite the obscurity! And the Networking TS combined with the Coroutines TS may make this entire style of coding obsolete ...

Issue #536, opened 2017-06-24
beast::http::string_view_body and boost::string_view


octopus-prime:
Using `beast::http::string_view_body` and` boost::string_view`
```
void blub()
{
    beast::http::request<beast::http::string_view_body> request;
    request.body = boost::string_view("Hello, world!");
}
```
gives this error:
```
gcc.compile.c++ bin/gcc-7/release/example/foo.o

    "g++"  -ftemplate-depth-128 -march=native -ftemplate-depth=256 -std=c++17 -O3 -finline-functions -Wno-inline -Wall -fPIC  -DBOOST_ALL_DYN_LINK -DNDEBUG  -I"/home/mike/workspace/Beast/include" -I"include" -c -o "bin/gcc-7/release/example/foo.o" "example/foo.cpp"

example/foo.cpp: In function ‘void blub()’:
example/foo.cpp:168:51: error: no match for ‘operator=’ (operand types are ‘beast::http::string_view_body::value_type {aka boost::basic_string_ref<char, std::char_traits<char> >}’ and ‘boost::basic_string_view<char, std::char_traits<char> >’)
  request.body = boost::string_view("Hello, world!");
                                                   ^
In file included from /home/mike/workspace/Beast/include/beast/core/string.hpp:11:0,
                 from /home/mike/workspace/Beast/include/beast/core/static_string.hpp:12,
                 from /home/mike/workspace/Beast/include/beast/core.hpp:29,
                 from /home/mike/workspace/Beast/include/beast.hpp:13,
                 from example/foo.cpp:8:
/usr/include/boost/utility/string_ref.hpp:77:27: note: candidate: constexpr boost::basic_string_ref<charT, traits>& boost::basic_string_ref<charT, traits>::operator=(const boost::basic_string_ref<charT, traits>&) [with charT = char; traits = std::char_traits<char>]
         basic_string_ref& operator=(const basic_string_ref &rhs) BOOST_NOEXCEPT
                           ^~~~~~~~
/usr/include/boost/utility/string_ref.hpp:77:27: note:   no known conversion for argument 1 from ‘boost::basic_string_view<char, std::char_traits<char> >’ to ‘const boost::basic_string_ref<char, std::char_traits<char> >&’
...failed gcc.compile.c++ bin/gcc-7/release/example/foo.o...
```

vinniefalco:
[...]

PR #544, opened 2017-06-26
Generic string bodies

octopus-prime:
Pull request for #536


test/http/doc_examples.cpp
```diff
@@ -307,6 +311,50 @@ class doc_http_samples_test
         BEAST_EXPECTS(! ec, ec.message());
     }
 
+    void
+    doConstAndMutableBody()
+    {
+        test::pipe c{ios_};
```
vinniefalco: Wow yeah, you understand that `test::pipe` mumbo-jumbo and you even wrote the tests - this is great!

test/http/doc_examples.cpp
```diff
@@ -307,6 +311,50 @@ class doc_http_samples_test
...
+
+        std::array<char, 15> const body {{"Hello, world!\n"}};
+        {
+            request<const_body<std::array<char, 15>>> req;
+            req.version = 11;
+            req.method(verb::put);
+            req.target("/");
+            req.body = body;
+            req.prepare_payload();
+            write(c.client, req);
```
vinniefalco: This is interesting because note that you are writing the null terminator into the body. That's one of the problems with using `std::array` as the body type, its always the same size.

example/common/const_body.hpp
```diff
@@ -14,29 +14,21 @@
...
-        has_member_function_data<T, const typename T::value_type*, boost::mpl::vector<>, boost::function_types::const_qualified>::value
-    >;
+namespace detail {
+
+template<class T, class = void>
+struct is_const_container : std::false_type { };
+
+template< class T >
+struct is_const_container<T, beast::detail::void_t<
+    decltype( std::declval<T&>().size() ),
```
vinniefalco: Nice job on the traits! @pdimov showed me this trick for checking the return value:
```
    std::declval<std::size_t&>() = std::declval<T&>().size()
```

You can do something similar for checking the return value of `data`. 
```
    std::declval<std::void const*&>() = std::declval<T&>().data()
```

example/common/mutable_body.hpp
```diff
@@ -14,33 +14,23 @@
 #include <beast/core/detail/type_traits.hpp>
 #include <boost/asio/buffer.hpp>
 #include <boost/optional.hpp>
-#include <boost/tti/has_type.hpp>
-#include <boost/tti/has_member_function.hpp>
-#include <boost/mpl/identity.hpp>
```
vinniefalco: Thank you for not dragging TTI and MPL into Beast.. lol

example/common/const_body.hpp
```diff
@@ -39,7 +39,7 @@ template <typename Container>
 struct const_body
 {
     static_assert(sizeof(typename Container::value_type) == 1, "Const character requirements not met");
-    static_assert(detail::is_const_container<Container>::value, "Const container requirements2 not met");
+    static_assert(detail::is_const_container<Container>::value, "Const container requirements not met");
```
vinniefalco: I would use the names **ConstContainer** and **MutableContainer** since those are suggestive of the concept requirements.

test/http/doc_examples.cpp
```diff
@@ -351,10 +351,41 @@ class doc_http_samples_test
...
+            BEAST_EXPECTS(res.body == std::vector<char>(body.begin(), body.end()),
+                std::string(body.begin(), body.end()));
         }
     }
 
+    void
+	checkConstBody()
+    {
+    	BEAST_EXPECTS(::detail::is_const_container<std::string>::value == true,
+    	    "is_const_container<std::string>::value");
```
vinniefalco: Consider this instead:
```
BOOST_STATIC_ASSERT(::detail::is_const_container<std::string>::value);
```

example/common/const_body.hpp
```diff
@@ -0,0 +1,100 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef BEAST_EXAMPLE_COMMON_CONST_BODY_HPP_
+#define BEAST_EXAMPLE_COMMON_CONST_BODY_HPP_
+
+#include <beast/config.hpp>
+#include <beast/http/error.hpp>
+#include <beast/http/message.hpp>
+#include <beast/core/detail/type_traits.hpp>
```
vinniefalco: Is this needed?

vinniefalco:
This should be rebased against **v68**, and you  should add the files `const_body.cpp` and `mutable_body.cpp` to this project:
https://github.com/vinniefalco/Beast/tree/v68/test/common

The .cpp file just includes the .hpp, to make sure it compiles by itself. Here's an example
https://github.com/vinniefalco/Beast/blob/v68/test/common/detect_ssl.cpp

test/http/doc_examples.cpp
octopus-prime: Something to do here? Replace 15 by 14?

example/common/const_body.hpp
octopus-prime: willfix

test/http/doc_examples.cpp
octopus-prime: yeah! willfix

example/common/const_body.hpp
octopus-prime: not sure. will check it

test/http/doc_examples.cpp
vinniefalco: Nothing to do here except to note that people using `std::array` need to be careful because the entire array will be written out no matter how big the string is.

octopus-prime:
After rebasing my branch i have to close this pull request and create a new one?!
Is my branch rebased? github says while comparing against v68 "These branches can be automatically merged."

vinniefalco:
Github often gets confused when projects attempt to keep a linear history, I wouldn't worry about it. But you included a merge commit which I typically try to avoid. Don't worry I will l fix it up.

octopus-prime:
After rebasing i get bad errors:

```
In file included from /home/mike/workspace/Beast/include/beast/http.hpp:21:0,
                 from /home/mike/workspace/Beast/include/beast.hpp:14,
                 from /home/mike/workspace/Beast/./example/doc/http_examples.hpp:8,
                 from /home/mike/workspace/Beast/test/http/doc_examples.cpp:8:
/home/mike/workspace/Beast/include/beast/http/parser.hpp: In instantiation of ‘class beast::http::parser<true, mutable_body<std::vector<char> >, std::allocator<char> >’:
/home/mike/workspace/Beast/test/http/doc_examples.cpp:336:65:   required from here
/home/mike/workspace/Beast/include/beast/http/parser.hpp:52:5: error: static assertion failed: BodyWriter requirements not met
     static_assert(is_body_writer<Body>::value,
     ^~~~~~~~~~~~~
```

Any idea why body concept check stopped working?

octopus-prime:
I followed your suggestions. So you can try another review.

vinniefalco:
Code looks great, I squashed it down and now its part of version 68 (see https://github.com/vinniefalco/Beast/pull/541) Thanks!

Issue #545, opened 2017-06-26
implicit-fallthrough warnings while builing http-tests


octopus-prime:
Beast/v68
gcc 7.0.1 -std=c++17
boost 1.63

So many warnings, you can't see real errors...

```
In file included from /home/mike/workspace/Beast/include/beast/http/serializer.hpp:333:0,
                 from /home/mike/workspace/Beast/include/beast/http.hpp:24,
                 from /home/mike/workspace/Beast/include/beast.hpp:14,
                 from /home/mike/workspace/Beast/./example/doc/http_examples.hpp:8,
                 from /home/mike/workspace/Beast/test/http/doc_examples.cpp:8:
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp: In member function ‘void beast::http::serializer<isRequest, Body, Fields, ChunkDecorator>::get(beast::error_code&, Visit&&) [with Visit = beast::http::detail::write_lambda<beast::test::pipe::stream>&; bool isRequest = false; Body = file_body; Fields = beast::http::basic_fields<std::allocator<char> >; ChunkDecorator = beast::http::no_chunk_decorator]’:
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:65:9: warning: this statement may fall through [-Wimplicit-fallthrough=]
         s_ = do_init;
         ^~
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:69:5: note: here
     case do_init:
     ^~~~
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:89:26: warning: this statement may fall through [-Wimplicit-fallthrough=]
         BOOST_FALLTHROUGH;
                          ^
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:92:5: note: here
     case do_header:
     ^~~~
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:98:9: warning: this statement may fall through [-Wimplicit-fallthrough=]
         s_ = do_header_only;
         ^~
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:99:5: note: here
     case do_header_only:
     ^~~~
/home/mike/workspace/Beast/include/beast/http/impl/serializer.ipp:110:9: warning: this statement may fall through [-Wimplicit-fallthrough=]
         s_ = do_body + 1;
         ^~
```

vinniefalco:
[...]

PR #549, opened 2017-06-27
implicit-fallthrough warnings

octopus-prime:
Issue #545


vinniefalco:
These warnings were highly annoying. But not as annoying is the travis failures caused by this:
https://github.com/vinniefalco/Beast/issues/546

PR #552, opened 2017-06-28
Bug fixed in file_service.hpp

octopus-prime:
```
            // Make sure the file is there
            if(! boost::filesystem::exists(full_path))
            {
                // Send a HEAD response
                send(head(req, full_path));
            }
            else
            {
                // Send a Not Found result
                send(not_found(req, rel_path));
            }
```

==>

```
            // Make sure the file is there
            if(boost::filesystem::exists(full_path))
            {
                // Send a HEAD response
                send(head(req, full_path));
            }
            else
            {
                // Send a Not Found result
                send(not_found(req, rel_path));
            }
```

### DIFF
--- a/example/server-framework/file_service.hpp
+++ b/example/server-framework/file_service.hpp
@@ -157,7 +157,7 @@ public:
             boost::filesystem::path full_path = root_ / rel_path;
 
             // Make sure the file is there
-            if(! boost::filesystem::exists(full_path))
+            if(boost::filesystem::exists(full_path))
             {
                 // Send a HEAD response
                 send(head(req, full_path));


vinniefalco:
That which is not tested...contains bugs!

Please change the commit message to "Fix HEAD response in file_service" and add "* Fix HEAD response in file_service" to the CHANGELOG.md

Thanks!

octopus-prime:
Added 2nd fix for BEAST_FALLTHROUGH in config.
You remember `__attribute__` instead of `_attribute__`?

;-)

vinniefalco:
AGAIN?!!!

octopus-prime:
The fix was not in v70...

vinniefalco:
>The fix was not in v70...

What happened is I assumed that the bug was only on account of 7.0.1 being pre-release and I forgot about the _attribute__ misspelling, I threw out the commit with the fix.

PR #555, opened 2017-06-28
Add options for building examples and tests.

xsacha:
Also move zlib test sources to test/zlib.

New attempt at options. CMake is cleaner now which makes it easier.


vinniefalco:
This got merged already

Issue #553, opened 2017-06-28
buffer overflow handling for string_body and mutable_body


octopus-prime:
Beast/v70

Was looking at test coverage to find untested parts to increase coverage.
So i discovered this:

```
            try
            {
                body_.resize(len + n);
            }
            catch(std::length_error const&)
            {
                ec = error::buffer_overflow;
                return;
            }
            ec.assign(0, ec.category());
```

There you catch `std::length_error` only.
What about the exceptions thrown by corresponding Allocator?
e.g. `std::bad_alloc`

http://en.cppreference.com/w/cpp/string/basic_string/resize

vinniefalco:
They are not caught!

octopus-prime:
Is this good or bad?

vinniefalco:
I think its bad, what do you propose for the fix? What exceptions should it catch, and what errors should it map them to?

octopus-prime:
I would suggest catching `std::exception` and always mapping to `error::buffer_overflow`.

vinniefalco:
I'm okay with that, would you like to submit a pull request? `mutable_body` needs it too (I think)

octopus-prime:
Yes `mutable_body` has code duplicated from `string_body`, since Reader and Writer are exactly the same... So they should use the same Reader/Writer in future?!

I can do the fixes and making a pull request.

What about this:

```
            if(content_length)
            {
                if(*content_length > (std::numeric_limits<std::size_t>::max)())
                {
                    ec = make_error_code(errc::not_enough_memory);
                    return;
                }
                ec.assign(0, ec.category());
                body_.reserve(static_cast<std::size_t>(*content_length));
            }
            else
            {
                ec.assign(0, ec.category());
            }
```

There is no try/catch at all.

vinniefalco:
>So they should use the same Reader/Writer in future?!

If you mean that `string_body` should include the header file for `mutable_body` or vice versa, no, because library code cannot use code from `example/`. It has to be duplicated. I think that's okay.

>There is no try/catch at all.

Looks like another area that needs `try` and `catch`

vinniefalco:
By the way, these are the kinds of bugs we like, easy to fix and doesn't change any public interfaces. As long as there is no major design problem with interfaces, then we're in good shape - things like this can be cleaned up over time (not saying we shouldn't fix it now but this is not as severe as an interface design flaw).

octopus-prime:
So you create a new branch first?

[...]

PR #558, opened 2017-06-29
Fix buffer overflow handling for string_body and mutable_body

octopus-prime:
issue #553

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 71:
+
+* Fix buffer overflow handling for string_body and mutable_body
+
+--------------------------------------------------------------------------------
+
 Version 70:
 
 * Serialize in one step when possible
--- a/example/common/mutable_body.hpp
+++ b/example/common/mutable_body.hpp
@@ -109,14 +109,21 @@ struct mutable_body
                 if(*content_length > (std::numeric_limits<
                         std::size_t>::max)())
                 {
-                    ec = boost::system::errc::make_error_code(
-                            boost::system::errc::not_enough_memory);
+                    ec = beast::http::error::buffer_overflow;
                     return;
                 }
-                ec.assign(0, ec.category());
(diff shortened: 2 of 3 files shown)


vinniefalco:
Is there an easy way to add a test for this? Maybe with a custom string allocator that has a small max?

octopus-prime:
custom string allocator ... i never tried before...
will have a look to find infos about...

please note:

this bugfix is the fix only. it does not contain optimizations you suggested yesterday.

and, if i add tests:

do i have to do git magic to reduce the number of commits again?

vinniefalco:
>do i have to do git magic to reduce the number of commits again?

Yes. But I can squash it for you if you want.

octopus-prime:
> But I can squash it for you if you want.

fine. would not like to make the fix for a 3rd time :D (i have to work hard on my git skills, i know)

the tests would only be possible for `mutable_body`, since `string_body` is tight bounded to `std::string`
so no chance to use a custom allocator with `string_body`

vinniefalco:
No point to `mutable_body` test, since files in `example/` are excluded from coverage.

octopus-prime:
So not adding tests?!

vinniefalco:
There's no way to test it in a meaningful way until we can control the allocator. See: https://github.com/vinniefalco/Beast/issues/559

octopus-prime:
Roger that. So you can merge after build is okay.

> There's no way to test it in a meaningful way until we can control the allocator. See: #559

That the reasons why i suggested the `mutable_body`...

vinniefalco:
>That are the reasons why i suggested the mutable_body...

Right, but `mutable_body` is a little bit too broad to be a public interface, that's why I think it belongs in example/. However, since we already have `string_body` its okay to broaden the scope a little bit to include `basic_string_body<CharT, Traits, Allocator>` for the public interfaces. This will allow us to test.

PR #562, opened 2017-07-01
Fix can/cannot thinko in FAQ.

tzlaine:
Regarding HTTP/2, the FAQ says:
"Users cannot work with HTTP/1 now; we should not deny them that functionality today to wait for a newer protocol tomorrow."

I think you meant "Users *can* work with HTTP/1 now; we should not deny them that...".


vinniefalco:
What I mean is that there is currently no standard solution for working with the HTTP protocol, which is true (not counting Beast). But your statement is also true. I will merge it

vinniefalco:
This will go into **v71**

PR #563, opened 2017-07-01
std::pair "last" -> "first" in http_message.qbk

tzlaine:


vinniefalco:
How did this get past the editor's desk!!!

doc/9_1_http_message.qbk
```diff
@@ -195,7 +195,7 @@ with arbitrary variadic lists of parameters. This allows the container
 to fully support allocators.
 
 The solution used in the library is to treat the message like a `std::pair`
-for the purposes of construction, except that instead of `first` and `last`
+for the purposes of construction, except that instead of `first` and `second`
 we have `fields` and `body`. This means that single-argument constructors
```
vinniefalco: Actually line 199 is out of date. There is no more `fields` data member of `message`. Instead, the **Fields** template argument type is now a base class of `header` (which is a base class of `message`). I'd like your input on how this can be reworded to reflect the current state of affairs.

vinniefalco:
This will go into **v71**

PR #576, opened 2017-07-02
Version 71

vinniefalco:
* Fix extra ; warning
* Documentation revision
* Fix spurious on_chunk invocation
* Call prepare_payload in HTTP example
* Check trailers in test
* Fix buffer overflow handling for string_body and mutable_body
* Concept check in basic_dynamic_body
* Tidy up http_sync_port error check

WebSockets:

* Fine tune websocket op asserts
* Refactor websocket composed ops
* Allow close, ping, and write to happen concurrently
* Fix race in websocket read op
* Fix websocket write op
* Add cmake options for examples and tests

API Changes:

* Return `std::size_t` from `Body::writer::put`

Actions Required:

* Return the number of bytes actually transferred from the input buffers in user defined `Body::writer::put` functions.


vinniefalco:
merged

PR #593, opened 2017-07-04
Prototype file_body_linux implementation

chriskohlhoff:


include/beast/http/file_body_linux.hpp
```diff
@@ -0,0 +1,256 @@
...
+//
+
+#ifndef BEAST_HTTP_FILE_BODY_LINUX_HPP
+#define BEAST_HTTP_FILE_BODY_LINUX_HPP
+
+#ifdef __linux__
+# define BEAST_NO_LINUX_FILE 0
+#else
+# define BEAST_NO_LINUX_FILE 1
+#endif
```
vinniefalco: Should we use `<boost/config.hpp>` and `BOOST_LINUX` or whatever it is? Also, can we please check if the macro is already set:
```
#ifndef BEAST_NO_LINUX_FILE
# ifdef __linux__
#  define BEAST_NO_LINUX_FILE 0
# else
#  define BEAST_NO_LINUX_FILE 1
# endif
#endif
```

include/beast/http/file_body_linux.hpp
```diff
@@ -0,0 +1,256 @@
...
+            other.file_ = -1;
+            other.size_ = 0;
+        }
+
+        /// Move assignment
+        value_type& operator=(value_type&& other) noexcept
+        {
+            value_type tmp(std::move(other));
+            std::swap(file_, tmp.file_);
+            std::swap(size_, tmp.size_);
```
vinniefalco: I think that the move assignment should specify that the moved-from object is left as-if default-constructed. You like keeping objects around (e.g. http_worker) and I imagine that the moved-from object could be left with an open file unintentionally. There's precedent for this, I believe that's how `basic_io_object` works (correct me if wrong). After a move you get a clean socket ready for reuse.

include/beast/http/file_body_linux.hpp
```diff
@@ -0,0 +1,256 @@
...
+            file_mode mode, error_code& ec)
+        {
+            int f;
+            switch(mode)
+            {
+            default:
+            case file_mode::read:   f = O_RDONLY; break;
+            case file_mode::scan:   f = O_RDONLY; break;
+            case file_mode::write:  f = O_WRONLY | O_CREAT; break;
+            case file_mode::append: f = O_WRONLY | O_APPEND; break;
```
vinniefalco: Consider these additional mode flags:
https://github.com/vinniefalco/NuDB/blob/master/include/nudb/impl/posix_file.ipp#L206

include/beast/http/file_body_linux.hpp
```diff
@@ -0,0 +1,256 @@
...
+            file_ = ::open(path.string().c_str(), f, 0777);
+            if(file_ == -1)
+            {
+                ec.assign(errno, beast::system_category());
+                return;
+            }
+            struct stat s;
+            if(::fstat(file_, &s) == -1)
+            {
+                ec.assign(errno, beast::system_category());
```
vinniefalco: `beast::` not necessary, its in `file_body_stdc` just because that code is extracted into the docs (Which doesn't assume the beast:: namespace). It doesn't hurt though. Although maybe when we put this in the boost:: namespace it could cause trouble.

include/beast/http/file_body_linux.hpp
```diff
@@ -0,0 +1,256 @@
...
+                ::close(file_);
+                file_ = -1;
+                return;
+            }
+            size_ = s.st_size;
+            ec.assign(0, ec.category());
+            signal_init<SIGPIPE>();
```
vinniefalco: Why? Just wondering. Is this unfriendly to the app? Are we taking up a signal slot or something? I don't really speak signals.

include/beast/http/file_body_linux.hpp
chriskohlhoff: The move-construction into `tmp` should already cover this.

include/beast/http/file_body_linux.hpp
chriskohlhoff: You mentioned elsewhere that you had another implementation of posix-based files you can use, so I assume you'll merge that before this becomes a real feature. I only planned to implement just enough here to support the prototype.

include/beast/http/file_body_linux.hpp
chriskohlhoff: The `sendfile()` function raises a SIGPIPE signal if the connection is broken while it is still writing data. This terminates the process by default. On linux this is suppressed for functions like `sendmsg` by passing `MSG_NOSIGNAL`, but it seems there is no equivalent for `sendfile`.

include/beast/http/file_body_linux.hpp
chriskohlhoff: I've added the extra `#ifdef` check. I think a proper feature test should be left until this gets merged properly -- it might really be better as a "posix" rather than "linux" file body, as the implementation will just differ in the write specialisations.

include/beast/http/file_body_linux.hpp
vinniefalco: We don't want `terminate` heh

vinniefalco:
I've incorporated this into an experimental branch, but there is considerable work to do on the concepts and basic implementation. I feel that we have sufficiently proved that the technique is viable so I will close this now and work on refining the basics, then revisit this for a full public interface. Thanks for helping!

vinniefalco:
I have added the **File** concept and implemented three models:
`file_stdio`, `file_posix`, and `file_win32`

The type alias `file` refers to the best choice for the platform. Users can always ask for a specific type of file like `file_stdio` if they want.

There is now a `file_body` which is an alias for `basic_file_body<file>`. It uses the **File** concept to implement the traditional reader and writer.

Finally, I have added `file_win32_body.ipp` . This specializes the `basic_file_body` class for `file_win32`. There are no user-facing types anymore (like `file_body_win32`), except of course for the specialization.  The implementation for `basic_file_body<file_win32>` is rock solid. It hooks in at the lowest layer (`write_some` and `async_write_some`). The preprocessor checks are correct, for example if they don't have overlapped (if that even matters anymore).

The implementation defaults to not using TransmitFile if the body is chunked. I added a feature, `serializer::limit` which lets the caller control the maximum number of bytes used in a call to write, and the TransmitFile implementation respects this. The code is also ready to enable sending just a range of a file (there is no interface to set it yet though).

There is a small infrastructure that provides tools to make this type of optimization easier. I had to put the two-phase construction back (i.e. `init(error_code&)` in reader and writer). The serializer allows accessing the internal instance of `reader`, so you can store state variables there instead of putting them on the message. And Beast now supports readers that want to construct with a non-const message. In other words, for algorithms which require modifying members of the body, such as hidden state. In the case of `file_body`, reading from the file counts as logically non-const since it updates the file position, even though the reader is bitwise-const.

Version 76 has the `file_body_win32`: https://github.com/vinniefalco/Beast/tree/v76

Some benchmarks:
```
http-server-fast

100,000 requests to data_64K.html
file_stdio                      1,840 req/s     118,151Kb/s
file_win32                      1,854 req/s     118,997Kb/s
file_body_win32                 2,611 req/s     167,659Kb/s
file_body_win32, tx_header      2,524 req/s     162,035Kb/s

100,000 requests to data_1K.html
file_stdio                      4,678 req/s     5,135 Kb/s
file_win32                      4,824 req/s     5,295Kb/s
file_body_win32                 4,542 req/s     4,985Kb/s
file_body_win32, tx_header      4,727 req/s     5,188Kb/s
```

"tx_reader" is an experiment in using `::TransmitFile` to also send the header. It requires a memory allocation though, to flatten the buffer sequence holding the header. Its not great enough to keep, and the code is simpler without it. As you can see, the implementation holds its own for large files although for small ones there's a tiny loss. I could add a threshold on the size of the range if you think its a good idea.

I feel really good about this, unfortunately it totally breaks the posix port but I guess you expected that. Please let me know your thoughts before I merge v76 (code review?)

vinniefalco:
@chriskohlhoff This needs to be updated, as I have changed the concepts to better support the use-case for `::sendfile` and `::Transmitfile`. I do not develop on Linux so I don't have the setup necessary to make this production quality.

Issue #585, opened 2017-07-04
cpuid / sse42 broken on gcc/clang


vinniefalco:
Seems it only works on Windows I need help getting it functioning elsewhere.

mika-fischer:
What exactly is broken?

vinniefalco:
It compiles but I dont think I got the compiler flags set to allow the intrinsics to be used. And I dont think the macros for detecting if intrinsics are available on gcc/clang are right. This all works on Windows by the way, 20% speedup. Here it is:
https://github.com/vinniefalco/Beast/blob/v72/include/beast/core/detail/cpu_info.hpp#L11

mika-fischer:
1) You need `-march=<something that has sse4.2>` or `-msse42` directly to make gcc provide these intrinsics. Probably the same with clang. MSVC doesn't have or need this.
2) You should `#include <x86intrin.h>` (GCC, clang) or `#include <intrin.h>` (MSVC) to provide the prototypes.
3) The detection is wrong, try this
```
# if defined(__SSE4_2__) || defined(_MSC_VER)
#  define BEAST_NO_INTRINSICS 0
# else
#  define BEAST_NO_INTRINSICS 1
# endif
```
4) I find the double negation confusing, why not BEAST_USE_INTRINISCS?

vinniefalco:
Could I trouble you to submit a pull request? Then Travis will kick in and we'll see if it works

>I find the double negation confusing, why not BEAST_USE_INTRINISCS?
That's the Boost style

octopus-prime:
Don't use the __asm stuff! You can use this for gcc and clang:

```
#include <cpuid.h>

int main()
{
    std::uint32_t id = 0;
    std::uint32_t eax = 0;
    std::uint32_t ebx = 0;
    std::uint32_t ecx = 0;
    std::uint32_t edx = 0;

    bool supported = __get_cpuid(id,  &eax,  &ebx,  &ecx, &edx);

    std::cout << std::boolalpha << supported << std::endl;
    std::cout << eax << std::endl;
    std::cout << ebx << std::endl;
    std::cout << ecx << std::endl;
    std::cout << edx << std::endl;

    return 0;
}
```

Tested with
- Apple LLVM version 7.3.0 (clang-703.0.31)
- gcc 4.8.1 (wandbox)

vinniefalco:
-> https://github.com/vinniefalco/Beast/pulls

mika-fischer:
[...]

PR #612, opened 2017-07-06
Using SSE4.2 intrinsics in basic_parser if available

octopus-prime:
issue #585

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 cmake_minimum_required (VERSION 3.5.2)
 
-project (Beast VERSION 73)
+project (Beast VERSION 74)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 option (Beast_BUILD_EXAMPLES "Build examples" ON)
--- a/include/beast/version.hpp
+++ b/include/beast/version.hpp
@@ -18,7 +18,7 @@
     This is a simple integer that is incremented by one every time
     a set of code changes is merged to the master or develop branch.
 */
-#define BEAST_VERSION 73
+#define BEAST_VERSION 74
 
 #define BEAST_VERSION_STRING "Beast/" BOOST_STRINGIZE(BEAST_VERSION)
 


glenfe:
Why is this assuming that __get_cpuid exists just as long as the compiler is not MSVC (or some compiler pretending to be MSVC)?

glenfe:
Why is this assuming that __get_cpuid exists just as long as the compiler is not MSVC (or some compiler pretending to be MSVC)?

vinniefalco:
This has to be rebased against the **v74** which I am about to put up - GitHub is behaving weirdly because these two branches are similar.

vinniefalco:
Please continue the discussion here https://github.com/vinniefalco/Beast/pull/613

octopus-prime:
@glenfe Since currently supported compilers are msvc, gcc and clang. IMO gcc and clang are happy with `__get_cpuid`

vinniefalco:
@glenfe What's a better way?

octopus-prime:
> This has to be rebased against the v74 which I am about to put up - GitHub is behaving weirdly because these two branches are similar.

any action needed by me?

vinniefalco:
>any action needed by me?

Yes, just rebase it interactively (`git fetch upstream` then `git rebase -i upstream/v74` where `upstream` is the remote `git@github.com:vinniefalco/Beast.git`), delete the commits that aren't yours, and accept

glenfe:
Supporting only three compilers is perfectly fine, but let's not assume that if not compiler X detected, that it must be one of Y or Z. i.e. The way the original code was detecting C++ implementations (i.e. an error directive with a meaningful message if not one of the supported compilers) was better.

For something like this, it isn't a big deal, because ultimately it will just result in a compile time error anyway, but in general as a good practice for Boost libraries: If you're targeting implementation specifics, do it only after you've explicitly detected the implementation.

glenfe:
Also, prefer use BOOST_MSVC from Boost.Config to determine if the compiler is truly MSVC, because compilers that choose to pretend to be MSVC may define _MSC_VER, and may not have the same behavior (or the same intrinsics).

octopus-prime:
@glenfe 

Can you give a working example for this pseudocode using the right BOOST_xxx macros?

```
if windows
...
else if gcc or clang
...
else
...
end
```

vinniefalco:
I am confused on how to do this as well.

glenfe:
For just that pseudo code:
```
#if defined(BOOST_MSVC) && BOOST_MSVC >= x
// ...
#elif (defined(BOOST_GCC) && BOOST_GCC > y) || (defined(BOOST_CLANG) && BOOST_CLANG > z)
// ...
#else
// ...
#endif
```
Of course, within each implementation detection block, you will want to check additional implementation specific things.

octopus-prime:
@glenfe Thanks!
@vinniefalco If we want to use these `x, y and z` versions - what are the minimal requirements for the compiler versions? I only found `When using Microsoft Visual C++, Visual Studio 2015 Update 3 or later is required.` So what versions are required for gcc and clang?

BTW:
Does anybody know whether we can use the clang version on OSX too?
The Apple-LLVM compiler says something about version 7.x and 8.x.
And http://llvm.org/ says something about a LLVM 5.0 will come **in future**.
So clang versions on OSX are useless??

vinniefalco:
Beast needs good C++11 support, so Visual Studio 2015 Update 3 is a reasonable minimum.

Right now the farthest back Beast goes on gcc is 4.8. That's probably a reasonable lower limit, any lower and you start running into compiler bugs that are very hard to fix in Beast.

For clang, I don't know.

octopus-prime:
Ignoring the compiler versions for the moment.
When we change the definition of `BEAST_NO_INTRINSICS` to

```
#ifndef BEAST_NO_INTRINSICS
# if defined(BOOST_MSVC) || ((defined(BOOST_GCC) || defined(BOOST_CLANG)) && defined(__SSE4_2__))
#  define BEAST_NO_INTRINSICS 0
# else
#  define BEAST_NO_INTRINSICS 1
# endif
#endif
```

and all other cpuid and SSE4.2 stuff will guarded by `!BEAST_NO_INTRINSICS` so we are safe with

```
#if defined(BOOST_MSVC)
#include <intrin.h> // __cpuid
#else
#include <cpuid.h>  // __get_cpuid
#endif
```

?!

@glenfe @vinniefalco What do you think?

vinniefalco:
I don't know how any of those BOOST macros work, and I am unfamiliar with the other toolchains' macros. So I cannot evaluate this change.

PR #618, opened 2017-07-07
removed redundant flush() from example

akrzemi1:
`<<std::endl` already flushes the stream; no need for a separate call.


vinniefalco:
Part of **v74** now - thank you!

PR #620, opened 2017-07-07
Version 75

vinniefalco:
* Use file_body for valid requests, string_body otherwise.
* Construct buffer_prefix_view in-place
* Shrink serializer buffers using buffers_ref
* Tidy up BEAST_NO_BIG_VARIANTS
* Shrink serializer buffers using buffers_ref
* Add serializer::limit

Documentation preview:
http://vinniefalco.github.io/stage/beast/v75

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Version 75:
 * Tidy up BEAST_NO_BIG_VARIANTS
 * Shrink serializer buffers using buffers_ref
 * Add serializer::limit
+* file_body tests
 
 --------------------------------------------------------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Version 75:
 
 * Use file_body for valid requests, string_body otherwise.
+* Construct buffer_prefix_view in-place
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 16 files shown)


vinniefalco:
>SSE4.2 still does not work.

I don't know how to fix it correctly

> I don't understand why the responses (string_response/file_response) must be member of http_worker 

Try running a benchmark of http-server-fast compared to some of the other servers. Its easily 35% to 60% faster. It might look strange but it helps performance.

> double free or corruption (out): 0x00007ff7140008c0 ***

No symbols? Did you try bisecting?

>I don't understand too much of serialized.ipp...
>No chance to use some other technique here?

I don't think so. It is what it is, this was the most direct way to express the intent of the implementation. I need debug symbols are more information if I am going to figure out why it is crashing for you.

>What are e.g. do_body, do_body + 2, etc?

Those are the steps in the state machine for advancing the buffers and preparing the new buffers. Its all quite optimized.

>leading to fall-through warning.

I thought we fixed those warnings?

octopus-prime:
> I thought we fixed those warnings?

I believe is gcc 7.0.1 bug. Nevertheless: no switch/case - on fall-through ;-)

> Its easily 35% to 60% faster. It might look strange but it helps performance.

That's incredible much. So okay - responses are members ;-)

> I don't know how to fix it correctly

I think the SSE4.2 pull request was good enough for the moment and could be refined later.
So somebody must request a 3rd pull - rebased on a new branch. That makes contributors tired.

> No symbols? Did you try bisecting?

The problem here is to reproduce the crash. The test does not crash always. It's like a "defective contact". So may be it's something in multithreading...

vinniefalco:
>I think the SSE4.2 pull request was good enough for the moment and could be refined later.

I'll rebase it.

PR #622, opened 2017-07-08
Using SSE4.2 intrinsics in basic_parser if available

octopus-prime:
Still issue #585


vinniefalco:
Do the CMakeLists.txt and Jamroot correctly set the compiler switches to enable the instructions?

octopus-prime:
I don't know cmake voodoo, since i use bjam.

I have to modify Jamroot for building.
And I need to build before commit - for testing.
So I can't commit Jamroot.

3 pull requests are enough. Somebody else can enable SSE4.2 for gcc and clang now.

vinniefalco:
I feel your pain

PR #652, opened 2017-07-15
beast::detail::is_invocable: Move Only Parameters

RobertLeahy:
Previously reported that function objects taking move only types by value were not invocable.


test/core/type_traits.cpp
```diff
@@ -37,6 +38,11 @@ struct is_invocable_udt3
     int operator()(int);
 };
 
+struct is_invocable_udt4
+{
+    void operator()(std::unique_ptr<int>);
```
vinniefalco: **NOICE!!!**

vinniefalco:
This will go into **version 82**, thanks!

PR #654, opened 2017-07-15
Version 82

vinniefalco:
* Documentation tidying
* is_invocable works with move-only types
* Use std::function and reference wrapper
* Add session_alloc to example/common

WebSocket

* stream tidying
* Add rd_close_ to websocket stream state
* stream uses flat_buffer
* accept requires a message
* Add wstest benchmark tool

API Changes:

* Rename to flat_static_buffer and flat_static_buffer_base

Actions Required:

* Rename static_buffer to flat_static_buffer_base
* Rename static_buffer_n to flat_static_buffer

Documentation preview:
http://vinniefalco.github.io/stage/beast/v82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 
 WebSocket
 
-* websocket::stream tidying
+* stream tidying
 * Add rd_close_ to websocket stream state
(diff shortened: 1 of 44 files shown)


vinniefalco:
Merged

PR #662, opened 2017-07-17
fixed build error (src/test/benchmarks/utf8_checker.cpp:99:21: error: 'clock_type' is a private member of 'beast::utf8_checker_test::timer')

adrianimboden:
When I tried to build the project (clang 4.0.1), this compilation error occurred.

### DIFF
--- a/test/benchmarks/utf8_checker.cpp
+++ b/test/benchmarks/utf8_checker.cpp
@@ -22,9 +22,11 @@ public:
 
     class timer
     {
+    public:
         using clock_type =
             std::chrono::system_clock;
 
+    private:
         clock_type::time_point when_;
 
     public:


vinniefalco:
That's bizarre... which branch is this? Why doesn't the error show up on CI?

adrianimboden:
I cloned the latest develop branch. The only reason I could think of, is that the CI does not build the performance tests.

include/beast/websocket/stream.hpp
```diff
@@ -627,6 +627,14 @@ class stream
         return wr_opcode_ == detail::opcode::text;
     }
 
+
+    /// Returns `true` if either the server or the client sent a close frame.
+    bool 
+    closed() const
+    {
+        return wr_close_ || rd_close_;
```
vinniefalco: This will be unnecessary soon. If `rd_close_` is ever true, it means that `websocket::error::closed` has already been delivered to the caller. If `wr_close_` is true, it means the caller invoked `close` or `async_close`. And soon, those functions will also drain the connection and deliver `websocket::error::closed` to the caller to indicate a proper closure. So this function won't make much sense.

adrianimboden: Ah perfect. I got caught by this, because the websocket::stream::close method currently asserts when the socket is already closed.

vinniefalco: You shouldn't be calling `async_close` twice.  If there is a scenario you can figure out that causes `async_close` to assert even when you only call it once, then that's a bug and needs to be fixed. The only case I can think of is if the implementation receives a close frame, sends one in response, and then before it delivers `websocket::error::close`, the application calls `async_close`. I can't prove that its possible though.

Now that I look over the async code it appears it is missing the assert present in the synchronous code, i will open an issue.

And I will refactor the close code to be more robust.

vinniefalco: Now that I look at it more closely, there's a smell coming out of this close code...you are on to something!

vinniefalco: Are you using the synchronous version of `close`?

adrianimboden: Yes, I use the synchronous version of close. But only in the shutdown case, everything else is asynchronous.

the function stream<NextLayer>::close(close_reason const& cr, error_code& ec) asserts just before checking it to set the error code

I did some debugging yesterday, and I'm pretty sure, that the close function wasn't called twice. It was called just after after upgrading the connection, but before transmitting or receiving any data yet, if I remember correctly.

PR #709, opened 2017-08-01
Use unique name for test project

danieljames:
See: https://lists.boost.org/boost-testing/2017/08/8393.php


vinniefalco:
This is merged in version 98, thank you for the submission!

PR #720, opened 2017-08-06
Fix header image URL in server-framework README

agnat:
All said. Looking forward to trying out beast. Thanks and congrats, @vinniefalco.

### DIFF
--- a/example/server-framework/README.md
+++ b/example/server-framework/README.md
@@ -1,5 +1,5 @@
 <img width="880" height = "80" alt = "Beast"
-    src="https://raw.githubusercontent.com/boostorg/beast/master/doc/images/readme.png">
+    src="https://raw.githubusercontent.com/boostorg/beast/master/doc/images/readme2.png">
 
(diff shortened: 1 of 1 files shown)


vinniefalco:
I have a confession to make, that **server-framework** example is terribly over-engineered. It is slow to compile and simply does too much. I'm working on a top to bottom rewrite of all the examples to make them simple and self contained (each in just one file). I will preserve some of the cool individual features of the server-framework example: pipelining, SSL detection, WebSocket+HTTP on the same port, but it will be broken up across several examples instead of all piled into one.

I will probably get rid of that "PortHandler" concept since its a poor implementation of HTTP server request-routing, I think it is doing more harm than good. I appreciate your effort at trying to tidy up the repository though! Thanks!

agnat:
Ok. Wow. Just let me browse the *new* examples...

vinniefalco:
**advanced-server** and **advanced-server-ssl** are the most similar to the old **server-framework**.

I would love to hear what you think of these new examples, hopefully there is something there you can use!

PR #724, opened 2017-08-08
V101

vinniefalco:
* Refactor all examples


vinniefalco:
@agnat The formula was correct but there was another big. I've fixed it and factored the queue to work better:
https://github.com/vinniefalco/beast/blob/v103/example/advanced/server/advanced_server.cpp#L390

PR #739, opened 2017-08-14
fixed compilation error when using clang 4.0.1 with -std=c++1z

adrianimboden:


vinniefalco:
The reason I added the braces is because without them, they fail to compile on one of these configurations (I can't remember which one):
http://www.boost.org/development/tests/develop/developer/beast.html

I think the only good way to fix this to make it work on all the different compilers is to introduce local variables. I will work on a fix that uses this style.

vinniefalco:
Can you please see if this is resolved in **version 106**, which was merged to *develop*?

adrianimboden:
Yes, works perfectly

PR #753, opened 2017-08-28
Add lowest_layer() to test::stream

djarek:
This allows the stream to be used as the next layer type in a stack of network stream layers.

Signed-off-by: Damian Jarek <damian.jarek93@gmail.com>


vinniefalco:
Hmmmmmmmmmmm......................... good idea!!!

vinniefalco:
Part of #754 now

PR #766, opened 2017-09-05
Remove field_range copy constructor

HowardHinnant:
*  This presence of the user-declared copy constructor makes
   the compiler-provided copy assignment operator deprecated.
   This change allows the compiler to provide both copy members
   without deprecated behavior.


vinniefalco:
Merged in version 113

Issue #723, opened 2017-08-07
http_server_fast (http server file range support)


maytrue:
I modify http_server_fast demo, run  as http server  and return media data from file to vlc player.

I parse "Range: bytes=" to support range, and player can play well, but stop when seek to end.

http_server_fast stop at 
![screen shot 2017-08-07 at 4 26 36 pm](https://user-images.githubusercontent.com/1849557/29021768-343ebc82-7b99-11e7-8c86-0bf6624fc00e.png)

vinniefalco:
The `file_body` does not currently support serving a range, but I have done some of that work already. I can add a function which lets you specify the range you want to serve. I was thinking something like this:
```
template<class File>
class basic_file_body<File>::value_type
{
    ...
    /// Set the range of bytes to deliver
    void range(std::size_t first, std::size_t last);
```

To work around your issue, try this instead:
```
if(nread == 0)
    return boost::none;
```

Thanks

maytrue:
It  works, thanks

vinniefalco:
I'll keep this open since I do plan to add built-in support for "range" on `file_body`.

ecorm:
> I'll keep this open since I do plan to add built-in support for "range" on file_body.

Was range support for `file_body` ever implemented? I'm scouring through the documentation and source code and can't find such support.

ecorm:
I should add that I don't really need this range request support right now; I was just curious to know if it was supported.

ashtum:
@ecorm No, there is no support for specifying a range in basic_file_body. But we can add this if you have a use case for it.

ecorm:
> @ecorm No, there is no support for specifying a range in basic_file_body. However we can add this if you have a use case for it.

The use case would obviously be video streaming: https://developer.mozilla.org/en-US/docs/Web/Media/Audio_and_video_delivery/Setting_up_adaptive_streaming_media_sources

I currently don't have a need for video streaming in my project, but I'm surprised range support hasn't been requested more by others.

A Beast-provided way of parsing the Range field would go nicely with file_body ranges if the latter were to be implemented.

vinniefalco:
> The use case would obviously be video streaming

Right, but are people using Beast to build video streamers? I doubt it, as there are already commercial solutions which Beast cannot hope to compete with. What I mean when I say use-case, is the specific motivating example which informs us on how to proceed (going in order of least design work to most design work):

1. Define this as "responsibility of the user",
2. Provide a working example using best practices upon which users can build, or
[...]

PR #786, opened 2017-09-20
Allow serving the same file multiple times in parallel

ngladitz:
Without this change the second opening of an
already opened file fails with:

  "The process cannot access the file because it
  is being used by another process"

### DIFF
--- a/include/boost/beast/core/impl/file_win32.ipp
+++ b/include/boost/beast/core/impl/file_win32.ipp
@@ -121,6 +121,7 @@ open(char const* path, file_mode mode, error_code& ec)
         h_ = boost::detail::winapi::INVALID_HANDLE_VALUE_;
     }
     boost::detail::winapi::DWORD_ desired_access = 0;
+    boost::detail::winapi::DWORD_ share_mode = 0;
     boost::detail::winapi::DWORD_ creation_disposition = 0;
     boost::detail::winapi::DWORD_ flags_and_attributes = 0;
 /*
@@ -139,6 +140,8 @@ open(char const* path, file_mode mode, error_code& ec)
     case file_mode::read:
         desired_access = boost::detail::winapi::GENERIC_READ_;
 
+        share_mode = boost::detail::winapi::FILE_SHARE_READ_;
+
         creation_disposition = boost::detail::winapi::OPEN_EXISTING_;
 
         flags_and_attributes = 0x10000000; // FILE_FLAG_RANDOM_ACCESS
@@ -146,6 +149,8 @@ open(char const* path, file_mode mode, error_code& ec)
 
     case file_mode::scan:           
         desired_access = boost::detail::winapi::GENERIC_READ_;
+
+        share_mode = boost::detail::winapi::FILE_SHARE_READ_;
 
         creation_disposition = boost::detail::winapi::OPEN_EXISTING_;
 
@@ -209,7 +214,7 @@ open(char const* path, file_mode mode, error_code& ec)
     h_ = ::CreateFileA(
         path,
         desired_access,
-        0,
+        share_mode,
         NULL,
         creation_disposition,
         flags_and_attributes,
--- a/include/boost/beast/core/impl/file_win32.ipp
+++ b/include/boost/beast/core/impl/file_win32.ipp
@@ -120,9 +120,9 @@ open(char const* path, file_mode mode, error_code& ec)
         boost::detail::winapi::CloseHandle(h_);
         h_ = boost::detail::winapi::INVALID_HANDLE_VALUE_;
     }
-    boost::detail::winapi::DWORD_ dw1 = 0;
-    boost::detail::winapi::DWORD_ dw2 = 0;
-    boost::detail::winapi::DWORD_ dw3 = 0;
+    boost::detail::winapi::DWORD_ desired_access = 0;
+    boost::detail::winapi::DWORD_ creation_disposition = 0;
+    boost::detail::winapi::DWORD_ flags_and_attributes = 0;
 /*
                              |                    When the file...
     This argument:           |             Exists            Does not exist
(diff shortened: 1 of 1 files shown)


vinniefalco:
I like it, but this needs to use a constant from Boost.WinAPI and not FILE_SHARE_READ from `<windows.h>`. So you just need to change the constant to `boost::detail::winapi::FILE_SHARE_READ_`

Thanks!

ngladitz:
Thanks! I updated the PR accordingly.

vinniefalco:
This looks very good! So you're serving files? Are you using `file_body`? Is it working? Did you know it uses `::TransmitFile` on Windows?

ngladitz:
Yes, I am using `file_body`. Except for the issue in this MR it seems to be working well for me.
I wasn't aware that it uses `::TransmitFile` but that sounds awesome. Thanks!

One thing I am unsure about (but haven't tested either) is that the implementation is using `CreateFileA` (which is "ANSI" i.e. code page specific) but the reference seems to suggest the path is encoded in UTF-8.

vinniefalco:
>is that the implementation is using CreateFileA (which is "ANSI" i.e. code page specific) but the reference seems to suggest the path is encoded in UTF-8.

Yeah...eventually I will have to fix this. Pull requests welcome :)

vinniefalco:
What happens with this change if you open the file for writing? That should be exclusive mode (i.e. produce an error)

vinniefalco:
>I wasn't aware that it uses ::TransmitFile

If you're not aware, then it is probably working :)

vinniefalco:
Your commit is part of version 118: https://github.com/boostorg/beast/pull/794

ngladitz:
> What happens with this change if you open the file for writing? That should be exclusive mode (i.e. produce an error)

With the change opening a file twice for writing should fail but I think it would allow one open for reading and one open for writing to the same file.

vinniefalco:
Opening a file twice for writing should always fail. However, I think opening a file for writing and also opening it for reading within the same process should also fail...

ngladitz:
I've create a new set of commits.
The first is cosmetic (I can undo it if you dislike it).
The second only makes use of `FILE_SHARE_READ_` when the file is opened for reading.

This should restore the error behavior in case the file is opened once for reading and once for writing.

I think the Posix implementation currently still allows shared access under all circumstances.

include/boost/beast/core/impl/file_win32.ipp
```diff
@@ -120,9 +120,10 @@ open(char const* path, file_mode mode, error_code& ec)
         boost::detail::winapi::CloseHandle(h_);
         h_ = boost::detail::winapi::INVALID_HANDLE_VALUE_;
     }
-    boost::detail::winapi::DWORD_ dw1 = 0;
-    boost::detail::winapi::DWORD_ dw2 = 0;
-    boost::detail::winapi::DWORD_ dw3 = 0;
+    boost::detail::winapi::DWORD_ desired_access = 0;
+    boost::detail::winapi::DWORD_ share_mode = 0;
+    boost::detail::winapi::DWORD_ creation_disposition = 0;
+    boost::detail::winapi::DWORD_ flags_and_attributes = 0;
```
vinniefalco: This looks good although I would put `share_mode` before `desired_access`

include/boost/beast/core/impl/file_win32.ipp
```diff
@@ -137,66 +138,86 @@ open(char const* path, file_mode mode, error_code& ec)
...
-        dw1 = boost::detail::winapi::GENERIC_READ_ |
-              boost::detail::winapi::GENERIC_WRITE_;
-        dw2 = boost::detail::winapi::CREATE_ALWAYS_;
-        dw3 = 0x08000000; // FILE_FLAG_SEQUENTIAL_SCAN
+        desired_access = boost::detail::winapi::GENERIC_READ_ |
+                         boost::detail::winapi::GENERIC_WRITE_;
+
+        creation_disposition = boost::detail::winapi::CREATE_ALWAYS_;
+
+        flags_and_attributes = 0x08000000; // FILE_FLAG_SEQUENTIAL_SCAN
```
vinniefalco: Perhaps we should add `boost::detail::winapi::FILE_FLAG_SEQUENTIAL_SCAN_`? It wouldn't become available until Boost 1.66.0 but can get it in now, see: https://github.com/boostorg/winapi/tree/develop/include/boost/detail/winapi

vinniefalco:
It looks like you're on a roll, and getting good use from `file_body`. What do you think about this issue: https://github.com/boostorg/beast/issues/723

include/boost/beast/core/impl/file_win32.ipp
ngladitz: Both `FILE_FLAG_RANDOM_ACCESS_` and `FILE_FLAG_SEQUENTIAL_SCAN_`?

Not entirely sure what you mean by getting them now.
Copy the headers into Beast? Copy just those definitions? Require consumers to get the pre-release headers?

ngladitz:
> It looks like you're on a roll, and getting good use from file_body. What do you think about this issue: #723

I could think of a use case or two for this but it might be a while before I get around to it.

e.g. I've got large archives (files) containing deflated streams. I was pondering serving those streams individually and having the browser transparently inflate them (where supported). Might require decorating the data (not sure if you can e.g. just stick a gzip header in front of a deflate stream).

vinniefalco:
These examples also have an extraneous strand for `listener`:
* http-server-async-ssl
* websocket-server-async
* websocket-server-async-ssl

ngladitz:
> These examples also have an extraneous strand for listener:
> 
>     http-server-async-ssl
>     websocket-server-async
>     websocket-server-async-ssl

Different PR/Issue presumably?

vinniefalco:
>Different PR/Issue presumably?

I would just put them all into the one commit that already removes the strand

ngladitz:
> I would just put them all into the one commit that already removes the strand

Sorry I mean what do strands have to do with the file_win32 changes (i.e. this PR)?

vinniefalco:
oops...looks like I commented in the wrong commit!!!

PR #789, opened 2017-09-21
Fix writing headers into std::ostream

inetic:


vinniefalco:
Whoops! I wonder why clang and gcc didn't generate a compile error?

vinniefalco:
Ah, probably because it is a template function and it is not called! I appreciate the pull request.

Just a hint, but if you want this pull request to be even more amazing, consider adding:

* A small test (in test/beast/http/write.cpp) which constructs a message and writes it to a `std::stringstream`, checks the result.

vinniefalco:
Your commit is part of version 118: https://github.com/boostorg/beast/pull/794

inetic:
Thanks. You're right, a test for it would be nice. I'll try to add one when time permits.

vinniefalco:
@inetic I already added one, it was easy - thanks!

vinniefalco:
These examples also have an extraneous strand for listener:

* http-server-async-ssl
* websocket-server-async
* websocket-server-async-ssl

vinniefalco:
Merged, thanks!

PR #803, opened 2017-10-03
Fix enum and non-enum in conditional exp. gcc warning

hoditohod:
Fix "enumeral and non-enumeral type in conditional expression” GCC warning. 
Identical to line 364.

### DIFF
--- a/include/boost/beast/websocket/impl/read.ipp
+++ b/include/boost/beast/websocket/impl/read.ipp
@@ -1075,7 +1075,8 @@ loop:
                 // _Start the WebSocket Closing Handshake_
                 do_fail(
                     cr.code == close_code::none ?
-                        close_code::normal : cr.code,
+                        close_code::normal : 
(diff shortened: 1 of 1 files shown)


vinniefalco:
I wonder why this warning doesn't appear on Travis (or does it?) Either way, I will merge it.

hoditohod:
GCC seems to be bit buggy with this warning:
- I have 2 build configurations: one optimized release, and another with bullseye coverage for tests. The warning is only emitted with the second build, though the related compiler flags are the same
- Beast is included as system include, gcc isn't supposed to emit a warning from beast headers, but it does
- the warning itself doesn't state the exact line of the issue, only the list of instantiated templates. It took me some time to figure out which statement triggered it.

I saw the same static_cast a few lines above, so I thought it might be worth to fix upstream, hence the PR.

vinniefalco:
Merged, thank you!

Issue #764, opened 2017-09-04
Refactor HTTP operations


vinniefalco:
* Use stackless coroutines
* Don't persist the output area of dynamic buffers in the composed operation state
* Increase code coverage

vinniefalco:
They should use stackless fauxroutines and avoid post where possible

PR #810, opened 2017-10-07
[WIP] Refactor HTTP async* operations

djarek:
## Changes:
- Use stackless coroutines instead of hand-rolled state machines.
- Added the ```is_continuation()``` helper to retrieve the state of a stackless coroutine.
- Added the ```prepare_or_error()``` helper to avoid issues with yield labels crossing initialization of catch blocks.
- Remove caching of mutable buffer in ```read_some_op```.
- Renamed ```bytes_transferred``` to ```total_bytes_consumed_``` in ```read_some_op``` since it stored the number of bytes consumed by the parser.

## To Do:
- [ ] write_some_op
- [ ] write_msg_op
- [ ] write_op

Resolves: #764


include/boost/beast/core/coroutine.hpp
```diff
@@ -0,0 +1,30 @@
...
+namespace boost {
+namespace beast {
+
+inline bool is_continuation(asio::coroutine& coroutine) noexcept
+{
+    asio::detail::coroutine_ref ref{coroutine};
+    // We need to reassign the current coroutine state, because
+    // coroutine_ref marks the coroutine as complete if it is not modified.
+    auto state = (ref = static_cast<int>(ref));
+    return state > 0;
```
vinniefalco: Oh damn, I see what you did here, this is very smart!

*BUT*..... we can't be peeking into the `asio::detail::` namespace. And I don't think this should be a public interface.

An alternative is to just copy all the Asio stackless coroutine code and then it becomes ours to do with as we please.

vinniefalco: If you DO add a public header file you need to add it to the "convenience header", in this case `<boost/beast/core.hpp>`.

include/boost/beast/http/impl/read.ipp
```diff
@@ -34,17 +35,30 @@ namespace detail {
 
 //------------------------------------------------------------------------------
 
+template<typename DynamicBuffer>
+boost::optional<typename DynamicBuffer::mutable_buffers_type>
+prepare_or_error(DynamicBuffer& buffer, error_code& ec)
```
vinniefalco: I like this very much, this would be helpful to use everywhere. It should probably be available as a `detail` function, maybe in `<boost/core/detail/type_traits.hpp>`

include/boost/beast/http/impl/read.ipp
```diff
@@ -34,17 +35,30 @@ namespace detail {
...
+{
+    try
+    {
+        ec.assign(0, ec.category());
```
vinniefalco: If there's an error we are paying for clearing `ec` needlessly. It might be would be more efficient to clear the error if the call to `prepare` succeeds. We would have to study the output in Compiler Explorer (https://godbolt.org/)

include/boost/beast/http/impl/read.ipp
```diff
@@ -34,17 +35,30 @@ namespace detail {
...
+        return {buffer.prepare(read_size_or_throw(buffer, 65536))};
+    }
+    catch(const std::length_error& error)
+    {
+        ec = error::buffer_overflow;
+        return {};
```
vinniefalco: Does this return `boost::none` or does it return an engaged, default-constructed `mutable_buffers_type`?

include/boost/beast/http/impl/read.ipp
```diff
@@ -34,17 +35,30 @@ namespace detail {
...
+template<typename DynamicBuffer>
+boost::optional<typename DynamicBuffer::mutable_buffers_type>
+prepare_or_error(DynamicBuffer& buffer, error_code& ec)
+{
+    try
+    {
+        ec.assign(0, ec.category());
+        return {buffer.prepare(read_size_or_throw(buffer, 65536))};
+    }
+    catch(const std::length_error& error)
```
vinniefalco: `error` is the name of a public symbol, FYI, you might want to rename it to `e`

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
     error_code ec,
     std::size_t bytes_transferred)
 {
-    switch(state_)
+    // Temporary variables, not to be used across yields.
+    // They have to be declared here due to initialization & yield issues.
+    size_t consumed = 0;
```
vinniefalco: `std::size_t`

include/boost/beast/core/coroutine.hpp
vinniefalco: While I like the idea, we have a problem. I have another branch waiting which updates Beast to use the new version of Boost.Asio which is styled after the Networking-TS. See: https://github.com/vinniefalco/beast/commits/net-ts

The new version of Boost.Asio removes the `asio_is_continuation` hook and replaces it with free functions `post` and `defer` which perform similar behaviors. But of course it completely changes the program logic. So I don't think its a good idea for us to change the `is_continuation` code otherwise we will get pretty bad merge conflicts. This change will have to wait until after the new branch is merged.

That said, I like the technique of leveraging the `coroutine` state to know if we are already in the proper context. I like it so much that I think it should be a member function of `asio::coroutine`. I think you should submit it as a pull request.

include/boost/beast/http/impl/read.ipp
```diff
@@ -18,6 +18,7 @@
 #include <boost/beast/core/handler_ptr.hpp>
 #include <boost/beast/core/read_size.hpp>
 #include <boost/beast/core/type_traits.hpp>
+#include <boost/beast/core/coroutine.hpp>
```
vinniefalco: I don't think this should be public (and it has other problems mentioned earlier) but for future reference when we add new public interfaces, it is best done in its own individual commit so that people who look at the changes will know that there's a new feature. Also, the docs need to be updated.

include/boost/beast/http/impl/read.ipp
```diff
@@ -101,6 +115,12 @@ public:
         asio_handler_invoke(
             f, std::addressof(op->h_));
     }
+
+    void do_upcall(error_code ec) const
+    {
+        s_.get_io_service().post(
+            bind_handler(std::move(*this), ec, 0));
```
vinniefalco: I prefer to do this explicitly inside the main function rather than as a separate function. Also I would not mark it `const`. Beast does not interpret `const` as a "bitwise const" as you have done here. Instead, `const` is roughly interpreted as "thread safe" - which this function is not.

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
     error_code ec,
     std::size_t bytes_transferred)
 {
-    switch(state_)
+    // Temporary variables, not to be used across yields.
+    // They have to be declared here due to initialization & yield issues.
```
vinniefalco: I know you mean well but this comment is not necessary, as we can assume the reader already understands the price of using stackless coroutines.

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
...
-            goto do_upcall;
-        ec.assign(0, ec.category());
-    }
-
-    do_read:
-        try
-        {
-            mb_.emplace(b_.prepare(
-                read_size_or_throw(b_, 65536)));
+            BOOST_ASIO_CORO_YIELD do_upcall(ec);
```
vinniefalco: I would prefer to put `BOOST_ASIO_CORO_YIELD` on its own line

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
...
         }
-        catch(std::length_error const&)
+        while (!ec && total_bytes_consumed_ == 0)
```
vinniefalco: I think the order should be reversed, i.e.
```
while(total_bytes_consumed_ == 0 && ! ec)
```

Rationale: `total_bytes_consumed_` will be non zero far more frequently than `ec` will be set.

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
...
-    do_upcall:
-        if(state_ >= 2)
-            goto upcall;
-        state_ = 3;
-        return s_.get_io_service().post(
-            bind_handler(std::move(*this), ec, 0));
+            mb = prepare_or_error(b_, ec);
+            if (ec)
+            {
+                BOOST_ASIO_CORO_YIELD do_upcall(ec);
```
vinniefalco: I would prefer to put BOOST_ASIO_CORO_YIELD on its own line. And I don't think we should have `do_upcall` as a separate function I think its more clear just inlining the call to `io_service::post`.

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
...
+                break;
+            }
 
-    case 3:
-        break;
+            BOOST_ASIO_CORO_YIELD s_.async_read_some(*mb, std::move(*this));
```
vinniefalco: Separate line please

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
...
+                else
+                {
+                    ec = error::end_of_stream;
+                }
+                break;
+            }
+            if (ec)
+            {
+                // No need to call do_upcall() - we're already being executed
+                // within the io_service's context.
```
vinniefalco: This comment is not necessary, the reader should already know this.

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +132,66 @@ operator()(
...
+                break;
+            }
+            b_.commit(bytes_transferred);
+            consumed = p_.put(b_.data(), ec);
+            total_bytes_consumed_ += consumed;
+            b_.consume(consumed);
+            if (ec == http::error::need_more)
+            {
+                ec.assign(0, ec.category());
+            }
```
vinniefalco: I would not use braces since it is a single statement.

include/boost/beast/http/impl/read.ipp
```diff
@@ -283,36 +296,23 @@ operator()(
...
-    switch(state_)
+    BOOST_ASIO_CORO_REENTER(*this)
     {
-    case 0:
-        if(Condition{}(p_))
+        if (Condition{}(p_))
         {
-            state_ = 1;
-            return s_.get_io_service().post(
+            BOOST_ASIO_CORO_YIELD s_.get_io_service().post(
```
vinniefalco: The problem with putting this all on one line is that it is harder to set breakpoints with visual studio, and it also moves the important part farther to the right in the editor. I try to keep the code narrow, it helps when I screenshare and do collaborative code reviews.

include/boost/beast/http/impl/read.ipp
```diff
@@ -414,32 +413,27 @@ operator()(
...
-        goto do_read;
     }
-upcall:
-    bytes_transferred = d.bytes_transferred;
-    d_.invoke(ec, bytes_transferred);
+    if (d.is_complete())
+    {
+        // Can't call the handler within the coroutine body
+        // because the data struct would be destroyed before
+        // we left the coroutine, causing use-after-free.
```
vinniefalco: Did you add this comment? How did you figure this out?

include/boost/beast/http/impl/read.ipp
```diff
@@ -414,32 +413,27 @@ operator()(
...
+            d.bytes_transferred += bytes_transferred;
+
         }
-        d.state = 2;
-        goto do_read;
     }
-upcall:
-    bytes_transferred = d.bytes_transferred;
-    d_.invoke(ec, bytes_transferred);
+    if (d.is_complete())
```
vinniefalco: This performs an unnecessary check on every continuation. Consider this instead:
```
    }
    return;
upcall:
    d_.invoke(ec, bytes_transferred);
}
```

And use `goto upcall` to invoke the final completion handler

vinniefalco:
I greatly appreciate your contribution, it is wonderful to see the community getting involved in the project. This is going to really help everyone since it will enable me to get even more done with the new libraries I am building on top of Beast. However, there's a learning curve to the style used to write the composed operations as you can see from my comments. Please don't let that discourage you!

I think that `prepare_or_error` should be in its own commit (in `boost::beast::detail::`) and come first, it looks quite the useful function and something that I will be adopting in new code.

include/boost/beast/http/impl/read.ipp
djarek: I'll check it out, though it's worth remembering the the most likely case is that there is no error. And if there is one, we've wasted 2 machine word writes, so I don't think this is a huge issue.

include/boost/beast/http/impl/read.ipp
djarek: Done.

include/boost/beast/http/impl/read.ipp
djarek: Yes I did, after I ran into a segfault in tests. I used asan to figure it out.

include/boost/beast/http/impl/read.ipp
djarek: Yeah, you're right, it doesn't belong in the public API. I'll move the is_continuation function to asio instead.

include/boost/beast/http/impl/read.ipp
vinniefalco: True

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +126,61 @@ operator()(
     error_code ec,
     std::size_t bytes_transferred)
 {
-    switch(state_)
+    size_t consumed = 0;
```
vinniefalco: You don't **have** to put this hear, you could put it in a braced scope where it is used:
```
{
    auto const consumed = p_.put(b_.data(), ec);
    total_bytes_consumed_ += consumed;
    b_.consume(consumed);
}
```

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +126,61 @@ operator()(
...
                 ec.assign(0, ec.category());
-                p_.put_eof(ec);
-                if(ec)
-                    goto upcall;
-                BOOST_ASSERT(p_.is_done());
-                goto upcall;
             }
-            ec = error::end_of_stream;
-            goto upcall;
+            BOOST_ASIO_CORO_YIELD
```
vinniefalco: I think you might need braces here, because `BOOST_ASIO_CORO_YIELD` is a compound statement:
```
if (ec == http::error::need_more)
{
    BOOST_ASIO_CORO_YIELD
    s_.get_io_service().post(
        bind_handler(std::move(*this), ec, 0));
}
```

This is why I indent it the way I do, I treat `BOOST_ASIO_CORO_YIELD` as if it was a separate statement:
https://github.com/boostorg/beast/blob/0bc2a414043f4a9430e71e6664d592e67ea6d282/include/boost/beast/websocket/impl/read.ipp#L162

include/boost/beast/http/impl/read.ipp
```diff
@@ -112,72 +126,61 @@ operator()(
...
-        state_ = 3;
-        return s_.get_io_service().post(
-            bind_handler(std::move(*this), ec, 0));
+            mb = prepare_or_error(b_, ec);
+            if (ec)
+            {
+                BOOST_ASIO_CORO_YIELD
+                    s_.get_io_service().post(
+                        bind_handler(std::move(*this), ec, 0));
+                break;
```
vinniefalco: Is the meaning of `break` defined for stackless coroutines?

djarek: ```BOOST_ASIO_CORO_YIELD break;``` terminates the coroutine.
```break;``` outside of a yield statement refers to the logical block you're in (if you're in a loop, it will break out of it).

include/boost/beast/http/impl/read.ipp
djarek: I've checked it and the difference is 1 assignment. One thing I'm wondering about is how should a composed operation behave if prepare() throws a different error (e.g. bad_alloc)?

vinniefalco: > how should a composed operation behave if prepare() throws a different error (e.g. bad_alloc)

I think it should propagate the exception as normal.

include/boost/beast/http/impl/read.ipp
vinniefalco: Should be `std::size_t`

PR #831, opened 2017-10-24
Version 126

vinniefalco:
* Add CppCon2017 presentation link
* Update README.md
* Update stream write documentation for end of stream changes
* Tidy up unused variable warnings
* Don't return end_of_stream on win32 file body writes
* Fix doc typo
* Fix shadowing in session_alloc
* Fix executor type compilation
* Add Travis tests with the default compilers
* Update Boost.WinAPI usage to the new location and namespace.
* Fix buffered_read_stream async_read_some signature

### DIFF
--- a/include/boost/beast/core/impl/buffered_read_stream.ipp
+++ b/include/boost/beast/core/impl/buffered_read_stream.ipp
@@ -144,11 +144,12 @@ buffered_read_stream(Args&&... args)
 
 template<class Stream, class DynamicBuffer>
 template<class ConstBufferSequence, class WriteHandler>
-auto
+BOOST_ASIO_INITFN_RESULT_TYPE(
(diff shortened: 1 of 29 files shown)


include/boost/beast/core/impl/buffered_read_stream.ipp
```diff
@@ -142,11 +144,11 @@ buffered_read_stream(Args&&... args)
 
 template<class Stream, class DynamicBuffer>
 template<class ConstBufferSequence, class WriteHandler>
-auto
+BOOST_ASIO_INITFN_RESULT_TYPE(WriteHandler, void(error_code))
```
djarek: The signature of an async_write_some handler of an AysncWriteStream should be void(error_code,size_t).

PR #832, opened 2017-10-24
prepare_or_error() helper function template

djarek:
Attempts to retrieve a mutable buffer sequence from a DynamicBuffer. If its maximal size has been exceeded it will set the error_code to buffer_overflow from the user-supplied enumeration type.
Will be used in #810.


vinniefalco:
Unfortunately this function exposes a problem with the way that beast's error codes are declared...

PR #838, opened 2017-10-26
Version 128

vinniefalco:
* Update doc links

HTTP:

* Add message::need_eof
* Use message::need_eof in example servers

API Changes:

* Remove serializer::keep_alive
* Remove serializer::chunked
* Add has_content_length_impl to Fields
* Add message::has_content_length
* Rename some basic_parser observers

Actions Required:

* Call message::keep_alive instead of serializer::keep_alive
* Call serializer::get::chunked instead of serializer::chunked
* Implement has_content_length_impl for user-defined Fields
* Remove the "is_" prefix from call sites invoking certain basic_parser members


doc/qbk/07_concepts/Fields.qbk
```diff
@@ -153,6 +161,9 @@ In this table:
             empty, the Transfer-Encoding field shall not appear when
             the associated __FieldsReader__ serializes the fields.
         ]]
+
+        If the result of adjusting the field value produces an empty
```
nbougalis: This reads weird. Prefer:

>If adjusting the field value results in an empty string, the field is removed from the container.

Or

>If the adjusted field value is empty, the field is removed from the container.

vinniefalco: Actually I wrote it that way on purpose. There is no requirement that implementations of **Fields** store strings at all. It should actually read:

*If the resulting adjusted field value would produce an empty string when serialized, the container behaves as if the field is removed*

Issue #850, opened 2017-10-28
CI report executable statistics for examples


vinniefalco:
The CI should report statistics for examples built with the release variant:

* Top data segment symbols, descending by size (unmangled names)
* Top text segment functions, descending by size
* Total size of executables
* Size reduction after stripping symbols

PR #851, opened 2017-10-28
Size checking of example programs

djarek:
Added a script which checks the example programs binary size (both with symbols and stripped) and outputs the statistics of symbols which contribute the most to the size of the stripped binary.

Resolves: #850


vinniefalco:
Maybe we should just leave out the link time optimization setting?

djarek:
Done.

vinniefalco:
How do I see the results?

PR #868, opened 2017-11-04
Version 137

vinniefalco:
* ConstBufferSequence mandates pointer equivalence
* Add FieldsWriter constructor requirement
* Tidy up some documented constructor syntax

### DIFF
--- a/doc/qbk/07_concepts/FieldsWriter.qbk
+++ b/doc/qbk/07_concepts/FieldsWriter.qbk
@@ -59,6 +59,16 @@ In this table:
         to end no earlier than after the `W` is destroyed.
     ]
 ][
+    [`W{f}`]
+    []
+    [
+        The implementation calls this constructor to indicate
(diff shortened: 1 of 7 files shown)


doc/qbk/07_concepts/FieldsWriter.qbk
```diff
@@ -59,6 +59,16 @@ In this table:
         to end no earlier than after the `W` is destroyed.
     ]
 ][
+    [`W{f}`]
```
djarek: Isn't this an optional requirement (only used if you use chunk-encoding)?

vinniefalco: Kind of. It has to be in the documentation or else user-defined FIelds can't be used as trailers. Obviously, this concept needs work. But there is no time to do it for this release. So for now, I just list this as a requirement. Although me and you both know it is not.

PR #872, opened 2017-11-06
CMake for GNU/Linux

ajneu:
Some changes to get it to run under Linux, with the following command

```bash
mkdir build
cd    build
cmake -DBOOST_ROOT=~/development/boost_own/boost_new_asio/ ..
make -j8
```

where I've compiled Boost myself and the following paths are on my system:
```bash
~/development/boost_own/boost_new_asio/include/boost/ # has e.g. asio/ etc.
~/development/boost_own/boost_new_asio/lib/           # has e.g. libboost_system.a libboost_system.so libboost_system.so.1.65.0
```

I was not able to check behaviour on Visual Studio, as I don't have windows.
Some changes might be necessary, to accomodate these changes and Visual Studio, but I cannot test it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 
 cmake_minimum_required (VERSION 3.5.1)
 
-project (Beast VERSION 144)
+project (Beast VERSION 145)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 option (Beast_BUILD_EXAMPLES "Build examples" ON)
--- a/include/boost/beast/version.hpp
+++ b/include/boost/beast/version.hpp
@@ -20,7 +20,7 @@
     This is a simple integer that is incremented by one every
     time a set of code changes is merged to the develop branch.
 */
-#define BOOST_BEAST_VERSION 144
+#define BOOST_BEAST_VERSION 145
 
 #define BOOST_BEAST_VERSION_STRING "Boost.Beast/" BOOST_STRINGIZE(BOOST_BEAST_VERSION)
 


vinniefalco:
>`find_package(Boost COMPONENTS system coroutine filesystem)`

`find_package` doesn't work when Beast is built "in-tree." The workflow for building tests and examples is to clone the Boost superproject, `cd libs/beast`, and run CMake from there. It will find the headers and libraries relative to that directory. This is the only way to sanely work on multiple versions of Boost at the same time.

Your solution will only work for the "installed" Boost which of course breaks in-tree builds. This is a general problem of combining Boost and CMake which we hope to resolve generally for all Boost libraries. But that could take a long time.

Until then, Beast's CMake support is limited to Windows, which supports "automatic linking."

vinniefalco:
Here is the workflow that your proposed change would break:
```
git clone git@github.com:boostorg/boost.git
cd boost
git submodule init
git submodule update
./bootstrap.sh
b2 variant=debug,release link=static runtime-link=static
cd libs/beast
mkdir bin
cd bin
cmake ..
```

After running `cmake` above, the resulting Makefile (or Visual Studio project file) will point to the system's "installed" version of boost (something like /usr/lib/boost/### and /usr/include/boost/###) instead of the version of boost clone in the tree above. This of course produces the wrong results.

ajneu:
Thanks for explaining. 

> Also, CMake is always behind the latest version of Boost when it comes to `find_package`. This is a source of endless frustration for both users and developers.

One could add a new version of boost... by trying this:
https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindBoost.cmake#L525

PR #875, opened 2017-11-07
Refactor basic_parser

djarek:
- Add missing noexcept specifiers and const qualifiers.
- Fix parser conversion constructor not copying header_limit_ and status_ fields.
- Use default special member functions when possible.
- Zero-initialize non-class members.

### DIFF
--- a/include/boost/beast/http/impl/basic_parser.ipp
+++ b/include/boost/beast/http/impl/basic_parser.ipp
@@ -26,20 +26,6 @@ namespace beast {
 namespace http {
 
 template<bool isRequest, class Derived>
-basic_parser<isRequest, Derived>::
-~basic_parser()
-{
-}
-
-template<bool isRequest, class Derived>
-basic_parser<isRequest, Derived>::
-basic_parser()
-    : body_limit_(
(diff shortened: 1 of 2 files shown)


vinniefalco:
The commit message is not "sterile" (i.e. it contains colorful language). Consider this instead:
```
Refactor basic_parser:

* Add missing noexcept specifiers and const qualifiers.
* Fix parser conversion constructor not copying header_limit_ and status_ fields.
* Use default special member functions when possible.
* Zero-initialize non-class members.
```

include/boost/beast/http/impl/basic_parser.ipp
```diff
@@ -49,6 +35,8 @@ basic_parser(basic_parser<
     , buf_(std::move(other.buf_))
     , buf_len_(other.buf_len_)
     , skip_(other.skip_)
+    , header_limit_(other.header_limit_)
```
vinniefalco: What about `body_limit_`?

vinniefalco:
Merged to **version 140**, minus the extraneous `noexcept`

PR #876, opened 2017-11-09
fix async_read_some() completion signature for yield_context

cbodley:
this adds a test case that calls async_read_some() with a yield_context. the test fails to compile on the return statement of async_read_some():
```
 libs/beast/test/beast/http/read.cpp:328:50:   required from here
./boost/beast/http/impl/read.ipp:544:28: error: void value not ignored as it ought to be
     return init.result.get();
```
full error message here: https://pastebin.com/raw/Xfqx8kKR

the async_completion in async_read_some() is just missing ReadHandler's size_t argument in its signature


vinniefalco:
Weird...I thought I fixed this already. In any event thanks a ton for the submission!

vinniefalco:
Merged to **version 140**

Issue #886, opened 2017-11-13
revisit mutable input area of static_buffer and basic_flat_buffer


vinniefalco:
These input areas are mutable to optimize the websocket implementation. Perhaps websocket can use a `buffers_adapter` over the dynamic buffer output area (return value of `prepare) instead, and keep the composed operation trivial. We can investigate a specialization of `buffers_adapter` for `boost::asio::mutable_buffer` for optimization.

vinniefalco:
Thinking about it further, websocket can't use `buffers_adapter`. It has to modify the input area, because the data is already committed (in the implementation of `read`). The safest way to resolve this is to give websocket a private implementation of `static_buffer` with a mutable input area, and revert the public dynamic buffer implementations to making their input areas constant buffer sequences.

vinniefalco:
This is done, `data` no longer returns a mutable buffer sequence. And `static_buffer` provides the `mutable_data` member which allows modification.

PR #888, opened 2017-11-13
Disallow mutation of const qualified flat_buffers

djarek:
Returning a mutable buffer sequence from a const qualified DynamicBuffer may result in unexpected results when a callee mutates it.

Resolves: #886


djarek:
Unfortunately, there's a larger design issue here, which requires a different solution.

PR #897, opened 2017-11-21
Fix inspection report issues and typo

zerotypos-found:


vinniefalco:
Part of #898

zerotypos-found:
FYI, you can see the inspection report by
1. Go to http://www.boost.org/
2. Click "Development" at the right menu bar
3. Click "Testing" at the right menu bar
4. Then, there is "Inspection Reports".

However, [the report](http://boost.cowic.de/rc/docs-inspect-develop.html) is not updated frequently and so this report does not include the inspection results of Boost.Beast yet.

You can also **make the report by yourself** by
1. Run `b2` at `BOOST_ROOT/tools/inspect`.
2. Then, the `inspect` tool is built (the tool is located somewhere).
3. Run `inspect` and pass it the path to boost root's directory as a command-line argument (e.g. `inspect "$BOOST_ROOT" > inspect.html`).

(This PR fixed only a fraction of inspection report issues.)

PR #903, opened 2017-11-23
Fix UB in buffers_cat_view

djarek:
- `std::array<char, N>` is not a valid type for in-place polymorphic storage, due to the fact that it might not have the required alignment.
- Removed unnecessary casts to `void*` in `iter()`.


vinniefalco:
`std::aligned_union` is not available in some versions of the stdlib even though they claim to be C++11, try `std::aligned_storage`. Otherwise, I like where you are going with this!

vinniefalco:
Merged some of it

PR #910, opened 2017-11-26
Use Boost.Asio string_view if available.

Jiwan:
This PR tries to align **Boost.Beast**'s string_view with **Boost.Asio**. Since ``std::string_view`` and ``std::experimental::string_view`` do not have a ``clear`` member function, ``a_string_view = string_view();`` is used instead ([see n4288](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4288.html)).


Jiwan:
Note that I am still unsure if using the ``detail`` folder of **Boost.Asio** is wise.

include/boost/beast/core/string.hpp
```diff
@@ -10,21 +10,45 @@
 #ifndef BOOST_BEAST_STRING_HPP
 #define BOOST_BEAST_STRING_HPP
 
+#include <boost/asio/detail/string_view.hpp>
```
vinniefalco: Hmm... I don't know about this! Including a detail header is not best practices

include/boost/beast/core/string.hpp
```diff
@@ -10,21 +10,45 @@
...
+
+#ifndef BOOST_ASIO_NO_STRING_VIEW
+# if defined(BOOST_ASIO_HAS_STD_STRING_VIEW)
+#  define BOOST_ASIO_HAS_STRING_VIEW 0
+# else
+#  define BOOST_ASIO_HAS_STRING_VIEW 1
```
vinniefalco: I don't think we are allowed to adjust these macros. The user needs to adjust them not our library.

include/boost/beast/core/string.hpp
```diff
@@ -10,21 +10,45 @@
...
+# endif
+#endif
+
 #include <boost/beast/core/detail/config.hpp>
 #include <boost/version.hpp>
+
+#if BOOST_ASIO_NO_STRING_VIEW
 #include <boost/utility/string_view.hpp>
+#endif
```
vinniefalco: I think we need @chriskohlhoff to weigh in on this, to decide if he wants to support other libraries piggy-backing on the string view declaration.

include/boost/beast/core/string.hpp
Jiwan: I was simplified things a bit more.

include/boost/beast/http/detail/rfc7230.hpp
```diff
@@ -352,8 +352,8 @@ increment()
         {
             it = first;
         };
-    v.first.clear();
-    v.second.clear();
+    v.first = string_view();
+    v.second = string_view();
```
vinniefalco: Consider this instead:
```
v.first = {};
v.second = {};
```

vinniefalco: Actually, I'll make this change myself in its own commit.

Jiwan: :+1:

omartijn:
It'd be really great if this issue could be revisited. The fact that asio and beast use different string_view implementations feels sub-optimal. It's mostly small things, like not being able to directly use `boost::beast::http::message::operator[](boost::beast::http::field::host)` as parameter to `boost::asio::ip::tcp::resolver::async_resolve`.

It's trivial to work around, but it doesn't make the code more readable. In addition, I quite like the idea of using `std::string_view` (or `std::experimental::string_view`) when available.

vinniefalco:
Yes I believe we can do something. In "stand-alone" (not implemented yet) mode Beast can use standalone Asio, require C++17, and use `std::string_view`.

Otherwise, Beast can require C++11 and use Boost.Asio. In this configuration, Beast will use `boost::string_view`, unless the Asio macro `BOOST_ASIO_HAS_STD_STRING_VIEW` is defined, in which case Beast will use `std::string_view` and require C++17.

PR #919, opened 2017-12-02
Support for noncopyable completion handlers (HTTP Read/Write)

djarek:
HTTP async_read_* and async_write_* will now accept CompletionTokens which produce move-only CompletionHandlers.

### DIFF
--- a/test/extras/include/boost/beast/test/stream.hpp
+++ b/test/extras/include/boost/beast/test/stream.hpp
@@ -488,7 +488,7 @@ async_read_some(
             return boost::asio::post(
                 in_->ioc.get_executor(),
                 bind_handler(
-                    init.completion_handler,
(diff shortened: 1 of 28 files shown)


test/extras/include/boost/beast/test/stream.hpp
```diff
@@ -702,18 +702,11 @@ class stream::read_op_impl : public stream::read_op
         lambda(lambda&&) = default;
         lambda(lambda const&) = default;
 
-        lambda(state& s, Buffers const& b, Handler&& h)
+        template <typename DeducedHandler>
```
vinniefalco: Style: `template<class DeducedHandler>`

vinniefalco:
This is part of #920 now

PR #922, opened 2017-12-04
Move only handler ptr

djarek:
It is no longer a reference counted object and now has semantics close to a std::unique_ptr.


vinniefalco:
Part of #920

PR #925, opened 2017-12-06
Work around OOM in travis

djarek:
Split the websocket read test to reduce compile time memory usage  by a small amount, which ought to be enough to let it compile in a constrained environment like travis.

### DIFF
--- a/test/beast/websocket/read2.cpp
+++ b/test/beast/websocket/read2.cpp
@@ -1,0 +1,254 @@
+//
+// Copyright (w) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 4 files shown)


vinniefalco:
http://www.doctorwho.tv/brand/assets/old-brand/img/profile-heros/ood.png

vinniefalco:
OH....sorry, you said OOM.

test/beast/websocket/read2.cpp
```diff
@@ -0,0 +1,254 @@
...
+//
+
+// Test that header file is self-contained.
+#include <boost/beast/websocket/stream.hpp>
+
+#include "test.hpp"
+
+#include <boost/asio/write.hpp>
+
+#include <boost/beast/core/buffers_to_string.hpp>
```
vinniefalco: This include is not needed

djarek:
In #928

PR #931, opened 2017-12-09
Remove first_ and last_ from win32 file_body

djarek:
These members were never assigned values other than `0` and `size_` respectively.


vinniefalco:
You have it exactly backwards. Rather than removing `first_` and `last_`, we should be implementing the **range** feature, where `file_body` can be used to deliver a subset of the file.

vinniefalco:
See https://github.com/boostorg/beast/issues/723

PR #935, opened 2017-12-10
Enable body composition

### LINKED ISSUE #884: Implementation of sum/product Body types is impossible
It is impossible to compose Bodies into sum types, e.g.:
```cpp
template <typename... Ts>
struct variant_body
{
    using value_type = std::variant<typename Ts::value_type...>;
    struct reader;
    struct writer;
};

using my_body = variant_body<beast::http::empty_body, json_body, beast::http::string_body>;
```
due to the requirement of Body::reader and Body::writer that the message's body type is exactly the same type as Body. The same restriction applies to product types(not as useful as a variant_body). Not sure how to make this use-case possible without complicating other use-cases.

djarek:
This change enables composing bodies into multiple layers e.g. compressed_body<json_body<T>>. Resolves #884.

Actions required:
- change body writer constructor signature to
	writer(header<isRequest, Fields> const&, value_type const&)
- change body reader constructor signature to
	reader(header<isRequest, Fields>&, value_type&)

### DIFF
--- a/include/boost/beast/http/serializer.hpp
+++ b/include/boost/beast/http/serializer.hpp
@@ -62,7 +62,7 @@ public:
 
     static_assert(is_body_writer<Body>::value,
         "BodyWriter requirements not met");
-
+    using body_value_type = typename Body::value_type;
     /** The type of message this serializer uses
 
         This may be const or non-const depending on the
@@ -73,10 +73,16 @@ public:
 #else
     using value_type =
         typename std::conditional<
-            std::is_constructible<typename Body::writer,
+            (std::is_constructible<typename Body::writer,
+                header<isRequest, Fields>&, body_value_type&>::value &&
+            ! std::is_constructible<typename Body::writer,
+                header<isRequest, Fields> const&,
+                    body_value_type const&>::value) ||
+            /* Deprecated BodyWriter Concept (v1.66) */
+            (std::is_constructible<typename Body::writer,
                 message<isRequest, Body, Fields>&>::value &&
             ! std::is_constructible<typename Body::writer,
-                message<isRequest, Body, Fields> const&>::value,
+                message<isRequest, Body, Fields> const&>::value),
             message<isRequest, Body, Fields>,
             message<isRequest, Body, Fields> const>::type;
 #endif
@@ -188,6 +194,8 @@ private:
     bool header_done_ = false;
     bool more_;
 
+    serializer(value_type& msg, std::true_type);
+    serializer(value_type& msg, std::false_type);
 public:
     /// Constructor
     serializer(serializer&&) = default;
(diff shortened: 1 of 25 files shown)


test/beast/http/serializer.cpp
```diff
@@ -115,10 +149,20 @@ class serializer_test : public beast::unit_test::suite
         }
     }
 
+    void testBodyWriterCtor()
+    {
+        response<deprecated_body> resp{};
+        request<deprecated_body> req{};
```
vinniefalco: Why are you value-initializing?

djarek: Probably a copy-paste leftover. Will fix.

vinniefalco:
Please rebase against my **v150** and make the following changes:
If `BOOST_BEAST_ALLOW_DEPRECATED` is non zero (e.g. `#if BOOST_BEAST_ALLOW_DEPRECATED`) then allow the deprecated constructors. Otherwise, cause a `static_assert` if those constructors are instantiated. You will have to do some tricks to make the assert dependent on template parameter. For example `static_assert(sizeof(Body)==0, "message")`. Have a look at how I did it in the **v150** branch and ask questions if needed.

vinniefalco:
The commit message should indicate the change (new constructor signatures), not the end result (enable body composition). For example:

```
New BodyReader and BodyWriter constructors:

fix #884

* BodyReader and BodyWriter constructors now require the header and
  body elements to be passed as distinct header and value_type objects.

This enables the composition of body types, for example:

    http::response<compressed_body<http::string_body>> res;

* The previous single-argument constructors are deprecated and will be
  removed in a subsequent version.

Actions Required:

The old concepts signatures are supported if BOOST_BEAST_ALLOW_DEPRECATED
is defined as 1.

Actions required:

* Change user-defined instances of BodyReader or BodyWriter constructor
  signatures to the two-argument form, OR
* Define the macro BOOST_BEAST_ALLOW_DEPRECATED in the project
```

include/boost/beast/http/serializer.hpp
```diff
@@ -62,7 +62,7 @@ class serializer
 
     static_assert(is_body_writer<Body>::value,
         "BodyWriter requirements not met");
-
+    using body_value_type = typename Body::value_type;
```
vinniefalco: tsk tsk... you can't just casually add a public nested type! This affects the public API. As library developers we have to prefer to be inconvenienced over adding unnecessary interfaces.

vinniefalco:
I've merged this into **v150**: https://github.com/vinniefalco/beast/commits/v150

All of my edits are in one commit: https://github.com/vinniefalco/beast/commit/9842af966dc9843e794e159cd580f06672ecd50a

The commits marked "[FOLD]" will be folded into the previous commit

PR #956, opened 2017-12-30
Build against installed Boost with CMake

TBBle:
These patches allow the CMake build script to build against a compiled version of Boost 1.66.0 or later.

This requires passing in an extra flag to CMake, so should not affect existing users in any way, but makes it convenient for working on beast.

Tested against the Boost.org 1.66.0 Windows installers for Visual Studio 2017 and Visual Studio 2015, and the Boost 1.66.0 version packaged for mingw-w64 in [msys2](http://www.msys2.org/) with gcc 7.2.0.

I see no reason it wouldn't also work on Linux, but there's no Linux packages for Boost 1.66.0 yet, as far as I can tell.

Note that CMake doesn't yet support Boost 1.66.0 from the upstream packages on Windows, pending [CMake Issue 17575](https://gitlab.kitware.com/cmake/cmake/issues/17575). An updated `FindBoost.cmake` is available in [CMake Merge Request 1625](https://gitlab.kitware.com/cmake/cmake/merge_requests/1625).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,11 @@ else()
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wrange-loop-analysis")
     endif ()
+    if (WIN32)
+        # -mbig-obj is the gas equivalent of /bigobj for MSVC
+        # -O2 to work around string table size issues: https://stackoverflow.com/a/29479701/166389
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj -O2")
+    endif ()
 endif()
 
 # Must come before Boost includes, otherwise the
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ else()
     find_package (Threads)
 
     set( CMAKE_CXX_FLAGS
-      "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Wpedantic -Wno-unused-parameter")
+      "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-implicit-fallthrough")
 
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wrange-loop-analysis")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,8 +74,6 @@ include_directories (include)
 # Boost
 #
 
-get_filename_component (BOOST_ROOT ../../ ABSOLUTE)
-
 # VFALCO I want static but "b2 stage" builds a minimal set which excludes static
 add_definitions (-DBOOST_ALL_STATIC_LINK=1)
 
@@ -86,9 +84,23 @@ add_definitions (-DBOOST_ASIO_DISABLE_BOOST_DATE_TIME=1)
 add_definitions (-DBOOST_ASIO_DISABLE_BOOST_REGEX=1)
 add_definitions (-DBOOST_COROUTINES_NO_DEPRECATION_WARNING=1)
 
-include_directories (${BOOST_ROOT})
+OPTION(FIND_BOOST "Build beast against an existing Boost install, rather than in-tree")
+if (FIND_BOOST)
+  set(Boost_USE_STATIC_LIBS ON)
+  find_package(Boost 1.66.0 REQUIRED COMPONENTS system filesystem coroutine)
+  include_directories(${Boost_INCLUDE_DIRS})
+  link_directories(${Boost_LIBRARY_DIRS})
+  link_libraries(${Boost_COROUTINE_LIBRARY})
+  link_libraries(${Boost_CONTEXT_LIBRARY}) # Needed by coroutine
+  link_libraries(${Boost_FILESYSTEM_LIBRARY})
+  link_libraries(${Boost_SYSTEM_LIBRARY})
+else()
+  get_filename_component (BOOST_ROOT ../../ ABSOLUTE)
 
-link_directories(${BOOST_ROOT}/stage/lib)
+  include_directories (${BOOST_ROOT})
+
+  link_directories(${BOOST_ROOT}/stage/lib)
+endif()
 
 if (MINGW)
     link_libraries(ws2_32 mswsock)


CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ if (MSVC)
         #/wd4503      # decorated name length exceeded, name was truncated
         /W4           # enable all warnings
         /MP           # Multi-processor compilation
+        /source-charset:ISO-8859-1 # base64.hpp is in ISO-8859-1 encoding
```
vinniefalco: We should just fix the source file to not have high-ASCII chars

TBBle: Probably true. In this case, the non-ASCII character is (twice) in the copyright notice for the source pulled from http://www.adp-gmbh.ch/cpp/common/base64.html, and is that way in the original too, down to the encoding. It only complains if your local code page is non-European (cp932 in my case).

vinniefalco:
The problem is that this is not part of my workflow or the CI workflow, and CMake support for Boost is not the greatest. FindBoost.cmake is always out of date. Beast does not really support CMake builds except for the one case of generating a Visual Studio project file for in-tree builds, this is explained in the README.md:
https://github.com/boostorg/beast#building

TBBle:
Fair 'nuff. Since I've been working on improving Boost support in CMake recently, and have context, what sort of things do you feel need improving there? It's already pretty future-proof (you can just tell it to accept a given version even if it doesn't recognise it, and it'll assume the dependencies are the same as the most-recent version it knows), although the upstream Boost binaries for Windows changed filenames in 1.66.0 so that's been a minor hassle to resolve.

Historically, it has been the non-Windows builds that have shown wild variation in filenames between Boost releases. I spent a few hours trying to make sense of [CMake's and Boost's file-naming history](https://gitlab.kitware.com/cmake/cmake/issues/17575#note_358583) yesterday with an aim to rationalising it.

My personal use-case for this work is as a basis/known-good point for experimenting with beast on Standalone ASIO or the Networking TS reference implementation (I understand the latter is a subset of the former, and haven't looked yet to see if it's sufficient for beast's needs), partly to get practice with the Networking TS, and partly because I'm interested in using beast with an existing project where Boost isn't really practical to integrate at this time.

Once the CMake support for Boost 1.66.0 on Windows hits AppVeyor, or Boost 1.66.0 itself starts to hit Linux packages, then I can (and probably will) get CI builds going for this branch. If there were available now, I'd have included them in the pull request. I could do AppVeyor right now, by including the updated `FindBoost.cmake` in the repo, but that seems ugly for a pull request.

vinniefalco:
@djarek or @glenfe do you have any way to evaluate this?

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ else()
     find_package (Threads)
 
     set( CMAKE_CXX_FLAGS
-      "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Wpedantic -Wno-unused-parameter")
+      "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-implicit-fallthrough")
```
vinniefalco: I prefer not to turn off the warning, I believe the remaining warnings are fixed in 154.

CMakeLists.txt
```diff
@@ -58,6 +58,11 @@ else()
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wrange-loop-analysis")
     endif ()
+    if (WIN32)
+        # -mbig-obj is the gas equivalent of /bigobj for MSVC
+        # -O2 to work around string table size issues: https://stackoverflow.com/a/29479701/166389
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj -O2")
+    endif ()
```
vinniefalco: This change looks benign, and could be accepted

CMakeLists.txt
```diff
@@ -81,9 +84,23 @@ add_definitions (-DBOOST_ASIO_DISABLE_BOOST_DATE_TIME=1)
 add_definitions (-DBOOST_ASIO_DISABLE_BOOST_REGEX=1)
 add_definitions (-DBOOST_COROUTINES_NO_DEPRECATION_WARNING=1)
 
-include_directories (${BOOST_ROOT})
+OPTION(FIND_BOOST "Build beast against an existing Boost install, rather than in-tree")
+if (FIND_BOOST)
+  set(Boost_USE_STATIC_LIBS ON)
```
djarek: Why are you forcing the use of static libs?

TBBle: If I remember correctly, it was because linking against the dynamic libraries on Windows would then require the Boost DLLs for that version to be in your system path, or be copied into the library directory. The former doesn't appear to be the default, and the latter would normally be handled by `install` rules, and this CMakeLists.txt is not set up for the install stage, and I don't see there's any expectation that this will change.

Basically, this gives the best out-of-the-box experience in common scenarios, where you can immediately run unit-tests (i.e. through CTest, which is a different patch series I haven't put up for merging) after the build. On Windows it'll link to static libs, and (again, if I remember correctly) on Linux you generally don't *have* static libs so it'll link to the dynamic libraries anyway.

mloskot:
Building and testing against a released Boost could be considered as an extension to this [official Boost recommendation](https://svn.boost.org/trac10/wiki/StartModMaint#Gettingreadytoworkonalibrary):

>  The preferred environment library maintenance is to checkout the library's develop branch, or some other development branch, while other Boost libraries are as defined by the Boost super-project master branch. This causes local tests of your library to run against master for other Boost libraries at the point in time referenced by the Boost super-project. 

It could also simplify the procedure by allowing to clone just Beast, not the whole super project.

TBBle:
> It could also simplify the procedure by allowing to clone just Beast, not the whole super project.

That's *precisely* what I wanted to do, that started me down this path.

vinniefalco:
>It could also simplify the procedure by allowing to clone just Beast, not the whole super project.

Is this not possible using bjam?

mloskot:
@TBBle I have adopted your concept for Boost.GIL. Thanks!
@vinniefalco No idea, but developing a Boost library using IDE-friendly build setup than BBv2 simply is a winner

TBBle:
I'd characterise it as "Not of particular interest to the maintainer", and my original motivation (Beast using Standalone ASIO in a non-Boost environment) is no longer applicable.

CMakeLists.txt
pfultz2: The `CMAKE_CXX_FLAGS` should be set by the user(either from the command-line or in the toolchain file).  You can add flags for warnings with `add_compile_options`.

Also, you shouldn't set the C++ version. The user should decide that. You could set a default with `CMAKE_CXX_STANDARD`.

CMakeLists.txt
```diff
@@ -81,9 +84,23 @@ add_definitions (-DBOOST_ASIO_DISABLE_BOOST_DATE_TIME=1)
...
+  find_package(Boost 1.66.0 REQUIRED COMPONENTS system filesystem coroutine)
+  include_directories(${Boost_INCLUDE_DIRS})
```
pfultz2: Usage requirements should be applied to a target.

CMakeLists.txt
```diff
@@ -81,9 +84,23 @@ add_definitions (-DBOOST_ASIO_DISABLE_BOOST_DATE_TIME=1)
 add_definitions (-DBOOST_ASIO_DISABLE_BOOST_REGEX=1)
 add_definitions (-DBOOST_COROUTINES_NO_DEPRECATION_WARNING=1)
 
-include_directories (${BOOST_ROOT})
+OPTION(FIND_BOOST "Build beast against an existing Boost install, rather than in-tree")
```
pfultz2: Ugly but necessary workaround until boost officially supports cmake. In Boost.CMake, the projects just unashamedly use `find_package`, but the superproject disables the in-tree versions with [`bcm_ignore_package`](http://bcm.readthedocs.io/en/latest/src/BCMIgnorePackage.html#bcm-ignore-package).

CMakeLists.txt
vinniefalco: But this CMakeLists.txt is for building the tests and examples, not for use for building a user's project that uses beast.

pfultz2: But you want to be able build the tests and examples with a specific C++ version. There is semantic differences between the C++11 and C++17 that it makes sense to want to build the  tests and examples with C++17 instead of C++11.

vinniefalco: I want to make sure that the tests and examples compile in C++11, so I don't accidentally use a C++14 or C++17 feature when I am maintaining beast.

vinniefalco:
> I'd characterise it as "Not of particular interest to the maintainer",

I am interested but this is a seriously complex problem.

CMakeLists.txt
pfultz2: Yea but you have no way of testing for C++14 or C++17, either. Should users assume these are not supported since they are never tested for? 

Testing with C++11 will ensure everything will pass with C++11, but I dont see why that would negate testing with newer version as well.

CMakeLists.txt
TBBle: That was why I kept it as a separate commit, since it would be needed even when using CMake in the way you intended/documented (albeit with gcc or clang under Win32, which is not a super-common workflow.)

CMakeLists.txt
TBBle: I feel a lot of your comments here are directed at the code as it already stood, rather than this pull request, and probably should be handled in a separate (possibly preceding) change.

For this series, I didn't want to rewrite the CMake build system, simply extend what was already there to accommodate a different workflow. @vinniefalco has expressed reluctance to take on the workload of a full CMake-based build system, given the minimal benefits it provides to him as a maintainer, and almost-zero benefits it provides to users of this library (who would get it via Boost and hence never interact with this build-system at all).

Similarly, I wasn't particularly interested in trying to full CMake-ify this code, for the same reasons. If and when BCM or similar lands in Boost, then it's worth revisiting this, as the potential maintainer load would be much lower, and the user-visible benefits much higher.

PR #957, opened 2017-12-30
Version 151

vinniefalco:
* Sanitizer failures are errors
* Depend on container_hash
* Fix high-ASCII in source file

WebSocket:

* Control callback is invoked on the execution context
* Add stream_fwd.hpp
* Remove unnecessary include

API Changes:

* http::parser is not MoveConstructible
* permessage-deflate is a compile-time feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 
 cmake_minimum_required (VERSION 3.5.1)
 
-project (Beast VERSION 150)
+project (Beast VERSION 151)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 option (Beast_BUILD_EXAMPLES "Build examples" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,8 +150,8 @@ elseif ("${VARIANT}" STREQUAL "ubasan")
     else()
         set (CMAKE_BUILD_TYPE RELWITHDEBINFO)
         set (CMAKE_CXX_FLAGS
-          "${CMAKE_CXX_FLAGS} -DBOOST_BEAST_NO_SLOW_TESTS=1 -msse4.2 -funsigned-char -fno-omit-frame-pointer -fsanitize=address,undefined -fsanitize-blacklist=${PROJECT_SOURCE_DIR}/tools/blacklist.supp")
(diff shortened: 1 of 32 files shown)


CMakeLists.txt
```diff
@@ -150,8 +150,8 @@ elseif ("${VARIANT}" STREQUAL "ubasan")
     else()
         set (CMAKE_BUILD_TYPE RELWITHDEBINFO)
         set (CMAKE_CXX_FLAGS
-          "${CMAKE_CXX_FLAGS} -DBOOST_BEAST_NO_SLOW_TESTS=1 -msse4.2 -funsigned-char -fno-omit-frame-pointer -fsanitize=address,undefined -fsanitize-blacklist=${PROJECT_SOURCE_DIR}/tools/blacklist.supp")
-        set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address,undefined")
+          "${CMAKE_CXX_FLAGS} -DBOOST_BEAST_NO_SLOW_TESTS=1 -msse4.2 -funsigned-char -fno-omit-frame-pointer -fsanitize=address,undefined -fno-sanitize-recover=address,undefined -fsanitize-blacklist=${PROJECT_SOURCE_DIR}/tools/blacklist.supp")
+        set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address,undefined -fno-sanitize-recover=address,undefined")
```
glenfe: Are there any drawbacks to having just `-f-no-sanitize-recover=all` ?

include/boost/beast/http/parser.hpp
```diff
@@ -84,21 +84,17 @@ class parser
...
+    /// Constructor (disallowed)
     parser(parser const&) = delete;
 
-    /// Assignment
+    /// Assignment (disallowed)
     parser& operator=(parser const&) = delete;
 
-    /** Constructor
+    /// Constructor (disallowed)
+    parser(parser&& other) = delete;
```
glenfe: No need for `parser(parser&& other) = delete;` since you explicitly delete the copy constructor. e.g. You're not deleting the move-assignment operator since you're letting the explicit deletion of the copy-assignment operator imply that.

vinniefalco: This is so it shows up in the generated documentation

CMakeLists.txt
vinniefalco: I have no idea I just copy pasted that

PR #969, opened 2018-01-05
Version 152

vinniefalco:
* Refactor detect_ssl_op

WebSocket:

* Redistribute the read tests in the translation units
* Refactor error headers
* Add WebSocket error conditions

API Changes:

* Refactor WebSocket errors (API Change):

Actions Required:

* Code which explicitly compares error_code values against the
  constant `websocket::error::handshake_failed` should compare
  against `websocket::condition::handshake_failed` instead.

* Code which explicitly compares error_code values against the
  constant `websocket::error::failed` should compare
  against `websocket::condition::protocol_violation` instead.


include/boost/beast/websocket/impl/error.ipp
```diff
@@ -11,71 +11,112 @@
...
+error_codes::
+default_error_condition(error e) const noexcept
+{
+    switch(e)
     {
-        return condition.value() == ev &&
-            &condition.category() == this;
-    }
+    default:
+    case error::closed:
```
djarek: The fallthroughs here might generate a lot of warnings with `-Wimplicit-fallthrough` enabled.

TBBle: `-Wimplicit-fallthrough` already generates a lot of noise for the existing codebase. I [proposed](https://github.com/boostorg/beast/pull/956/commits/1a3c8d2856ddfdca14c544a7bf860788be0489e2) adding `-Wno-implicit-fallthrough` to the CMake file, and I assumed it was already set in the Jamfile, for the examples. I'd suggest addressing that separately and globally.

include/boost/beast/websocket/error.hpp
```diff
@@ -20,22 +21,230 @@ namespace websocket {
...
+
+        Error codes with this value will compare equal to @ref condition::protocol_violation
+    */
+    bad_close_code,
+
+    /** The WebSocket close frame payload size was invalid
+
+        Error codes with this value will compare equal to @ref condition::protocol_violation
+    */
+    bad_close_size,
```
TBBle: Looking at this and the next, they appear to match some close_code values? My initial reaction is that `websocket::close_reason::code` should be returning a close_code enum, with its own category specifically for close_code. Then the error_condition could still compare equal for `condition::protocol_violation` on those codes, or any other close_codes that mean that. Users could also then tell if they have a valid close_reason from looking at the category of the error_code they received.

include/boost/beast/websocket/error.hpp
```diff
@@ -20,22 +21,230 @@ namespace websocket {
...
+*/
+#if ! BOOST_BEAST_DOXYGEN
+    unused1 = 2, // failed
+    unused2 = 3, // handshake_failed
+#endif
 
-    /// buffer overflow
+    /** The WebSocket operation caused a dynamic buffer overflow
+    */
     buffer_overflow,
```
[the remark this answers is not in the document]
TBBle: Not in the error_code, they're separate categories, so duplication is likely and expected. It might make sense to have a single condition that can match both, if there's a use-case where you'd check that. You could also provide equivalence to `boost::system::errc::no_buffer_space` for example, if you don't want your own error_condition value.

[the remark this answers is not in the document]
djarek: I think this came up some time ago when we were considering using internally `prepare_or_error`, however, the problem is that you can't have such a generic function for both websockets and http because of the duplication of that error code.

include/boost/beast/websocket/impl/error.ipp
vinniefalco: That sounds to me like a defect in gcc or clang, these should not be producing warnings. Are you using a release version of the compiler? What version are you on where you see these warnings?

include/boost/beast/websocket/error.hpp
vinniefalco: Close codes only represent a subset of the possible errors that can take place. And sometimes the value is not set correctly by the remote host. So the error code provides more concise information. I don't think it is a good idea to try to hook the close code system into a category. Especially because there can be custom close codes (for example in the 3,000 to 3,999 range) which are application specific and may have different meanings.

include/boost/beast/websocket/error.hpp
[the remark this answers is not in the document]
vinniefalco: A common condition seems like a good idea.

include/boost/beast/websocket/impl/error.ipp
TBBle: gcc-7.2.0 mingw-w64 build in msys. According to [the implementer of the warning](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/), it's been implemented only now that we have `[[fallthrough]]` to annotate deliberate switch fallthrough, and it's part of `-Wextra` in gcc 7. It might be considered a defect that it's active with `-std=c++11`, I guess, but since gcc has attributes to indicate deliberate fallthrough on all C++ levels, I can see the rationale here.

Assuming all target compilers correctly ignore unknown `[[]]` attributes (need to check this...), then the choice is either document that you need `-Wno-implicit-fallthrough` with gcc 7 or later, get gcc policy changed, or add `[[gnu::fallthrough]]` or some recognised comment to the falling-through lines.

include/boost/beast/websocket/error.hpp
TBBle: The "subset of possible errors" is why you'd have a category. The category decides how to deal with the interactions of that subset with the real world. Application specific close codes could easily work by deriving from a `ws_close_code_category`, and handling the 3XXX codes, and then upcalling for anything else.

The idea of the category is that if you have different meanings for error numbers (i.e. close_code 7 != `boost::beast::websocket::error` 7, then having different categories lets the category author define the interoperation.

That said, close codes (e.g. none) might fall into that somewhat grey area where "non-zero success and/or zero failure" is not clearly supported or disallowed, and there's an ongoing need to clarify this.

I also see that these two error codes are not set based on the peer-provided close code, but determined locally, so my idea of replacing these is not right. The general idea that close_code fits in the error_code model however, I do think is valid, modulo the above paragraph.

It's also possible this demonstrates my under-experience with Beast itself.

vinniefalco: I think it is more subjective. For example there is a group who thinks that the error code system should be used to reflect HTTP status codes (e.g. 200 OK). I am not in that group.

include/boost/beast/websocket/impl/error.ipp
vinniefalco: I've already dealt with the fallthrough warnings throughout Beast by using the `BOOST_FALLTHROUGH` macro. But I don't think the following code should cause a warning:
```
switch(n)
{
case 0:
case 1:
    std::cout << "low\n";
    break;
default:
    std::cout << "high\n";
break;
}
```

Or should it?

TBBle: Why wouldn't it? It's implicitly falling through from `case 0` to `case 1`. Do you intend that because `case  0` labels a null statement, the warning should treat it as labelling the same statement as `case 1`, and not complain? It's not a bad case to make, although the counter-position that comes to mind is that if that null statement was only because a macro had disappeared other statements, it's still carrying the same risk for the reader as the more-obvious implicit fallthrough.

include/boost/beast/websocket/error.hpp
TBBle: Indeed, I'm one of those folks. That was my point about "non-zero success", as that turns out to be one of the technical sticking points to letting the implementors decide that; specifically, the semantic of `error_code::operator bool` is not clear, and usage guidance in blog-posts conflicts over it (as well as whether HTTP status codes make sense in `error_code`).

Lack of message/string/data support in `error_code` is also pretty clear, and various approaches show up in projects like Boost.outcome (gives you a pointer into a ring-buffer) or our local `error_code` usage (locally subclassed `std::error_code` with a string added, and we just accept that slicing to `std::error_code` will lose that detail), to carry those around. It seems that as far as the standardisation effort was concerned, that's what the `what` in `system_error` was for, and if you weren't going to put your `error_code` into `system_error` and throw it, carrying a specific string was just left off-the-table. FIlesystem TS similarly gives richer results when throwing `path_error` than when simply setting an `error_code`, so this is not a once-off choice.

Anyway, this isn't really a discussion for here. `error_code` practice is still evolving, I know of at least two people who've indicated plans to write papers addressing issues like this, and I also have a (smaller) one planned, but far down my backlog.

include/boost/beast/websocket/impl/error.ipp
glenfe: This particular case should not generate any warnings:
```cpp
switch (n) {
case 0:
case 1:
    /* code */
}
```
I wouldn't use BOOST_FALLTHROUGH for these cases.

TBBle: I haven't checked if this construct is actually the cause of the surprise warnings in gcc 7, but if it is, then maybe worth raising with gcc? A quick look at the discussion in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=7652 suggests it's not supposed to warn in this case ("empty body"), which is [apparently consistent with clang's behaviour](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=7652#c13)

PR #970, opened 2018-01-08
File error code consistency

TBBle:
While playing with `std::error_code`, I found a couple of places using a `boost::system::error_code` API that doesn't work with `std::error_code`, and replaced them with a cross-compatible approach.

I also replaced in-passing a couple of places that were directly calling `::GetLastError()` instead of `boost::winapi::GetLastError()`. I was hoping to get rid of the `static_cast<int>` in a few places, but that's built-in to the way `error_code` has been defined, sadly.

### DIFF
--- a/include/boost/beast/core/impl/file_posix.ipp
+++ b/include/boost/beast/core/impl/file_posix.ipp
@@ -224,7 +224,7 @@ size(error_code& ec) const
 {
     if(fd_ == -1)
     {
-        ec.assign(errc::invalid_argument, generic_category());
+        ec = make_error_code(errc::invalid_argument);
         return 0;
     }
     struct stat st;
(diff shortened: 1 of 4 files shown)


vinniefalco:
Seems pretty reasonable, thanks!

vinniefalco:
@TBBle Since you're on the subject of error codes, would you mind taking a peek at this pull request? It could use a code review. Error codes and error conditions are pretty arcane, so an extra set of eyes on it might help me avoid doing something silly! Thanks
https://github.com/boostorg/beast/pull/969

TBBle:
@vinniefalco Sure, happy to have a look.

vinniefalco:
This will go into 153

vinniefalco:
Merged into 153

PR #994, opened 2018-01-22
Fix iterator version of basic_fields::erase

inetic:
When used, was causing compile errors.

Not sure whether const_casting here is the best approach,
please let me know if you can think of a better solution and I'll
change it.

### DIFF
--- a/include/boost/beast/http/impl/fields.ipp
+++ b/include/boost/beast/http/impl/fields.ipp
@@ -609,11 +609,11 @@ basic_fields<Allocator>::
 erase(const_iterator pos) ->
     const_iterator
 {
-    auto next = pos.iter();
+    auto next = pos;
     auto& e = *next++;
     set_.erase(e);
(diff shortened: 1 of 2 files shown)


vinniefalco:
>Not sure whether const_casting here is the best approach

@pdimov suggests it is correct:

* a precondition of deallocate is that the pointer came from allocate
* allocate gives non-const pointers
* therefore, the element can't have been const to begin with, and the const_cast is safe

inetic:
Thanks for the info, and happy it's OK. Out of curiosity, would you know what's happening with Travis? The Linux builds took about a day to start and the Mac build has been hanging there for more than two days now.

pdimov:
Travis has a ~3000 macOS job backlog, https://www.traviscistatus.com/

And the queued macOS jobs reduce the Linux parallelism from 5 to 3, so even Linux slows down.

Issue #982, opened 2018-01-15
file_mode::write_new fails on windows


chrisics:
Due to what seems to be compliance issue of the standard c library of Microsoft, 
The call open() of file_stdio, with file_mode "write_new" translates to a call to 

std::fopen (file_name, "wbx");

And this fails with "invalid argument passed to c-function". 

The "x" is not supported: See https://msdn.microsoft.com/en-us/library/yeby3zcb.aspx

(Beast version 144)

vinniefalco:
Can you please submit a pull request with a fix?

vinniefalco:
This should be fixed now

PR #999, opened 2018-01-28
file_mode::write_new available only on C++17 or later

chrisics:
Fixes issue https://github.com/boostorg/beast/issues/982

Since we don't want to loose the "write_new" option to open files, 
and since this option is only available in std::fopen since c++17 
(apart from the posix and win32 api's)
I suggest to assert that we are dealing with c++17, in case we are using std::fopen.

### DIFF
--- a/include/boost/beast/core/impl/file_stdio.ipp
+++ b/include/boost/beast/core/impl/file_stdio.ipp
@@ -105,6 +105,10 @@ open(char const* path, file_mode mode, error_code& ec)
         return;
     }
 #else
+    #if (__cplusplus <= 201402L)
+        // supported by std::fopen since c++17
+        BOOST_ASSERT(mode != file_mode::write_new);
+    #endif
     f_ = std::fopen(path, s);
     if(! f_)
     {


include/boost/beast/core/impl/file_stdio.ipp
```diff
@@ -105,6 +105,10 @@ open(char const* path, file_mode mode, error_code& ec)
         return;
     }
 #else
+    #if (__cplusplus <= 201402L)
+        // supported by std::fopen since c++17
+        BOOST_ASSERT(mode != file_mode::write_new);
```
vinniefalco: How about if we make the `file_mode::write_new` constant available only in C++17 and later? And the documentation can reflect this?

vinniefalco:
I think we would be better make `write_new` available only in C++17 this way they get an error at compile time instead of an assertion at run-time. And no one can accidentally make a program using Beast that only works in C++17 yet still compiles in C++11. What do you think?

vinniefalco:
Hmm this problem is a bit more nuanced than I originally thought. Only `file_stdio` doesn't support `write_new` on C++11, the other file implementations do support it (`file_win32` and `file_posix`). So the constant cannot be made conditionally available at compile time depending on language version.

Furthermore, `__cplusplus` is not always a reliable method of checking for features. We should define a macro, say `BOOST_BEAST_HAS_C11_FILE_MODES` and if that is not set, then throw an exception on `write_new` and `append_new`. Then we can work out the logic for setting that macro separately. The macro should be documented.

vinniefalco:
This took forever, but I have added code to deal with the missing modes in some cases. And I have removed the superfluous file mode (`append_new`.. lol). However I think this case is still a problem. Availability of `"wbx"` mode is a C level feature not C++, I think. I'm not sure this is the right way to test it. I am investigating...

vinniefalco:
I believe this can be fixed by emulating the mode. In fact I have already merged code to do so:
https://github.com/boostorg/beast/blob/develop/include/boost/beast/core/impl/file_stdio.hpp#L103

The problem is that the condition `__cplusplus <= 201402L` is not necessarily indicative of lack of support for the feature. For example, my `_MSC_VER` is 1910, my `__cplusplus` is less than `201402L`, but "wbx" works.

vinniefalco:
I believe this is resolved in the **master** and **develop** branches, could you please try it?

PR #1002, opened 2018-01-31
fix memory leak in advanced server examples

bebuch:
emplace_back may throw and at this time the pointer created with new is a raw unmanaged pointer that will never be deleted.

Beast is C++11, unfortunately the simple solution make_unique is introduced by C++14.


bebuch:
Ah, of course, thank you! I will change the request. :-)

vinniefalco:
Merged to v154, thanks!

Issue #941, opened 2017-12-16
Provide convenience alias for public type traits


djarek:
A convenience alias should be provided for public traits, similarly to C++14's or ASIO's traits, example:
```cpp
template<typename T>
using get_lowest_layer_t = typename get_lowest_layer<T>::type
```

vinniefalco:
Which of the other public traits need a similar treatment?

djarek:
Seems that `get_lowest_layer` is the only trait this is applicable to, because C++11 doesn't allow variable templates.

vinniefalco:
We should just change the API of `get_lowest_layer` to be the alias. This would be a breaking change. I doubt anyone will be using this obscure type.

PR #1016, opened 2018-02-10
Provide the type trait alias get_lowest_layer_t

djarek:
Added a C++14-style type alias (get_lowest_layer_t<T>) for the type trait get_lowest_layer<T>.

Resolves: #941

Signed-off-by: Damian Jarek <damian.jarek93@gmail.com>


vinniefalco:
But this requires C++14 does it not?

glenfe:
No, it requires only C++11 template aliases. (i.e. BOOST_NO_CXX11_TEMPLATE_ALIASES is not defined).

PR #1021, opened 2018-02-15
Add field enum entry for "Access-Control-Expose-Headers"

ivilata:
This response header is described both in [CORS](https://www.w3.org/TR/cors/) and [Fetch](https://fetch.spec.whatwg.org/) specifications.

### DIFF
--- a/include/boost/beast/http/field.hpp
+++ b/include/boost/beast/http/field.hpp
@@ -38,6 +38,7 @@ enum class field : unsigned short
     access_control_allow_headers,
     access_control_allow_methods,
     access_control_allow_origin,
+    access_control_expose_headers,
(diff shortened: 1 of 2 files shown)


vinniefalco:
I'm always glad to add more fields but note that the source code is generated from this script:
https://github.com/boostorg/beast/blob/develop/tools/make_field.sh

Which uses this file as input:
https://github.com/boostorg/beast/blob/develop/tools/field.txt

If I run the script again, your change would be lost. I think you should also add it to field.txt and test it, if you wouldn't mind - thanks!

ivilata:
Oh sorry, I wasn't aware about that generation step. I'll create a different PR following that approach.

Thanks!

PR #1022, opened 2018-02-15
Version 155

vinniefalco:
* Fix memory leak in advanced server examples
* Fix soft-mutex assert in websocket stream
* Fix fallthrough warnings
* Tidy up bind_handler doc

### DIFF
--- a/include/boost/beast/core/bind_handler.hpp
+++ b/include/boost/beast/core/bind_handler.hpp
@@ -28,8 +28,8 @@ namespace beast {
     which are not matched to placeholders are silently discarded.
 
     The passed handler and arguments are forwarded into the returned
-    handler, which inherits the associated allocator and associated
-    executor of the original handler.
+    handler, whose associated allocator and associated executor will
+    will be the same as those of the original handler.
 
     Unlike `boost::asio::io_context::wrap`, the returned handler can
     be used in a subsequent call to `boost::asio::post` instead of
--- a/include/boost/beast/websocket/detail/stream_base.hpp
+++ b/include/boost/beast/websocket/detail/stream_base.hpp
@@ -17,6 +17,7 @@
 #include <boost/beast/core/buffers_suffix.hpp>
 #include <boost/beast/core/error.hpp>
 #include <boost/asio/buffer.hpp>
+#include <boost/type_index.hpp>
 #include <cstdint>
 #include <memory>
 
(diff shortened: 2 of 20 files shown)


include/boost/beast/core/bind_handler.hpp
```diff
@@ -23,18 +23,18 @@ namespace beast {
...
     the original handler with the list of bound arguments. Any
-    parameters passed in the invocation will be subtituted for
+    parameters passed in the invocation will be substituted for
     placeholders present in the list of bound arguments. Parameters
     which are not matched to placeholders are silently discarded.
+
     The passed handler and arguments are forwarded into the returned
-    handler, which provides the same `io_context` execution guarantees
-    as the original handler.
+    handler, which inherits the associated allocator and associated
```
djarek: I think inherits might be misleading because it might cause the reader to think that the binder **inherits from** the associated allocator and executor. Perhaps **forwards** is a better name for this relationship?

include/boost/beast/websocket/impl/close.ipp
```diff
@@ -125,31 +123,26 @@ operator()(
     BOOST_ASIO_CORO_REENTER(*this)
     {
         // Maybe suspend
-        if(! d.ws.wr_block_)
+        if(d.ws.wr_block_.try_lock(this))
```
djarek: What if paused_close_.emplace() (line 136) throws an exception?

include/boost/beast/core/bind_handler.hpp
vinniefalco: true, willfix

include/boost/beast/websocket/impl/close.ipp
vinniefalco: It should propagate to the thread calling `io_context::run` like any other exception

vinniefalco: That's right, if an exception occurs in a composed operation the only subsequent permissible operation on the stream is destruction.

include/boost/beast/websocket/detail/stream_base.hpp
```diff
@@ -25,6 +26,86 @@ namespace beast {
...
+class type_mutex
+{
+    boost::typeindex::type_index ti_ = typeid(type_mutex);
+
+public:
+    type_mutex() = default;
+    type_mutex(type_mutex const&) = delete;
+    type_mutex& operator=(type_mutex const&) = delete;
+
+    type_mutex(type_mutex&& other)
```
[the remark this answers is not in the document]
vinniefalco: Good point!!

PR #1031, opened 2018-02-19
*Not yet working* WebSocket: fixes masking for big endian processors

smipi1:


vinniefalco:
I think this is the wrong approach. Integers are by their nature endian-independent with respect to mathematical operations. Shifting an integer to the right by a certain number of bits has the same effect on the value no matter the platform. The code was _supposed_ to be written in a way that is endian-agnostic by taking advantage of this fact. I see what you've done with `byte_at` but I think it would be better to perform the correct transformation in `prepare_key` instead, so that the masking code is not changed.

smipi1:
The problem with the current implementation on develop, is that it uses the key as an integer, but it also ends up exposing the key with regard to the underlying memory representation.

smipi1:
It might be better not to treat the key as an integer at all, but a buffer. That keeps the key and the buffer consistent.

smipi1:
We have something that works, but it completely removes the optimizations for 32 and 64 bit keys.

vinniefalco:
I see what you mean, this looks troublesome:
```
        *reinterpret_cast<std::uint32_t*>(p) ^= key;
```

I could treat the key as a buffer instead but then we couldn't perform the mask operation 4 or 8 bytes at a time without relying on undefined behavior. This optimization should not be disregarded, it is a hot path for servers.

smipi1:
I understand. That was the approach I was trying to take. Make sure that the compiled code on x86 does not change.

smipi1:
#1032 works for all architectures, and illustrates what the code conceptually is trying to do.

vinniefalco:
My thinking is to change the key to be stored little-endian always, so that this line works:
```
  *reinterpret_cast<std::uint32_t*>(p) ^= key;
```
And then just change the preamble and postambles to access the key as a buffer instead of using shifts

smipi1:
The following:

    *reinterpret_cast<std::uint32_t*>(p) ^= key;

Immediately makes the assumption that the key is an endian-less buffer.

smipi1:
If the pre and post ambles cast the key to a buffer and then access the bytes, it should be okay.

smipi1:
The only trick remaining then lies in frame.hpp. It should use a native_uint32_buf_t buffer cast to be consistent with this assumption.

vinniefalco:
I think the key rotation also has to reverse direction, no?

vinniefalco:
Note that the masking algorithms are also used to encode frames when the stream is operating in client mode, something in `write` (frame.hpp) might need adjusting as well

smipi1:
Therein lies the tricky problem. Shifting an integer, but treating it like a buffer results in endianness being exposed. So yes... the direction should reverse for both the shift and the barrel rotate operations.

This all is a result of assuming that the key can be directly applied with an XOR.

smipi1:
That is what I was trying to abstract away by extracting the rotate_key and byte_at functions.

vinniefalco:
Well instead of shifting in the preamble and postamble we could just alias the key to an array of bytes and access the bytes individually. The only ifdef would need to be on the rotate function

smipi1:
Agreed, unless you rotate in array-of-bytes space

smipi1:
That might be cleanest.

smipi1:
Should I clean it up and send another PR?

vinniefalco:
If it isn't too much trouble, that would be great. So to be clear, we're talking about modifying the original implementation to keep the fast 32 and 64 bit functions but just make them work with both endian types?

smipi1:
That's the idea. I will go with the assumption that the key is an integer representation of a native buffer. That makes it consistent with:
    *reinterpret_cast<std::uint32_t*>(p) ^= key;

To be consistent with this, the key will be copied to the header as a buffer using:
    ::new(&b[n]) native_uint32_buf_t{(std::uint32_t)fh.key};

The head and the tail will use an aliased uint8_t array cast of the key to correctly apply the mask and perform the rotation.

smipi1:
Agreed?

vinniefalco:
Yes mostly, although I would rotate the prepare key as an integer using your routines `ror`, `rol`, and `rotate_key` since the key rotation happens inside the loop.

smipi1:
Also... Should I also remove the redundant endian conversion funcions in frame.hpp?

vinniefalco:
>Should I also remove the redundant endian conversion funcions in frame.hpp?

I think so, but note that `fh.key` should have the same value for a given mask no matter the platform. It is the "prepared key" that will be endian-less. That key is generated in `prepare_key`.

smipi1:
Off course. `fh.key` is big endian (network native), but the prepared key is whatever is native to the host.

smipi1:
Just to check that we understand each other correctly: The `fh.key` as sent over the wire should have the same byte order as the masked buffer being sent over the wire, right?

smipi1:
I am also missing where the key rotation is being performed in a loop... Are you referring to the fact that `mask_inplace_fast` might be called in a loop?

vinniefalco:
Sorry I misspoke, there is no key rotation in the innermost loop

smipi1:
Okay. If there is no reason to resort to shifting or barrel shifting for optimization in the fast paths, the code becomes endian independent.

vinniefalco:
Any progress on this? Should I attempt the fix?

vinniefalco:
Ah yes, I remember now why I wrote my own endian routines, because Boost.Endian causes warnings on msvc:
```
>C:\Users\vinnie\src\boost\boost/endian/buffers.hpp(282): warning C4127: conditional expression is constant (compiling source file C:\Users\vinnie\src\boost\libs\beast\test\beast\websocket\stream.cpp)
1>C:\Users\vinnie\src\boost\boost/endian/buffers.hpp(282): note: consider using 'if constexpr' statement instead (compiling source file C:\Users\vinnie\src\boost\libs\beast\test\beast\websocket\stream.cpp)
1>C:\Users\vinnie\src\boost\boost/endian/buffers.hpp(409): note: see reference to function template instantiation 'T boost::endian::detail::load_little_endian<T,4>(const void *) noexcept' being compiled
```

include/boost/beast/websocket/detail/frame.hpp
```diff
@@ -272,7 +272,7 @@ read_close(
     {
         std::uint8_t b[2];
         buffer_copy(buffer(b), cb);
-        cr.code = big_uint16_to_native(&b[0]);
+        cr.code = (::new(&b[0]) big_uint16_buf_t())->value();
```
vinniefalco: This change causes unit test failures

vinniefalco:
I believe this will resolve this issue, please check it:
https://github.com/boostorg/beast/pull/1034

vinniefalco:
The is resolved in the latest version

PR #1032, opened 2018-02-19
*Works* WebSocket fixes for big endian processors

smipi1:


include/boost/beast/websocket/detail/frame.hpp
```diff
@@ -272,7 +273,7 @@ read_close(
     {
         std::uint8_t b[2];
         buffer_copy(buffer(b), cb);
-        cr.code = big_uint16_to_native(&b[0]);
+        cr.code = (::new(&b[0]) big_uint16_buf_t)->value();
```
vinniefalco: is this a separate bug?

smipi1: Oops... Yes. Sorry.

vinniefalco: if you could please submit that in a separate pull request I will add it to 156

vinniefalco: This causes test failures and is generally not correct, it clobbers the value in `b`

vinniefalco:
The is resolved in the latest version

PR #1034, opened 2018-02-20
Version 157

vinniefalco:
* Fix teardown for TIME_WAIT
* Fix big-endian websocket masking

### DIFF
--- a/include/boost/beast/websocket/detail/mask.hpp
+++ b/include/boost/beast/websocket/detail/mask.hpp
@@ -11,6 +11,7 @@
 #define BOOST_BEAST_WEBSOCKET_DETAIL_MASK_HPP
 
 #include <boost/beast/core/detail/config.hpp>
+#include <boost/beast/core/detail/type_traits.hpp>
 #include <boost/asio/buffer.hpp>
 #include <array>
 #include <climits>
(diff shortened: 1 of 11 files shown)


smipi1:
Hi @vinniefalco,

I can confirm that this resolves the issues for us.

I verified this on:
1. `x86_64`: All our unit and component tests are still passing, and
1. `mips` (`GCA4531`): All data is correctly transferred to and from our Websocket portal.

I can also confirm that the close reasons are correctly received on both `x86_64` and `mips`.

Regards,
@smipi1

include/boost/beast/websocket/detail/mask.hpp
```diff
@@ -83,181 +84,59 @@ using maskgen = maskgen_t<std::minstd_rand>;
...
-                unsigned>(sizeof(key) - i);
-            key = ror(key, 8*d);
-            n -= d;
-            p += d;
-            BOOST_FALLTHROUGH;
-        }
-        default:
-            break;
-        }
+        for(int i = 0; i < 4; ++i)
```
smipi1: @vinniefalco,

I quite like this approach. It is clean, readable, and expresses the developer intent. Just a question though:

Does this generate equally efficient x86_64 code when the input buffer is not aligned at the start?

include/boost/beast/websocket/detail/mask.hpp
```diff
@@ -83,181 +84,59 @@ using maskgen = maskgen_t<std::minstd_rand>;
...
     }
-
-    // Mask 8 bytes at a time
-    for(auto i = n / sizeof(key); i; --i)
+    if(n > 0)
     {
-        *reinterpret_cast<
-            std::uint64_t*>(p) ^= key;
-        p += sizeof(key);
+        rol(key, 4 - n);
```
smipi1: I know this is inconsequential because `mask` is used in the loop, but doing the `rol` before the loop trips up the readers eye. Consider doing the `rol` after the loop.

include/boost/beast/websocket/detail/mask.hpp
```diff
@@ -83,181 +84,59 @@ using maskgen = maskgen_t<std::minstd_rand>;
...
-    boost::asio::mutable_buffer const& b,
-        std::uint64_t& key)
+mask_inplace(boost::asio::mutable_buffer& b, prepared_key& key)
 {
     auto n = b.size();
-    auto p = reinterpret_cast<std::uint8_t*>(b.data());
-    if(n >= sizeof(key))
+    auto mask = key; // avoid aliasing
+    auto p = reinterpret_cast<unsigned char*>(b.data());
+    while(n >= 4)
```
smipi1: I suspect that using a `sizeof()` to find the best natural type in stead of `4` might generate optimal code on more architectures and compilers.

include/boost/beast/websocket/detail/mask.hpp
```diff
@@ -83,181 +84,59 @@ using maskgen = maskgen_t<std::minstd_rand>;
...
 
-using prepared_key =
-    std::conditional<sizeof(void*) == 8,
-        std::uint64_t, std::uint32_t>::type;
+using prepared_key = std::array<unsigned char, 4>;
 
 inline
 void
-prepare_key(std::uint32_t& prepared, std::uint32_t key)
+prepare_key(prepared_key& prepared, std::uint32_t key)
```
smipi1: @vinniefalco 
Nobody seems to be using this. Do you still need it?

smipi1:
Hi @vinniefalco,

All-in-all I am very happy with the changes. Do you have any idea when a boost version might be released that includes this?

Regards,
@smipi1

include/boost/beast/websocket/detail/mask.hpp
smipi1: Sorry... vim configuration trouble. Definitely used.

include/boost/beast/websocket/detail/mask.hpp
vinniefalco: You can thank Peter Dimov for this code, and the generated assembly is amazing. Alignment is irrelevant since it uses **movdqu**. You can see the results for various compilers here: https://godbolt.org/g/eNsgnY

include/boost/beast/websocket/detail/mask.hpp
vinniefalco: Inconsequential, the compilers know exactly what is going on here in fact on clang (I think) it unrolls the loop and does 16 bytes at a time.

vinniefalco:
>Do you have any idea when a boost version might be released that includes this?

https://github.com/boostorg/boost/wiki/Release-Schedule

Soon!

include/boost/beast/websocket/detail/mask.hpp
smipi1: Nice! The mips compiler and architecture likely is not that smart, but I don't care too much.

include/boost/beast/websocket/detail/mask.hpp
smipi1: Ack

PR #1052, opened 2018-03-01
Version 162

vinniefalco:
* Add asio_handler_invoke overloads for stream algorithms
* Improve websocket::stream::control_callback javadoc

### DIFF
--- a/include/boost/beast/core/detail/bind_handler.hpp
+++ b/include/boost/beast/core/detail/bind_handler.hpp
@@ -14,6 +14,7 @@
 #include <boost/asio/associated_allocator.hpp>
 #include <boost/asio/associated_executor.hpp>
 #include <boost/asio/handler_continuation_hook.hpp>
+#include <boost/asio/handler_invoke_hook.hpp>
 #include <boost/core/ignore_unused.hpp>
 #include <boost/is_placeholder.hpp>
 #include <functional>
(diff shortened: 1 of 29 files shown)


include/boost/beast/core/detail/bind_handler.hpp
```diff
@@ -153,6 +154,14 @@ class bound_handler
         return asio_handler_is_continuation(std::addressof(h->h_));
     }
 
+    template<class Function>
+    friend
+    void asio_handler_invoke(Function&& f, bound_handler* h)
+    {
+        using boost::asio::asio_handler_invoke;
+        asio_handler_invoke(f, std::addressof(h->h_));
```
sehe: Shouldn't perfect forwarding be applied to `f`?

vinniefalco: I used `Function&&` to combine `Function const&` and `Function&` into a single overload, `asio_handler_invoke` does not take ownership. Do I still need to use perfect forwarding?

glenfe: You can just make it `Function&` since `Function` is a template parameter. It will be appropriate for both `const U&` and `U&` for any type `U`.

vinniefalco: But what happens if it is invoked with a lambda? I get a compile error, "A non-const reference may only be bound to an lvalue"

glenfe: Yes. If that is something you want to allow here, then `Function&&` is fine (or two overloads, `const Function&`, and for `Function&`).

vinniefalco: In other words, it is correct as written :) :) :)

sehe: > I used Function&& to handle Function const& and Function& in a single overload, asio_handler_invoke does not take ownership. Do I still need to use perfect forwarding?

Thanks for clarifying. My bad

vinniefalco: Thanks for looking!

PR #1056, opened 2018-03-03
fix narrowing warnings

DanielaE:
Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/beast/websocket/detail/mask.hpp
+++ b/include/boost/beast/websocket/detail/mask.hpp
@@ -125,7 +125,7 @@ mask_inplace(boost::asio::mutable_buffer& b, prepared_key& key)
     {
         for(std::size_t i = 0; i < n; ++i)
             p[i] ^= mask[i];
-        rol(key, 4 - n);
+        rol(key, static_cast<unsigned>(4 - n));
     }
 }
 


include/boost/beast/websocket/detail/mask.hpp
```diff
@@ -125,7 +125,7 @@ mask_inplace(boost::asio::mutable_buffer& b, prepared_key& key)
     {
         for(std::size_t i = 0; i < n; ++i)
             p[i] ^= mask[i];
-        rol(key, 4 - n);
```
vinniefalco: I guess this is causing warnings for 64-bit builds?

cstratopoulos: @vinniefalco That's been my experience with x64 builds on Visual Studio with recent Beast

DanielaE: @vinniefalco : correct. I do all Boost testing on 64-bit msvc at /W4 to catch such (and similar) kinds of problems.

vinniefalco:
@cstratopoulos Yes, I think that is a good idea

vinniefalco:
Merged in 163, thank you!

DanielaE:
Thanks, Vinnie - the narrowing warning is gone now 😄

PR #1062, opened 2018-03-09
Add field enum entry for "Access-Control-Expose-Headers"

ivilata:
This response header is described both in [CORS](https://www.w3.org/TR/cors/) and [Fetch](https://fetch.spec.whatwg.org/) specifications.


ivilata:
This was previously PR #1021, I added the header to the file used as a source list of headers plus another file where I had forgotten to add the header.

There are probably other headers missing from various HTTP RFCs, but AFAIK this should complete the set of headers used for CORS.

vinniefalco:
Merged to 164

PR #1074, opened 2018-03-19
Replaced add_certificate_authority with equivalent code

bsergeev:
The replacement code handles multiple certificates in the memory buffer, which used to be passed to `boost::asio::ssl::context::add_certificate_authority()`.
If/when my PR for ASIO, https://github.com/chriskohlhoff/asio/pull/294, makes it into Boost ASIO, this code may be removed and call `add_certificate_authority()` restored.

### DIFF
--- a/example/common/root_certificates.hpp
+++ b/example/common/root_certificates.hpp
@@ -95,8 +95,41 @@ load_root_certificates(ssl::context& ctx, boost::system::error_code& ec)
         "-----END CERTIFICATE-----\n"
         ;
 
-    ctx.add_certificate_authority(
-        boost::asio::buffer(cert.data(), cert.size()), ec);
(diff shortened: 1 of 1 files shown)


bsergeev:
Since ASIO accepted my PR with the fix to `boost::asio::ssl::context::add_certificate_authority()` , this workaround won't be necessary, when updated Boost ASIO is released.

vinniefalco:
Thanks for pointing this out. I think we will just wait for the updated Asio, because fiddling with the certificates code is something I don't like to do. I'm not entirely happy with Beast's certificate code anyway since it doesn't work in all cases (it doesn't respect certificates in the operating system's root certificate store).

bsergeev:
Boost 1.67 includes my PR with the fix, so, no changes in Beast are needed!

PR #1103, opened 2018-04-24
Doc: text()/binary() option concerns writing

bebuch:
Make clear text()/binary() option is different from got_text()/got_binary().


vinniefalco:
This will go into 172

vinniefalco:
This got merged, thanks!

PR #1110, opened 2018-04-28
Version 169

vinniefalco:
* Use buffers_to_string in tests
* Use boost::void_t
* Refactor HTTP write_op implementation
* Use fully qualified namespace in BOOST_BEAST_HANDLER_INIT
* New flat_stream example class
* Use flat_stream in ssl_stream example code


CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ if (MSVC)
 
     add_compile_options(
         /bigobj       # large object file format
-        /permissive-  # strict C++
+        #/permissive-  # strict C++
```
djarek: Is this change intentional?

include/boost/beast/core/detail/post.hpp
```diff
@@ -0,0 +1,106 @@
...
+    using executor_type =
+        boost::asio::associated_executor_t<Handler, Executor>;
+
+private:
+    boost::asio::executor_work_guard<Executor> wg1_;
+    boost::asio::executor_work_guard<executor_type> wg2_;
+    Handler h_;
+
+public:
+    post_op(post_op&&) = default;
```
djarek: Redundant declarations?

example/common/flat_stream.hpp
```diff
@@ -0,0 +1,567 @@
...
+    flat_stream& operator=(flat_stream&&) = default;
+    flat_stream& operator=(flat_stream const&) = delete;
+
+    /** Constructor
+
+        Arguments, if any, are forwarded to the next layer's constructor.
+    */
+    template<class... Args>
+    explicit
+    flat_stream(Args&&... args);
```
djarek: This constructor should be `explicit`

example/common/flat_stream.hpp
```diff
@@ -0,0 +1,567 @@
...
+                    auto const result = coalesce(d.first, d.last, limit);
+                    auto const b = buffers_iterator_range(d.first, result.last);
+                    if(result.copy)
+                    {
+                        if(d.p)
+                        {
+                            d.alloc.deallocate(d.p, d.n);
+                            d.p = nullptr;
+                        }
+                        d.p = d.alloc.allocate(result.size);
```
djarek: Is there a reason that the op performs an allocation on every iteration, rather than once? if we need to allocate memory, we can just allocate it once, using `limit` bytes.

CMakeLists.txt
vinniefalco: Yes because another Boost library shits the bed otherwise

include/boost/beast/core/detail/post.hpp
vinniefalco: it isn't redundant, but I don't know if this file should even be there at all

example/common/flat_stream.hpp
vinniefalco: it is

example/common/flat_stream.hpp
vinniefalco: This is a straightforward implementation which doesn't try to get too clever with optimizations. The cost of the allocation is very small compared to the cost of invoking the `ssl::stream` initiating functions. So it isn't clear that allocating `limit` bytes and keeping it around is advantageous.

CMakeLists.txt
vinniefalco: https://github.com/boostorg/range/issues/68

example/common/flat_stream.hpp
djarek: I must be blind

vinniefalco:
I removed detail/post.hpp entirely since it is not used and is probably incorrect

vinniefalco:
The way the `flat_stream` works is that on writes, when a buffer sequence has length greater than one (i.e. multiple buffers in the sequence, which is what Beast emits when serializing HTTP headers), it allocates memory and copies the data into a single buffer which is then sent. It has a limit to how many kilobytes it will allocate. Beyond this limit, it breaks the sequence down into pieces where each piece is no larger than the limit. It sends big blocks of memory individually. This is designed to reduce the number of calls to the `ssl::stream` write function.

>Apologies in advance, if I've done something silly.

It actually looks like I did something stupid...it seems I applied this workaround only to the asynchronous version of write! Sorry about that :) I'll update it.

example/common/flat_stream.hpp
```diff
@@ -0,0 +1,567 @@
...
+            MutableBufferSequence >::value,
+        "MutableBufferSequence  requirements not met");
+    return stream_.async_read_some(
+        buffers, std::forward<ReadHandler>(handler));
+}
+
+template<class NextLayer>
+template<
+    class ConstBufferSequence,
+    class WriteHandler>
```
djarek: `WriteHandler` is a CompletionToken, rather than a CompletionHandler, so the template parameter name may be misleading.

vinniefalco: Yeah, that's a fair point. This could go both ways. Using `WriteHandler` instead of `CompletionToken` gives the reader a tiny bit more information. And it also allows the documentation toolchain to hyperlink the template parameter name, see:
https://www.boost.org/doc/libs/1_67_0/libs/beast/doc/html/beast/ref/boost__beast__http__async_write_some.html

Beast assumes that the user already knows Asio. So `WriteHandler` is more informative.

vinniefalco:
@djarek I rewrote the algorithm to use `buffers_prefix` instead of the iterator range (which definitely wasn't right since the buffer sequence is later moved, invalidating the iterators).

@badfd Please give this a try, it should be much better

badfd:
Now we're cooking with gas! :-) These changes deliver performance to a level slightly better than I had prior to porting.

Here are the new `time(1)` numbers:
```
real	1m6.156s
user	1m15.191s
sys	0m35.824s    # compare with 2m22.124s and 0m41.376s, above
```
and here (for the sake of consistency with previous posts) is the new, high-performing call chain:
```
- 29.82% HttpRequest::send   
  - 28.33% boost::beast::http::write<ssl_stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>, true, boost::beast::http::basi
     - 28.27% boost::beast::http::write<ssl_stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>, true, boost::beast::http::b
        - 28.05% boost::beast::http::write<ssl_stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>, true, boost::beast::http
           - 28.02% boost::beast::http::write_some<ssl_stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>, true, boost::bea
              - 28.00% boost::beast::http::detail::write_some_impl<ssl_stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>,
                 - 22.74% boost::beast::http::serializer<true, boost::beast::http::basic_string_body<char, std::char_traits<char>, std::
                    - 19.52% boost::beast::http::serializer<true, boost::beast::http::basic_string_body<char, std::char_traits<char>, st
                       - 16.10% boost::beast::http::detail::write_some_lambda<ssl_stream<boost::asio::basic_stream_socket<boost::asio::i
                          - 16.07% ssl_stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>::write_some<boost::beast::detail:
                             - 15.94% flat_stream<boost::asio::ssl::stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&> >::w
                                - 6.46% boost::asio::ssl::stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp>&>::write_some<bo
                                   - 6.42% boost::asio::ssl::detail::io<boost::asio::basic_stream_socket<boost::asio::ip::tcp>, boost::a
                                      - 5.97% boost::asio::write<boost::asio::basic_stream_socket<boost::asio::ip::tcp>, boost::asio::mu
                                         - 5.94% boost::asio::write<boost::asio::basic_stream_socket<boost::asio::ip::tcp>, boost::asio:
                                            - 5.90% boost::asio::detail::write_buffer_sequence<boost::asio::basic_stream_socket<boost::a
                                               - 5.76% boost::asio::basic_stream_socket<boost::asio::ip::tcp>::write_some<boost::asio::c
                                                  - 5.72% boost::asio::detail::reactive_socket_service_base::send<boost::asio::const_buf
                                                     - 5.67% boost::asio::detail::socket_ops::sync_send                                     
                                                        - 5.49% 0xec6d                                                                      
                                                           - 5.32% system_call_fastpath                                                     
                                                              - 5.31% sys_sendmsg                                                           
                                                                 - 5.30% __sys_sendmsg                                                      
                                                                    - 5.19% ___sys_sendmsg                                                  
                                                                       + 5.11% sock_sendmsg                                                 
                                + 4.65% detail::flat_stream_base::coalesce<boost::beast::detail::buffers_ref<boost::beast::buffers_prefi
                                + 4.50% boost::asio::buffer_copy<boost::asio::mutable_buffers_1, boost::beast::detail::buffers_ref<boost
                       + 3.35% boost::beast::detail::variant<boost::beast::buffers_prefix_view<boost::beast::buffers_suffix<boost::beast
                    + 2.10% boost::beast::detail::variant<boost::beast::buffers_suffix<boost::beast::detail::buffers_ref<boost::beast::b
                 + 5.15% boost::beast::http::serializer<true, boost::beast::http::basic_string_body<char, std::char_traits<char>, std::a
  + 1.37% boost::beast::http::message<true, boost::beast::http::basic_string_body<char, std::char_traits<char>, std::allocator<char> >,  
```

vinniefalco:
What dominates performance is the **number** of calls into the kernel per unit time, which will be equal to the number of calls to the underlying `boost::asio::ip::tcp::socket` `write_some` or `async_write_some` operations. If you measure the number of calls for the same workload you should see that it is less than or equal to the number of calls in the previous implementation.

vinniefalco:
In the next version merged to **develop** (170 I believe), `flat_stream.hpp` will be accessible using
```
#include <boost/beast/experimental/core/flat_stream.hpp>
```

It will be a public interface with documentation and a quick reference entry.

badfd:
That's great. Thanks for your time and effort in resolving this issue.

PR #1121, opened 2018-05-08
fix the root cert used by example to match the fingerprint in the comment

arvidn:
The comment specifies the fingerprint of the certificate to be
`‎de 28 f4 a4 ff e5 b9 2f a3 c5 03 d1 a3 49 a7 f9 96 2a 82 12`, but the actual certificate is identical to the one above. The certificate with the specified fingerprint can be found here:

  https://www.geotrust.com/resources/root_certificates/certificates/GeoTrust_Global_CA.pem

This patch updates the certificate to actually use the one the comment suggests it is.


vinniefalco:
Not sure how that happened, thanks!

PR #1135, opened 2018-05-18
Pointer-like observers for handler_ptr

djarek:
Add observer functions to handler_ptr to enable convenient checking of whether the handler_ptr is active or not.


vinniefalco:
Part of 171 now

PR #1156, opened 2018-06-10
Version 173

vinniefalco:
* Remove Autobahn testsuite doc note
* Fix buffers_adapter iterator value type
* Fix buffers_adapter max_size
* Fix buffers_prefix iterator decrement
* Add icy_stream Shoutcast stream filter


vinniefalco:
Merged

PR #1168, opened 2018-06-19
Delay construction of BodyReader in parser

djarek:
Enable usage of BodyReaders which need to inspect header contents in their constructor by delaying the construction of the reader until the header is complete.

### DIFF
--- a/include/boost/beast/http/impl/parser.ipp
+++ b/include/boost/beast/http/impl/parser.ipp
@@ -18,20 +18,12 @@ namespace beast {
 namespace http {
 
 template<bool isRequest, class Body, class Allocator>
-parser<isRequest, Body, Allocator>::
-parser()
-    : rd_(m_.base(), m_.body())
-{
-}
-
-template<bool isRequest, class Body, class Allocator>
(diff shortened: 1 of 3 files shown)


vinniefalco:
Hmm.. I don't know how I feel about this.  Both the parser and serializer need to construct the associated reader and writer objects immediately so the caller can access them if necessary to perform implementation-specific tasks. For example to configure settings related to reading or writing (such as setting a buffer size or choosing an allocator).

I realize that `parser` is deficient because it does not expose these interfaces yet (e.g. `parser::reader_impl()` to match `serializer::writer_impl()`) but they will need to be added. At that point we need to figure out how we want to get the headers to the reader. Perhaps it will be a separate function call.

For now this has to wait until after Cppcon.

vinniefalco:
The reader/writer system needs more design work as I have uncovered more problems with it...this will have to wait until 2019 because of the wg21 meeting and the Boost release. Sorry for the delay!

vinniefalco:
We will have to revisit this in the Great HTTP Refactor

PR #1197, opened 2018-07-17
[Test] Remove few unused variables

JoeLoser:
Problem:
- There are a few unused instances of `std::error_code` in some tests.

Solution:
- Remove the unused instances to silence the unused variable warnings I am getting during builds.
- Fix some extra whitespace issues in those files I touched with leading whitespace.

### DIFF
--- a/test/beast/http/write.cpp
+++ b/test/beast/http/write.cpp
@@ -757,7 +757,6 @@ public:
             }
             {
                 auto m = m0;
-                error_code ec;
                 response_serializer<Body, fields> sr{m};
                 sr.split(true);
                 for(;;)
@@ -771,7 +770,6 @@ public:
             }
             {
                 auto m = m0;
-                error_code ec;
                 response_serializer<Body, fields> sr{m};
                 sr.split(true);
                 for(;;)
@@ -804,7 +802,6 @@ public:
             }
             {
                 auto m = m0;
-                error_code ec;
                 response_serializer<Body, fields> sr{m};
                 sr.split(true);
                 for(;;)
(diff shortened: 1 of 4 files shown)


JoeLoser:
@vinniefalco valid concern. Does your test matrix test a Clang 6 variant (6.0 or 6.0.1) with -std=c++17? That would match my local build which triggered the warnings.

JoeLoser:
Right. I digged a bit into the top-level `CMakeLists.txt` along with your `build_and_test.sh` script in the tools folder.

I am a bit surprised MSVC on `/W4` didn't trigger any warnings as [C4101](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4101) is enabled on `/W3` or higher.  Similarly, the combination of `-Wall -Wextra -Wpedantic` should flag such occurrences (and it does). I'm guessing you don't see unused variable for these `std::error_code` on the UNIX builds with Travis in the logs with top of tree?

As far I can tell, I'd agree with your assessment that you should get a compiler warning here and it is likely an MSVC bug with the lambda nearby of the same type and identifier (though it is a tighter scope in the lambda) -- just a guess. I have not searched around on MSVC bug list.

In any regard, is this Travis error with the Apple Clang variant expected? It looks to be missing a symbol for linking -- unrelated to my specific changes at hand. Does this happen often, is it repeatable, etc?

vinniefalco:
> In any regard, is this Travis error with the Apple Clang variant expected? 

It is expected in the sense that I know about it, and it won't go away. It seems to have to do with recent changes in how the Boost libraries are compiled and located, which is confusing Apple clang on Travis for some reason. Two key Boost engineers that might know how to resolve it are unavailable for a week or two.

Of course, if you can figure it out then you will unlock a very special achievement: my eternal gratitude!

PR #1210, opened 2018-07-26
fix http_server_stackless_ssl.cpp example

felixguendling:
write to self_.stream_ instead of the socket_ directly


vinniefalco:
You're right!

PR #1211, opened 2018-07-29
Fix parse_dec algorithm

compmaniak:
Current `parse_dec` implementation is wrong in general. Iterating over `[it, last)` range implies that you shouldn't dereference an iterator pointing to `last`. However parsing actually stops when `*last` is not a digit. Also function unable to parse a number close to `numeric_limis<T>::max()` (for example 6553x for `unsigned short` argument).


vinniefalco:
Whoops! Thanks!

PR #1227, opened 2018-08-15
add socks proxy for example.

Jackarain:


vinniefalco:
Wow, there's a lot here :)

example/socks/http.cpp
```diff
@@ -0,0 +1,355 @@
...
+    boost::asio::yield_context yield)
+{
+    boost::system::error_code ec;
+
+    // These objects perform our I/O
+    tcp::resolver resolver{ioc};
+    tcp::socket socket{ioc};
+
+    socks_url url;
+    if (!parse_socks_url(socks_server, url))
```
vinniefalco: `parse_socks_url` does not return an error code, so `ec` will never get set

example/socks/http.cpp
```diff
@@ -0,0 +1,355 @@
...
+    auto const results = resolver.async_resolve(url.host, url.port, yield[ec]);
+    if (ec)
+        return fail(ec, "resolve");
+
+    // Make the connection on the IP address we get from a lookup
+    boost::asio::async_connect(socket, results.begin(), results.end(), yield[ec]);
+    if (ec)
+        return fail(ec, "connect");
+
+    socks::async_handshake(socket, host, static_cast<unsigned short>(std::atoi(port.c_str())),
```
vinniefalco: Where does one get a socks server to test with?

vinniefalco: If I am reading this right, then to operate a socks client you simply connect to the server and perform the handshake. From that point on, the socket behaves as a transparent tunnel? So no additional state has to be maintained for the lifetime of the connection, and the proxy connection algorithm can be expressed simply as a free function?

example/socks/http.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+int main(int argc, char** argv)
+{
+    // Check command line arguments.
+    if(argc != 5)
+    {
+        std::cerr <<
+            "Usage: socsk <host> <port> <target> <socks[4|5]://[user[:password]@]server:port>\n" <<
+            "Example:\n" <<
+            "    socsk www.example.com 80 / socks5://socks5server.com:1080\n";
```
vinniefalco: The example line is wrong. "socsk" is misspelled. I think this program should be renamed to "socks-client-coro" and the source file renamed to socks_client_coro.cpp

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/streambuf.hpp>
+#include <boost/asio/read.hpp>
+#include <boost/asio/write.hpp>
+
+#include <iostream>
+#include <string>
+#include <memory>
+
+using boost::asio::ip::tcp;
```
vinniefalco: We shouldn't have a `using` directive at file scope in a header file. I know it sucks typing all of that out but it is better for users if we don't do this.

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+
+namespace socks {
+
+    template <typename Handler>
+    class socks_client
+        : public std::enable_shared_from_this<socks_client<Handler>>
+    {
+    public:
+        enum {
```
vinniefalco: these enums don't need to be public

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+        };
+        enum {
+            SOCKS4_REQUEST_GRANTED = 90,
+            SOCKS4_REQUEST_REJECTED_OR_FAILED,
+            SOCKS4_CANNOT_CONNECT_TARGET_SERVER,
+            SOCKS4_REQUEST_REJECTED_USER_NO_ALLOW,
+        };
+
+    public:
+        explicit socks_client(tcp::socket& socket,
```
vinniefalco: the client should be templated on `NextLayer` and only require the SyncReadStream and SyncWriteStream concepts for synchronous client operation, or the AsyncReadStream and AsyncWriteStream concepts for asynchronous client operation.

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+            , handler_(handler)
+        {}
+
+        void handshake()
+        {
+            auto self = std::enable_shared_from_this<socks_client<Handler>>::shared_from_this();
+            switch (version_)
+            {
+            case SOCKS_VERSION_5:
+                boost::asio::spawn(socket_.get_io_context(), [this, self]
```
vinniefalco: The implementation of the composed operation should not use a stackful coroutine. Instead, it should be written using the stackless coroutine API that comes with Asio. There's a tutorial in the Beast docs which can guide you:
https://www.boost.org/doc/libs/1_68_0/libs/beast/doc/html/beast/using_io/writing_composed_operations.html

I can also help you get started by answering questions and give you some initial declarations, if you desire.

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+                });
+            break;
+            default:
+                BOOST_ASSERT("socks version error!" && false);
+            }
+        }
+
+    private:
+        template<typename type, typename source>
+        type read(source& p)
```
vinniefalco: This can be a `static` function. And even better, move it into a base class `detail::socks_base` so that it is not templated on `Handler`. Rationale: By making `read` and `write` needlessly dependent on a template parameter, the resulting executable will have extra copies of these functions for programs which use more than one handler.

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+        template<typename type, typename target>
+        void write(type v, target& p)
+        {
+            for (auto i = (int)sizeof(type) - 1; i >= 0; i--, p++)
+                *p = static_cast<unsigned char>((v >> (i * 8)) & 0xff);
+        }
+
+        void do_socks5_handshake(boost::asio::yield_context yield)
+        {
+            boost::asio::streambuf request;
```
vinniefalco: instead of `asio::streambuf`, consider using `beast::static_buffer` or `beast::flat_static_buffer`. Rationale: If the strings which make up the SOCKS handshake have an upper limit on their size, use of a static buffer can eliminate the need for a dynamic memory allocation. If you do this, you may need to also use `handler_ptr` since a composed operation state containing a static buffer is non-trivial.

example/socks/socks_client.hpp
```diff
@@ -0,0 +1,491 @@
...
+                else
+                {
+                    write<uint8_t>(2, req1); // 2 authentication methods
+                    write<uint8_t>(0, req1); // no authentication
+                    write<uint8_t>(2, req1); // username/password
+                }
+
+                request.commit(bytes_to_write);
+                boost::asio::async_write(socket_, request,
+                    boost::asio::transfer_exactly(bytes_to_write), yield[ec]);
```
vinniefalco: You can probably simplify this line (and others like it) to:
```
boost::asio::async_write( socket_, request.data(), yield[ec]);
```

Jackarain:
all right now, almost rewritten, thank you for taking the time to review the code.

Jackarain:
test server: 
socks5://98.162.25.29:1080
socks4://98.162.25.29:1080
or
http://www.gatherproxy.com/zh/sockslist

dmorilha:
very cool!

vinniefalco:
I'm working on integrating this, I will have a branch for you to look at this week.

vinniefalco:
See: https://github.com/vinniefalco/beast/commits/socks

Note that I am still working on this, there is a lot to clean up.

vinniefalco:
> all right now, almost rewritten, thank you for taking the time to review the code.

I was just comparing the original code versus your new code and I am very impressed by the improvements. This is close to being ready to merge (sorry for taking so long). However there are still some things which need to be done. If you could reset your pull request to this branch, I will comment on the remaining changes: https://github.com/vinniefalco/beast/commits/socks

If you want to squash the commits marked "[FOLD]" that would be fine as well, so it is all in one big commit. I left them unsquashed so you can see each change that I made, and hopefully understand the rationale (I'm happy to answer any questions if you have them).

Jackarain:
Great!
I will reset my pull request to a new branch! Thanks again!

vinniefalco:
I don't see the reset?

vinniefalco:
Reviewing...

example/contrib/socks-client/socks/detail/char_type.hpp
```diff
@@ -0,0 +1,193 @@
...
+        "70", "71", "72", "73", "74", "75", "76", "77", "78", "79", "7a", "7b", "7c", "7d", "7e", "7f",
+        "80", "81", "82", "83", "84", "85", "86", "87", "88", "89", "8a", "8b", "8c", "8d", "8e", "8f",
+        "90", "91", "92", "93", "94", "95", "96", "97", "98", "99", "9a", "9b", "9c", "9d", "9e", "9f",
+        "a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7", "a8", "a9", "aa", "ab", "ac", "ad", "ae", "af",
+        "b0", "b1", "b2", "b3", "b4", "b5", "b6", "b7", "b8", "b9", "ba", "bb", "bc", "bd", "be", "bf",
+        "c0", "c1", "c2", "c3", "c4", "c5", "c6", "c7", "c8", "c9", "ca", "cb", "cc", "cd", "ce", "cf",
+        "d0", "d1", "d2", "d3", "d4", "d5", "d6", "d7", "d8", "d9", "da", "db", "dc", "dd", "de", "df",
+        "e0", "e1", "e2", "e3", "e4", "e5", "e6", "e7", "e8", "e9", "ea", "eb", "ec", "ed", "ee", "ef",
+        "f0", "f1", "f2", "f3", "f4", "f5", "f6", "f7", "f8", "f9", "fa", "fb", "fc", "fd", "fe", "ff"
+    };
```
vinniefalco: Here we have 256 null-terminated, two character strings which is 768 bytes. If we remove the null terminators we can save a third of that and bring this down to 512 bytes, like this:
```
static char const* const s = "00" "01" "02" ...
```
(Note how the commas are removed)

Then we can just do `return string_view{s+c*2, 2};`

example/contrib/socks-client/socks/detail/host.hpp
```diff
@@ -0,0 +1,148 @@
...
+namespace socks {
+namespace detail {
+
+enum host_type
+{
+	host_unkwon,
+	host_ipv4,
+	host_ipv6,
+	host_domain,
+	host_invalid,
```
vinniefalco: There's tabs here instead of spaces

example/contrib/socks-client/socks/handshake.hpp
```diff
@@ -0,0 +1,53 @@
...
+template<
+    typename AsyncStream,
+    typename Handler
+>
+BOOST_BEAST_ASYNC_RESULT1(Handler)
+async_handshake_v4(
+    AsyncStream& stream,
+    const std::string& hostname,
+    unsigned short port,
+    std::string const& username,
```
vinniefalco: I think these `std::string const&` parameters would be better as `string_view`

example/contrib/socks-client/socks/handshake.hpp
```diff
@@ -0,0 +1,53 @@
...
+template<
+    typename AsyncStream,
+    typename Handler>
+BOOST_BEAST_ASYNC_RESULT1(Handler)
+async_handshake_v5(
+    AsyncStream& stream,
+    const std::string& hostname,
+    unsigned short port,
+    std::string const& username,
+    std::string const& password,
```
vinniefalco: `string_view` instead of `std::string const&`

example/contrib/socks-client/socks/handshake.hpp
```diff
@@ -0,0 +1,53 @@
...
+    bool use_hostname,
+    Handler&& handler);
+
```
vinniefalco: Where are the synchronous handshake functions?

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+#include <memory>
+#include <utility>
+#include <type_traits>
+
+
+namespace socks {
+
+namespace detail {
+
+template<typename type, typename source>
```
vinniefalco: `type` and `source` should be capitalized per the style

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+type read(source& p)
+{
+    type ret = 0;
```
vinniefalco: We should use `BOOST_STATIC_ASSERT` to ensure that `Type` is integral (and maybe even unsigned)

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    for (std::size_t i = 0; i < sizeof(type); i++)
+        ret = (ret << 8) | (static_cast<unsigned char>(*p++));
+    return ret;
+}
+
+template<typename type, typename target>
+void write(type v, target& p)
```
vinniefalco: For `read` and `write` why don't we just use `char const*&`  and `char*&` instead of `Source` and `Target`?

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+        *p = static_cast<unsigned char>((v >> (i * 8)) & 0xff);
+}
+
+} // detail
+
+template<
+    class Stream,
+    class Handler,
+    class Buffer,
+    class base_type = boost::beast::async_base<
```
vinniefalco: The `base_type` template parameter is not needed, it is better to just repeat the long typename where needed.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    for (auto i = (int)sizeof(type) - 1; i >= 0; i--, p++)
+        *p = static_cast<unsigned char>((v >> (i * 8)) & 0xff);
+}
+
+} // detail
+
+template<
+    class Stream,
+    class Handler,
+    class Buffer,
```
vinniefalco: We should probably not make `Buffer` a template parameter but instead have the implementation just choose the buffer type that is most appropriate

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    class base_type = boost::beast::async_base<
+        Handler, typename Stream::executor_type>>
+class socks4_op : public base_type
```
vinniefalco: `socks4_op` belongs in the `detail` namespace since it is not a public interface

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    {
+        boost::ignore_unused(bytes_transferred);
+
+        using detail::write;
+        using detail::read;
+
+        auto& response = *response_;
+        auto& request = *request_;
+
+        switch (ec ? 3 : step_)
```
vinniefalco: Instead of `switch` and `step_`, this should use `<boost/asio/coroutine.hpp>` and the macros `BOOST_ASIO_CORO_REENTER`, `BOOST_ASIO_CORO_YIELD`, it will make the code easier to reason about.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+            write<uint32_t>(address.to_uint(), req); // DST I
+
+            if (!username_.empty())
+            {
+                std::copy(username_.begin(), username_.end(), req);    // USERID
+                req += username_.size();
+            }
+            write<uint8_t>(0, req); // NULL.
+
+            request.commit(bytes_to_write);
```
vinniefalco: Instead of building the request here, in `case 0`, we should do it in the constructor. This way, we do not have to store the hostname and username strings, avoiding a memory allocation.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+
+        this->complete_now(ec);
+    }
+
+private:
+    Stream& stream_;
+
+    using BufferPtr = std::unique_ptr<Buffer>;
+    BufferPtr request_{ new Buffer() }; // std::make_unique c++14 or later.
+    BufferPtr response_{ new Buffer() };
```
vinniefalco: I don't think it is necessary to have two buffers. After sending the request, we can free the memory and then reallocate. Or perhaps we can recycle the one buffer - the smallest request is 9 bytes, and it looks like the response is always 8 bytes, so we can get away with just storing one `std::unique_ptr<char[], my_deleter>`

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    unsigned short port_;
+    std::string username_;
+    int step_ = 0;
+};
+
+template<
+    class Stream,
+    class Handler,
+    class Buffer,
+    class base_type = boost::beast::async_base<
```
vinniefalco: We can probably do away with `Buffer` and `base_type`

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+        bool use_hostname)
+        : base_type(std::move(handler), stream.get_executor())
+        , stream_(stream)
+        , hostname_(hostname)
+        , port_(port)
+        , username_(username)
+        , password_(password)
+        , use_hostname_(use_hostname)
+    {
+        (*this)({}, 0); // start the operation
```
vinniefalco: If we can build the initial request in the constructor that would be best, although it looks like we might need to save the username and password. Although we could do that in a single buffer instead of using two strings.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    {
+        boost::ignore_unused(bytes_transferred);
+
+        using detail::write;
+        using detail::read;
+
+        auto& response = *response_;
+        auto& request = *request_;
+
+        switch (ec ? 10 : step_)
```
vinniefalco: Instead of switch and step_, this should use <boost/asio/coroutine.hpp> and the macros BOOST_ASIO_CORO_REENTER, BOOST_ASIO_CORO_YIELD, it will make the code easier to reason about.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+                request.prepare(std::max<std::size_t>(bytes_to_write, 22)).data());
+
+            write<uint8_t>(detail::SOCKS_VERSION_5, req); // SOCKS VERSION 5.
+            write<uint8_t>(detail::SOCKS_CMD_CONNECT, req); // CONNECT command.
+            write<uint8_t>(0, req); // reserved.
+
+            if (use_hostname_)
+            {
+                write<uint8_t>(detail::SOCKS5_ATYP_DOMAINNAME, req); // atyp, domain name.
+                BOOST_ASSERT(hostname_.size() <= 255);
```
vinniefalco: If protocol strings have upper limits, these need to be enforced at runtime by something stronger than just an assert. The handshaking functions need to check the parameters and throw if they are out of bounds, e.g.
```
    if (hostname.size() > 255)
        BOOST_THROW_EXCEPTION(std::out_of_range{"hostname.size() > 255"});
```

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+template<
+    typename AsyncStream,
+    typename Handler
+>
+BOOST_BEAST_ASYNC_RESULT1(Handler)
+async_handshake_v4(
+    AsyncStream& stream,
+    const std::string& hostname,
+    unsigned short port,
+    std::string const& username,
```
vinniefalco: Should use `string_view` instead of `std::string`, and throw if the protocol length limits are exceeded (e.g. `hostname.size() > 255`)

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    net::async_completion<Handler, void(error_code)> init{ handler };
+    using HandlerType = typename std::decay<decltype(init.completion_handler)>::type;
+
+    static_assert(boost::beast::detail::is_invocable<HandlerType,
+        void(error_code)>::value, "Handler type requirements not met");
+
+    using Buffer = net::basic_streambuf<typename std::allocator_traits<
+        net::associated_allocator_t<HandlerType>>:: template rebind_alloc<char> >;
+
+    socks4_op<AsyncStream, HandlerType, Buffer>(stream, init.completion_handler,
```
vinniefalco: This needs to use the new `async_initiate`, perhaps @djarek can help?

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+template<
+    typename AsyncStream,
+    typename Handler>
+BOOST_BEAST_ASYNC_RESULT1(Handler)
+async_handshake_v5(
+    AsyncStream& stream,
+    const std::string& hostname,
+    unsigned short port,
+    std::string const& username,
+    std::string const& password,
```
vinniefalco: `string_view` instead of `std::string`, and throw on length overflows

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    net::async_completion<Handler, void(error_code)> init{ handler };
+    using HandlerType = typename std::decay<decltype(init.completion_handler)>::type;
+
+    static_assert(boost::beast::detail::is_invocable<HandlerType,
+        void(error_code)>::value, "Handler type requirements not met");
+
+    using Buffer = net::basic_streambuf<typename std::allocator_traits<
+        net::associated_allocator_t<HandlerType>>:: template rebind_alloc<char> >;
+
+    socks5_op<AsyncStream, HandlerType, Buffer>(stream, init.completion_handler,
```
vinniefalco: This needs to use the new `async_initiate`

example/contrib/socks-client/socks/query.hpp
```diff
@@ -0,0 +1,141 @@
...
+        : s_(s)
+    {
+    }
+
+    class const_iterator
+    {
+    public:
+        using value_type = std::pair<string_view, string_view>;
+        using difference_type = std::ptrdiff_t;
+        using reference = value_type&;
```
vinniefalco: Actually this is not right, `reference` can't be an actual reference it has to be a value type, e.g.
```
using reference = value_type;
```

And `value_type` can't be a `std::pair` or else the `const_iterator::operator->` won't work (it can't return a pointer). Instead, I suggest you use this declaration for `value_type`:
```
struct value_type
{
  string_view name;
  string_view value;
  value_type const* operator->() const noexcept
  {
    return this;
  }
};
```

Then you can use `using pointer = value_type const;` and things will work right

example/contrib/socks-client/socks/query.hpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        string_view
+        key() const noexcept
+        {
+            return value_.first;
+        }
+
+        string_view
+        value() const noexcept
```
vinniefalco: You won't need `key` and `value` if you use the struct `value_type` above

example/contrib/socks-client/socks/query.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+        BOOST_BEAST_DECL
+        void
+        increment();
+
+        string_view s_;
+        value_type value_;
+    };
+
+    BOOST_BEAST_DECL
```
vinniefalco: `BOOST_BEAST_DECL` is not needed here (my fault)

example/contrib/socks-client/socks/query.hpp
```diff
@@ -0,0 +1,141 @@
...
+    const_iterator
+    begin() const
+    {
+        return const_iterator{s_};
+    }
+
+    BOOST_BEAST_DECL
```
vinniefalco: `BOOST_BEAST_DECL` is not needed here (my fault)

example/contrib/socks-client/socks/uri.hpp
```diff
@@ -0,0 +1,135 @@
...
+    string_view port_;
+    string_view path_;
+    string_view query_string_;
+    string_view fragment_;
+
+public:
+    uri() = default;
+    ~uri() = default;
+
+    BOOST_BEAST_DECL
```
vinniefalco: all instances of `BOOST_BEAST_DECL` should be replaced with `inline` (my fault)

vinniefalco:
Do you need help?

Jackarain:
> Do you need help?

Yes, I need more help, and more improvements, thank you! :blush:

vinniefalco:
Which part do you need help with?

illera88:
Hi @Jackarain ,

looking forward for this to be merged.

vinniefalco:
I haven't forgotten about this, it will go in after 1.72 is released.

vinniefalco:
I believe this also has to be updated for asio's latest changes: completion token default value and C++14 auto return type deduction.

example/socks/Jamfile
```diff
@@ -0,0 +1,15 @@
...
+# Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/boostorg/beast
+#
+
+exe socks :
+    http.cpp
```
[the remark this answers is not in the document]
madmongo1: confirmed changed

example/socks/Jamfile
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
```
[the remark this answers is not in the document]
madmongo1: confirmed updated

example/socks/error.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+    class error_category_impl
+        : public boost::system::error_category
+    {
+        virtual const char* name() const BOOST_SYSTEM_NOEXCEPT
+        {
+            return "SOCKS";
+        }
+
+        virtual std::string message(int e) const
```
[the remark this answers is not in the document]
madmongo1: confirmed fixed

example/socks/http.cpp
```diff
@@ -0,0 +1,355 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#include "socks_client.hpp"
+
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
```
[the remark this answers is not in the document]
madmongo1: confirmed updated

example/socks/http.cpp
```diff
@@ -0,0 +1,355 @@
...
+    std::string username;
+    std::string password;
+};
+
+inline bool is_scheme_char(const char c)
+{
+    return isalpha(c) || isdigit(c) || c == '+' || c == '-' || c == '.';
+}
+
+inline bool parse_socks_url(const std::string& url, socks_url& result)
```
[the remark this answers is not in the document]
vinniefalco: Update: There's a separate URL library with docs and tests, so this url code can be okay as-is until that library becomes part of boost.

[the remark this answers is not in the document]
madmongo1: noted

example/socks/http.cpp
```diff
@@ -0,0 +1,355 @@
...
+    auto it = std::find(url.begin(), url.end(), ':');
+    if (it == url.end())
+        return false;
+
+    for (auto first = url.begin(); first != it; first++)
+    {
+        auto& c = *first;
+        if (!is_scheme_char(c))
+            return false;
+        result.scheme.push_back(static_cast<char>(std::tolower(c)));
```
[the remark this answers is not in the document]
madmongo1: has been refactored in terms of `iequals`

omegacoleman:
in terms of anyone willing to push this further, I started to do sth I could in my repo https://github.com/omegacoleman/beast here, feel free to pick them.

vinniefalco:
The problem is that SOCKS is an extensive protocol, and I do not feel comfortable putting something that only partially supports it in as an example.

omegacoleman:
I opted in because i thought it's useful to show others how to build protocols on top of beast. I was doing an unreleased project porting amazon s3 api to beast and the querystring parser here helped.

The incompleteness might not matter so much as an example, as boost.spirit got mini-xml examples which is also incomplete but very useful.

PR #1231, opened 2018-08-22
Experimental: URL parser & tests

coneiric:
Based on work by @Jackarain, an experimental URL parser and unit tests.

Experimental URL (absolute form URI) parser.

### DIFF
--- a/include/boost/beast/experimental/core/uri/buffer.hpp
+++ b/include/boost/beast/experimental/core/uri/buffer.hpp
@@ -1,0 +1,76 @@
+// Copyright (c) 2018 oneiric (oneiric at i2pmail dot org)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_CORE_URI_BUFFER_HPP
+#define BOOST_BEAST_CORE_URI_BUFFER_HPP
+
+#include <boost/beast/core/string.hpp>
+#include <boost/beast/experimental/core/uri/error.hpp>
+#include <boost/beast/experimental/core/uri/parts.hpp>
+#include <vector>
+
+namespace boost {
+namespace beast {
+namespace uri   {
+
+class buffer : public parts {
+  std::vector<char> data_;
+
+public:
+  using const_iterator = std::vector<char>::const_iterator;
+  using value_type = std::vector<char>::value_type;
+  using size_type = std::vector<char>::size_type;
+
+  std::vector<char> &data() { return data_; }
+
+  value_type back() const noexcept { return data_.back(); }
+
+  const_iterator begin() const noexcept { return data_.begin(); }
+
+  const_iterator end() const noexcept { return data_.end(); }
+
+  void push_back(char c) { data_.push_back(c); }
+
+  size_type size() const noexcept { return data_.size(); }
+
+  std::string part_from(const_iterator first, const_iterator last) {
+    BOOST_ASSERT(first >= data_.begin() && first < data_.end());
+    BOOST_ASSERT(last > data_.begin() && last <= data_.end());
+    BOOST_ASSERT(first < last);
+
+    return {first, last};
+  }
+
+  string_view uri() const noexcept { return {data_.data(), data_.size()}; }
+
+  boost::string_view update_uri()
+  {
+    const std::string auth_start = scheme_ == "file" ? ":///" : "://";
+    const boost::string_view new_uri =
+        scheme_ + auth_start
+        + (username_.empty() ? "" : user_info().to_string() + "@") + host_
+        + (port_.empty() ? "" : ":" + port_) + path_
+        + (query_.empty() ? "" : "?" + query_)
+        + (fragment().empty() ? "" : "#" + fragment_);
+    data_.clear();
+    data_.insert(data_.begin(), new_uri.begin(), new_uri.end());
+    return new_uri;
+  }
+
+  void clear()
+  {
+    data_.clear();
+    reset();
+  }
+};
+
+} // namespace uri
+} // namespace beast
+} // namespace boost
+
+#endif  // BOOST_BEAST_CORE_URI_BUFFER_HPP
--- a/include/boost/beast/experimental/core/uri/detail/parser.hpp
+++ b/include/boost/beast/experimental/core/uri/detail/parser.hpp
@@ -1,0 +1,469 @@
+//
+// Copyright (c) 2018 oneiric (oneiric at i2pmail dot org)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_CORE_URI_DETAIL_PARSER_HPP
+#define BOOST_BEAST_CORE_URI_DETAIL_PARSER_HPP
+
+#include <boost/beast/core/string.hpp>
+#include <boost/beast/experimental/core/uri/buffer.hpp>
+#include <boost/beast/experimental/core/uri/rfc3986.hpp>
+
+namespace boost {
+namespace beast {
+namespace uri {
+namespace detail {
+
+class parser_impl {
+  template <typename It>
+  It find_delimiter_or_mismatch(It first, It last,
+                                std::function<bool(char)> delimiter_func,
+                                std::function<bool(char)> match_func,
+                                const bool ending_segment, error_code &ec) {
+    // Search for segment delimiter
+    auto delimiter = std::find_if(first, last, delimiter_func);
+    if (delimiter == last && !ending_segment) {
+      ec = error::syntax;
+      return delimiter;
+    }
+    // Search for characters that do not belong in the segment
+    auto mismatch = std::find_if_not(first, delimiter, match_func);
+    if (mismatch != delimiter) {
+      ec = error::syntax;
+      return mismatch;
+    }
+    return delimiter;
+  }
+
+  template <typename It>
+  It append_decoded_or_char(It first, It last, buffer& out, error_code& ec)
+  {
+    if (*first == '%')
+      {
+        auto ch = pct_decode(++first, last, ec);
+        if (ec)
+          return first;
+        out.push_back(ch);
+        first += 2;
+      }
+    else
+      out.push_back(*(first++));
+
+    return first;
+  }
+
+  template <typename It>
+  It parse_scheme(It first, It last, buffer &out, error_code &ec) {
+    using beast::detail::ascii_tolower;
+
+    /*  scheme         ; = ALPHA / *(ALPHA / DIGIT / "-" / "." / "+") / ":"
+     */
+    if (!is_alpha(*first)) {
+      ec = error::syntax;
+      return first;
+    }
+    const auto scheme_delimiter = [](const char c) { return c == ':'; };
+    const auto is_scheme_char = [](const char c) {
+      return is_alpha(c) || is_digit(c) || c == '-' || c == '.' || c == '+' ||
+             c == ':';
+    };
+    constexpr bool ending_segment = false;
+    auto delimeter = find_delimiter_or_mismatch(
+        first, last, scheme_delimiter, is_scheme_char, ending_segment, ec);
+
+    if (ec)
+      return delimeter;
+
+    while (first != delimeter)
+      out.push_back(ascii_tolower(*(first++)));
+
+    out.scheme(out.part_from(out.begin(), out.end()));
+    out.push_back(*(first++));
+    return first;
+  }
+
+  template <typename It>
+  It parse_username(It first, It last, buffer &out, error_code &ec) {
+    /*  username       ; = *(unreserved / pct-encoded / sub_delims) / ":" / "@"
+     */
+    const auto size = out.size();
+    const auto username_delimiter = [](const char c) {
+      return c == ':' || c == '@';
+    };
+    constexpr bool ending_segment = false;
+    const auto delimeter = find_delimiter_or_mismatch(
+        first, last, username_delimiter, is_pchar, ending_segment, ec);
+
+    if (ec)
+      return delimeter;
+
+    while (first != delimeter) {
+      first = append_decoded_or_char(first, last, out, ec);
+      if (ec)
+        return first;
+    }
+    const auto start = out.begin() + size;
+    out.username(out.part_from(start, out.end()));
+    out.push_back(*(first++));
+    return first;
+  }
+
+  template <typename It>
+  It parse_password(It first, It last, buffer &out, error_code &ec) {
+    /*  password       ; = ":" / *(unreserved / pct-encoded / sub_delims) / "@"
+     */
+    const auto size = out.size();
+    const auto password_delimiter = [](const char c) { return c == '@'; };
+    constexpr bool ending_segment = false;
+    auto delimeter = find_delimiter_or_mismatch(first, last, password_delimiter,
+                                                is_pchar, ending_segment, ec);
+
+    if (ec)
+      return delimeter;
+
+    while (first != delimeter) {
+      first = append_decoded_or_char(first, last, out, ec);
+      if (ec)
+        return first;
+    }
+    const auto start = out.begin() + size;
+    out.password(out.part_from(start, out.end()));
+    out.push_back(*(first++));
+    return first;
+  }
+
+  template <typename It>
+  It parse_host(It first, It last, buffer &out, error_code &ec) {
+    if (*first == '[') {
+      first = parse_ipv6(first, last, out, ec);
+      return first;
+    } else { // Parse IPv4 or registered name
+      first = parse_ipv4_reg(first, last, out, ec);
+      return first;
+    }
+  }
+
+  template <typename It>
+  It parse_ipv6(It first, It last, buffer &out, error_code &ec) {
+    /*  IP-literal = "[" ( IPv6address / IPvFuture  ) "]"
+     *
+     *  IPv6address =                              6( h16 ":" ) ls32
+     *                /                       "::" 5( h16 ":" ) ls32
+     *                / [               h16 ] "::" 4( h16 ":" ) ls32
+     *                / [ *1( h16 ":" ) h16 ] "::" 3( h16 ":" ) ls32
+     *                / [ *2( h16 ":" ) h16 ] "::" 2( h16 ":" ) ls32
+     *                / [ *3( h16 ":" ) h16 ] "::"    h16 ":"   ls32
+     *                / [ *4( h16 ":" ) h16 ] "::"              ls32
+     *                / [ *5( h16 ":" ) h16 ] "::"              h16
+     *                / [ *6( h16 ":" ) h16 ] "::"
+     *
+     *  TODO: need better IPv6 parsing.
+     *  Current rule is serviceable, but doesn't cover all cases.
+     */
+    const auto size = out.size();
+    const auto ipv6_delimiter = [](const char c) { return c == ']'; };
+    const auto is_ipv6_char = [](const char c) {
+      return c == ':' || is_hex(c);
+    };
+    constexpr bool ending_segment = false;
+    // Pre-increment `first` to skip leading bracket
+    const auto delimeter = find_delimiter_or_mismatch(
+        ++first, last, ipv6_delimiter, is_ipv6_char, ending_segment, ec);
+
+    if (ec)
+      return delimeter;
+
+    while (first != delimeter)
+      out.push_back(*(first++));
+
+    // Pre-increment `first` to skip trailing bracket
+    if (++first != last && *first != ':' && *first != '/' && *first != '?' &&
+        *first != '#') {
+      ec = error::syntax;
+      return first;
+    }
+    const auto start = out.begin() + size;
+    out.host(out.part_from(start, out.end()));
+    if (first != last)
+      out.push_back(*first);
+
+    return first;
+  }
+
+  template <typename It>
+  It parse_ipv4_reg(It first, It last, buffer &out, error_code &ec) {
+    /*  IPv4address = dec-octet "." dec-octet "." dec-octet "." dec-octet
+     *
+     *  dec-octet   = DIGIT                 ; 0-9
+     *              / %x31-39 DIGIT         ; 10-99
+     *              / "1" 2DIGIT            ; 100-199
+     *              / "2" %x30-34 DIGIT     ; 200-249
+     *              / "25" %x30-35          ; 250-255
+     *
+     *  reg-name    = *( unreserved / pct-encoded / sub-delims )
+     *
+     *  TODO: need distinct IPv4 parsing.
+     *  Current rule captures IPv4, but is too loose, allowing invalid IPv4.
+     */
+    const auto size = out.size();
+    const auto ipv4_delimiter = [](const char c) {
+      return c == ':' || c == '/' || c == '?' || c == '#';
+    };
+    const auto is_ipv4_char = [](const char c) {
+      return is_unreserved(c) || is_sub_delims(c) || c == '%';
+    };
+    constexpr bool ending_segment = true;
+    const auto delimiter = find_delimiter_or_mismatch(
+        first, last, ipv4_delimiter, is_ipv4_char, ending_segment, ec);
+
+    if (ec)
+      return delimiter;
+
+    while (first != delimiter) {
+      first = append_decoded_or_char(first, last, out, ec);
+      if (ec)
+        return first;
+    }
+
+    const auto start = out.begin() + size;
+    out.host(out.part_from(start, out.end()));
+    return first;
+  }
+
+  template <typename It>
+  It parse_port(It first, It last, buffer &out, error_code &ec) {
+    /*  port          ; = ":" / *DIGIT
+     */
+    if (*first != ':') {
+      ec = error::syntax;
+      return first;
+    }
+    out.push_back(*(first++));
+    const auto size = out.size();
+    const auto port_delimiter = [](const char c) {
+      return c == '/' || c == '?' || c == '#';
+    };
+    constexpr bool ending_segment = true;
+    const auto delimiter = find_delimiter_or_mismatch(
+        first, last, port_delimiter, is_digit, ending_segment, ec);
+
+    if (ec)
+      return delimiter;
+
+    while (first != delimiter)
+      out.push_back(*(first++));
+
+    const auto start = out.begin() + size;
+    out.port(out.part_from(start, out.end()));
+    if (first != last)
+      out.push_back(*first);
+
+    return first;
+  }
+
+  template <typename It>
+  It parse_authority(It first, It last, buffer &out, error_code &ec) {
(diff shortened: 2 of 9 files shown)


include/boost/beast/experimental/core/impl/url_parser.ipp
```diff
@@ -0,0 +1,352 @@
...
+
+#include <boost/beast/experimental/core/url_parser.hpp>
+
+namespace boost {
+namespace beast {
+
+//--------------------------------------------------------------------------
+// scheme:[//[user[:password]@]host[:port]][/path][?query][#fragment]
+
+using boost::beast::string_view;
```
vinniefalco: Hmm.. probably better not to have a `using` statement at namespace scope. `string_view` is already a symbol in Beast you should be able to remove this `using` statement and still have everything compile.

include/boost/beast/experimental/core/impl/url_parser.ipp
```diff
@@ -0,0 +1,352 @@
...
+
+string_view url_parser::make_string_view(const char* b, std::size_t s)
+{
+    return string_view(b, s);
+}
+
+bool url_parser::isunreserved(const char c) const
+{
+    if (std::isalpha(c) || std::isdigit(c) ||
```
vinniefalco: These are slow, they will look at the current locale. Beast has its own private replacements which assume 7-bit ascii (the standard for HTTP and web). Examples:
https://github.com/boostorg/beast/blob/develop/include/boost/beast/core/string.hpp#L33
https://github.com/boostorg/beast/blob/develop/include/boost/beast/http/detail/rfc7230.hpp
https://github.com/boostorg/beast/blob/develop/include/boost/beast/http/detail/basic_parser.hpp#L55

Just to keep things interesting, note that if someone uses Beast and sets the current locale differently, then functions like `std::isdigit` can do the wrong thing with respect to RFC compliance. That is why we must avoid `std` char functions.

include/boost/beast/experimental/core/impl/url_parser.ipp
```diff
@@ -0,0 +1,352 @@
...
+    if (c == '!' || c == '$' || c == '\'' || c == '(' ||
+        c == ')' || c == '*' || c == '+' || c == ',' ||
+        c == '=')
+        return true;
+    return false;
+}
+
+//--------------------------------------------------------------------------
+
+bool url_parser::parse(const std::string& url)
```
vinniefalco: eventually this will need to use error codes but we can ship the first experimental version using plain bool.

include/boost/beast/experimental/core/impl/url_parser.ipp
```diff
@@ -0,0 +1,352 @@
...
+                continue;
+            }
+            return false;
+        case scheme:
+            if (std::isalpha(c) || std::isdigit(c) ||
+                c == '+' || c == '-' || c == '.')
+                continue;
+            if (c == ':')
+            {
+                scheme_ = make_string_view(part_start, b - part_start - 1);
```
vinniefalco: dunno why we need `make_string_view`, I think this also works:
```
scheme_ = { part_start, b - part_start - 1) };
```

include/boost/beast/experimental/core/impl/url_parser.ipp
```diff
@@ -0,0 +1,352 @@
...
+        {
+        case scheme_start:
+            if (std::isalpha(c))
+            {
+                state = scheme;
+                continue;
+            }
+            return false;
+        case scheme:
+            if (std::isalpha(c) || std::isdigit(c) ||
```
vinniefalco: Does the scheme allow uppercase? I'm not sure. If it does, we probably need to convert it to lower case.

include/boost/beast/experimental/core/url_parser.hpp
```diff
@@ -0,0 +1,56 @@
+//
+// Copyright (c) 2018 jackarain (jack dot wgm at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_EXPERIMENTAL_CORE_URL_PARSER_HPP
```
vinniefalco: I don't bother mentioning EXPERIMENTAL in the macro:
```
#ifndef BOOST_BEAST_CORE_URL_PARSER_HPP
```
Code is structured "as-if" it was public, except that it is place in the experimental/ directory. So if we never decide to publish it, then all that is necessary is to delete "experimental/" from all include paths and apply suitable file renaming.

include/boost/beast/experimental/core/url_parser.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace beast {
+
+    //--------------------------------------------------------------------------
+    // scheme:[//[user[:password]@]host[:port]][/path][?query][#fragment]
+
+    using boost::beast::string_view;
+
+    struct url_parser
+    {
+        string_view scheme_;
```
vinniefalco: MInor point, public members don't end with an underscore, e.g. `scheme` instead of `scheme_`.

include/boost/beast/experimental/core/url_parser.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+        bool isuchar(const char c) const;
+
+        bool ishsegment(const char c) const;
+
+        bool issubdelims(const char c) const;
+
+        //--------------------------------------------------------------------------
+
+        bool parse(const std::string& url);
```
vinniefalco: This parameter should have type `string_view`.

include/boost/beast/experimental/core/url_parser.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+        string_view make_string_view(const char* b, std::size_t s);
+
+        bool isunreserved(const char c) const;
+
+        bool isuchar(const char c) const;
+
+        bool ishsegment(const char c) const;
+
+        bool issubdelims(const char c) const;
```
vinniefalco: These should be `static` and probably not public.

include/boost/beast/experimental/core/url_parser.hpp
vinniefalco: Consider a free function instead:
```
url_parts parse_url(string_view s);
```

include/boost/beast/experimental/core/url_parser.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace boost {
+namespace beast {
+
+    //--------------------------------------------------------------------------
+    // scheme:[//[user[:password]@]host[:port]][/path][?query][#fragment]
+
+    using boost::beast::string_view;
+
+    struct url_parser
```
vinniefalco: This needs a Javadoc comment

include/boost/beast/experimental/core/url_parser.hpp
```diff
@@ -0,0 +1,56 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_EXPERIMENTAL_CORE_URL_PARSER_HPP
+#define BOOST_BEAST_EXPERIMENTAL_CORE_URL_PARSER_HPP
+
+#include <boost/beast/core/string.hpp>
+#include <cctype> // for std::isalpha etc...
```
vinniefalco: `<cctype>` is dangerous, RFC compliant programs should never be using anything from there.

include/boost/beast/experimental/core/impl/url_parser.ipp
coneiric: > Actually, I have written a lot of these already for some partial work I did on my own URI parser you can copy them

Awesome, I'll add those as replacements.

I wasn't familiar with the other Beast implementations, thanks for the links.

include/boost/beast/experimental/core/impl/url_parser.ipp
coneiric: Alright, I'll look at other error code implementations in preparation. I thought the bool return was coarse for all the error conditions.

include/boost/beast/experimental/core/impl/url_parser.ipp
coneiric: Was curious about the need for that function. I'll remove it, and use your suggestion.

include/boost/beast/experimental/core/impl/url_parser.ipp
coneiric: Will check the RFC, and convert if it allows uppercase.

test/beast/experimental/url_parser.cpp
```diff
@@ -0,0 +1,83 @@
+//
+// Copyright (c) 2016-2018 Vinnie Falco (vinnie dot falco at gmail dot com)
```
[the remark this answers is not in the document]
coneiric: :) Right on, will remove it, add mine.

test/beast/experimental/url_parser.cpp
```diff
@@ -0,0 +1,83 @@
...
+    doParse("a://1.1.1.1:80/#a",      "a", "", "", "1.1.1.1", "80", "/", "", "a");
+    doParse("a://1.1.1.1:80/a#a",     "a", "", "", "1.1.1.1", "80", "/a", "", "a");
+    doParse("a://1.1.1.1:80/a?b=1#",  "a", "", "", "1.1.1.1", "80", "/a", "b=1");
+    doParse("a://1.1.1.1:80/a?b=1#a", "a", "", "", "1.1.1.1", "80", "/a", "b=1", "a");
+    // IPv6
+    doParse("a://[::1]",                           "a", "", "", "::1");
+    doParse("a://[fe80:1010:1010:1010:1010:1010]", "a", "", "", "fe80:1010:1010:1010:1010:1010");
+    // Host literal
+    doParse("a://boost.org", "a", "", "", "boost.org");
+  }
```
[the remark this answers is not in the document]
coneiric: Thanks :)

test/beast/experimental/url_parser.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  void
+  run() override
+  {
+    testParse();
+  }
+};
+
+BEAST_DEFINE_TESTSUITE(beast,experimental,url_parser);
```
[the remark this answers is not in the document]
coneiric: No worries, will change.

test/beast/experimental/url_parser.cpp
```diff
@@ -0,0 +1,83 @@
...
+// Test that header file is self-contained.
+#include <boost/beast/experimental/core/url_parser.hpp>
+
+#include <boost/beast/unit_test/suite.hpp>
+#include <string>
+
+namespace boost {
+namespace beast {
+
+using boost::beast::string_view;
```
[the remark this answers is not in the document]
coneiric: Ok, will remove it.

test/beast/experimental/url_parser.cpp
```diff
@@ -0,0 +1,83 @@
...
+      string_view scheme = "",
+      string_view username = "",
+      string_view password = "",
+      string_view host = "",
+      string_view port = "",
+      string_view path = "",
+      string_view query = "",
+      string_view fragment = "")
+  {
+    boost::beast::url_parser parser;
```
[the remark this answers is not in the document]
coneiric: Ok, will try it out.

include/boost/beast/experimental/core/url_parser.hpp
coneiric: Oh, that makes complete sense, I like it. I'll make the edits.

include/boost/beast/experimental/core/url_parser.hpp
coneiric: Noted, will change.

include/boost/beast/experimental/core/url_parser.hpp
coneiric: Alright, will make `static`, and move them to private access/impl.

include/boost/beast/experimental/core/url_parser.hpp
coneiric: Ok, I hadn't thought about using a free function.

Guessing `url_parts` will be a wrapper around the public data members?

include/boost/beast/experimental/core/url_parser.hpp
coneiric: Will add one. New to exemplar comments, I'll do my best.

include/boost/beast/experimental/core/url_parser.hpp
coneiric: Will remove it, unnecessary with your `isalpha`, `isdigit`, and `ascii_tolower`.

include/boost/beast/experimental/core/url_parser.hpp
vinniefalco: >a wrapper around the public data members

We call that a POD type (plain old data). Anyway, this is something we can experiment with later, it has less priority than the other stuff.

include/boost/beast/experimental/core/url_parser.hpp
vinniefalco: The bar for stuff in experimental/ is lower so you could at a minimum get away with just one line:
```
/// Parses URLs
```
Or
```
/** Parses URLs
*/
```
Eventually you will want to add something to the quick reference:
https://github.com/boostorg/beast/blob/develop/doc/qbk/quickref.xml#L295

No one has ever thought to update the reference but hey, first time for everything I guess! If you don't do it, I will, so no big deal (and setting up the toolchain for building the docs is kind of a hassle so I will understand).

include/boost/beast/experimental/core/url_parser.hpp
vinniefalco: Related: https://github.com/boostorg/beast/issues/1232 (something to think about when this is merged)

coneiric:
Made the suggested changes, but still working on making `parse_url` better. Need to add the other RFC3986 functions, as well (pchar, etc.).

I forgot to update the Javadoc XML, but will do that in the next set of changes.

Jackarain:
good job!

coneiric:
> good job!

Thanks :)

@vinniefalco I added the remaining RFC3986 functions, error codes, and cleaned up `url_parser.ipp` a bit. Wanted to get it closer to your `uri` branch, but the diffs were getting quite large.

Ran into a problem with docs, after adding this line to `doc/qbk/quickref.xml`:
```xml
<member><link linkend="beast.ref.boost__beast__core__uri__url_parser__parse_url">uri::parse_url</link></member>
```
I got this error:
```
Error: no ID for constraint linkend: beast.ref.boost__beast__core__uri__url_parser__parse_url.
```
Tried a number of alternates for `linkend` value with similar results. I didn't find where IDs are defined for the other entries, so this confuses me a lot. Likely due to me being brand new to BoostBook and all.

Any ideas?

vinniefalco:
`core` is not a namespace, try `boost__beast__uri__url_parser__parse_url`

Also, you don't have to follow the **uri_parser** branch at all, that is just some exploratory coding and should not be considered canon.

vinniefalco:
> Likely due to me being brand new to BoostBook and all.

Actually these ref names are not a function of BoostBook at all but rather the XSL program which transforms the Doxygen XML into Quickbook:
https://github.com/boostorg/beast/blob/develop/doc/docca/include/docca/doxygen.xsl

coneiric:
> core is not a namespace, try boost__beast__uri__url_parser__parse_url

Think I may have tried that, but will keep messing with it until it works.

> Also, you don't have to follow the uri_parser branch at all

Alright, I wasn't sure the status of that code, a lot of it is really nice. Will write more of my own code.

> Actually these ref names are not a function of BoostBook

Thanks for the link, will look into Docca documentation (docs all the way down...).

coneiric:
@vinniefalco sorry for the entire rewrite, but it was cleaner than trying to do it another way.

I kept the functions for testing if a character `is_[some-set]`, but am not using them. Wasn't sure if you would want to keep them for other uses, and can remove if wanted.

Most of the diff is in `detail/parse.hpp` and the unit tests.

If you would like me to break up the PR into small commits, I can. Thanks for your help on this.

Still no luck on the doc issue, but I'm working on it.

vinniefalco:
Rewriting is fine, that is completely your choice. You don't have to break up the pull request, in fact I will squash it down into a single commit. That's usually how I introduce a new feature anyway. It is only later when making bug fixes that I use individual commits, one per corresponding GitHub issue. Unless I am doing a big refactoring, then I will typically squash all the changes down into one before merging.

include/boost/beast/experimental/core/uri/buffer.hpp
```diff
@@ -0,0 +1,54 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_CORE_URI_BUFFER_HPP
+#define BOOST_BEAST_CORE_URI_BUFFER_HPP
+
+#include <vector>
```
vinniefalco: standard library includes should come last

include/boost/beast/experimental/core/uri/buffer.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <boost/beast/experimental/core/uri/parts.hpp>
+
+namespace boost {
+namespace beast {
+namespace uri   {
+
+class buffer : public parts {
+  using const_iterator = std::vector<char>::const_iterator;
+  using value_type = std::vector<char>::value_type;
+  using size_type = std::vector<char>::size_type;
```
vinniefalco: Shouldn't these nested types be public?

include/boost/beast/experimental/core/uri/buffer.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  value_type back() const noexcept { return data_.back(); }
+
+  const_iterator begin() const noexcept { return data_.begin(); }
+
+  const_iterator end() const noexcept { return data_.end(); }
+
+  size_type size() const noexcept { return data_.size(); }
+
+  std::string part_from(const_iterator const start, const_iterator const end) {
```
vinniefalco: There's no need to mark parameter types passed by value as `const`, in fact it doesn't even change the signature of the function. Also you might want to use the names `first` and `last` so they are unambiguous.

include/boost/beast/experimental/core/uri/detail/parser.hpp
```diff
@@ -0,0 +1,340 @@
...
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_CORE_URI_DETAIL_PARSER_HPP
+#define BOOST_BEAST_CORE_URI_DETAIL_PARSER_HPP
+
+#include <vector>
+#include <boost/beast/core/string.hpp>
+#include <boost/spirit/include/qi.hpp>
```
vinniefalco: Whoa...!! I don't think this is a good idea.... Spirit is a massive dependency and if we go down this path then it will be all but impossible to make Beast independent of Boost (something which I will need to do with I propose it as a TS).

include/boost/beast/experimental/core/uri/buffer.hpp
coneiric: Ok, some projects like them first, will move it.

include/boost/beast/experimental/core/uri/buffer.hpp
coneiric: I didn't see a need to expose them publicly, but I'll move them.

include/boost/beast/experimental/core/uri/detail/parser.hpp
vinniefalco: What is wrong with the original code? It seemed to work and was decent...

vinniefalco: When considering adding a dependency to a library (like adding Spirit to Beast) it is good to get a second and third opinion and discuss with stakeholders. Keep in mind that there are security issues at play here. The parsers in Beast are the front line of the attack surface. The historical attacks which specifically target URL parsing implementations would fill a thick book. It is harder to provide security assurances because a security audit of Beast which used Spirit would have to include the Spirit source code in the analysis and not only would that become very expensive but then it raises questions about whether each update to Spirit has subsequently introduced a vulnerability in Beast.

My apologies for the efforts you expended making Qi work (it does look very nice). But perhaps we can chalk it up to a learning experience?

vinniefalco: Examples:
https://www.blackhat.com/docs/us-17/thursday/us-17-Tsai-A-New-Era-Of-SSRF-Exploiting-URL-Parser-In-Trending-Programming-Languages.pdf
http://www.cgisecurity.com/lib/URLEmbeddedAttacks.html
https://www.acunetix.com/blog/whitepaper-http-parameter-pollution/

coneiric: It makes sense that adding Spirit as a Beast dependency is too heavy handed. Especially from the security audit perspective, so much code to review. I am fine with doing another rewrite, though it would be nice if there was an easy solution here.

The original code was OK, but left a lot to be desired. The giant while loop going over a state machine took me some time to understand, and when I tried pulling it apart with functions, things got difficult. It also read more like C/old-C++ than modern C++.

Thanks for the links on URI/URL related bugs. Will digest those, and hopefully gain some new perspective/ideas on writing this parser.

vinniefalco: I would just keep the parser that that is there. This is an experimental interface which means it doesn't have to be perfect. It just needs a tidying to make it meet the minimum quality control standards. Future revisions can always include a rewrite of the parser, which will certainly be necessary before graduating to an official (non-experimental) public interface.

As engineers it is natural for us to look at something and want to improve it, of course I do that all the time. But it is also important to do a reasonably accurate cost/benefit analysis. Here,, users are going from nothing (no URL parser) to something (this URL parser). This is practically infinite improvement! There is little to be gained from rewriting it at this stage.

Doing a light tidying lets us get something into the tip of the develop and master branches sooner rather than later, and it lets us see all the pieces on the board together so to speak. We will probably be able to make better informed decisions with something that is at least semi-official in the experimental branch, especially when users pick it up and start providing feedback. This will let us work smarter instead of harder.

coneiric:
> I would just keep the parser that was there.

I couldn't bring myself to revert to a giant while-loop over a state machine, especially since it would need to be refactored for modularity anyway.

> Doing a light tidying lets us get something into the tip of the develop and master branches sooner rather than later

Getting something ready for use is my main goal, as well. Not sure this latest refactor can be considered "light tidying", but I do hope it gets us closer to the end goal.

Hopefully the changes are relatively easy to review. Thank you for all your help so far.

vinniefalco:
Whatever happened with this?

coneiric:
> Whatever happened with this?

Hi, just saw your comment. I updated the PR with your recommendations to revert closer to the original parser. Made a few updates, and moved the other rewritten parser to my `loops` branch.

Also, added some unit-tests for cases from the SSRF paper you linked.

Please let me know if there are more changes you would like me to make.

vinniefalco:
Okay I finally got some time to look at this. I don't see any of the code from the SOCKS proxy branch? This is just all my code with some added tests? I don't think this is a good idea, my public interfaces are experimental and have problems. For example `static_buffer` and the URI buffer concept in general are not fully thought out and have problems. My routines for categorizing the characters are okay I guess but that's all I would take. Otherwise, we should just base it on the code from the other pull request, which works in the larger context of the SOCKS proxy.

vinniefalco:
Ohhhh....I just had a look at parser.ipp, this is from the other branch?

coneiric:
> I just had a look at parser.ipp, this is from the other branch?

It is, I am actually using something closer to my `loops` branch in another project. I will update that branch with my most recent changes. The `loops` branch also includes changes for correctness, and a different `buffer` class loosely based on yours.

If you are not opposed to it, I would like to switch out the current parser in this PR with the one in my `loops` branch. It is much easier to make changes there, and (imho) it adheres more closely to RFC3986. This is especially true when compared with the parser from the SOCKS PR.

> my public interfaces are experimental 

I loved your use of string view, and was disappointed when I ran into problems getting it to work in another project. Will update the `buffer` class in my `loops` branch with the new interfaces currently in use. Let me know what you think :)

> Otherwise, we should just base it on the code from the other pull request, which works in the larger context of the SOCKS proxy.

The code in my `loops` branch is still loosely based on the code from @Jackarain's PR. The main difference being the split into multiple functions vs. the state machine, and corrections for RFC3986.

vinniefalco:
Yeah I'm okay with whatever implementation you think is best for the parser but I have concerns about using my half-baked ideas, even in experimental/

include/boost/beast/experimental/core/uri/parts.hpp
```diff
@@ -0,0 +1,71 @@
...
+    host_.clear();
+    port_.clear();
+    path_.clear();
+    query_.clear();
+    fragment_.clear();
+  }
+
+public:
+  void scheme(std::string part) { scheme_ = part; }
+  std::string scheme() const noexcept { return scheme_; }
```
kotbegemot: Why use a std::string? Can be use a string_view(C++17/boost.beast).

coneiric: I chose to use std::string because use in Kovri was a bit simpler that way. After a little bit of work, the performance benefits of string_view are worth converting to string when necessary.

coneiric:
Pinging @vinniefalco.

I rebased on the latest develop branch, updated the parser with std algorithms, and made some minor changes to the buffer/parts components.

vinniefalco:
ack

coneiric:
:) What do you think of the most recent set of changes? Is this PR close or ready to be merged?

vinniefalco:
I have not looked closely at it but I will do so just as soon as I am done with the preparation for Boost 1.69 which is coming out shortly

include/boost/beast/experimental/core/uri/detail/parser.hpp
```diff
@@ -0,0 +1,469 @@
...
+namespace boost {
+namespace beast {
+namespace uri {
+namespace detail {
+
+class parser_impl {
+  template <typename It>
+  It find_delimiter_or_mismatch(It first, It last,
+                                std::function<bool(char)> delimiter_func,
+                                std::function<bool(char)> match_func,
```
Eelis: Using std::function here seems like a premature pessimization. Have you considered taking the callback functor types as template parameters instead?

include/boost/beast/experimental/core/uri/rfc3986.hpp
```diff
@@ -0,0 +1,96 @@
...
+  unsigned constexpr zero = '0';
+  return (static_cast<unsigned>(c) - zero) < 10;
+}
+
+bool is_hex(char c) {
+  return (c == 'a' || c == 'b' || c == 'c' || c == 'd' || c == 'e' ||
+          c == 'f' || c == 'A' || c == 'B' || c == 'C' || c == 'D' ||
+          c == 'E' || c == 'F' || is_digit(c));
+}
+
```
Eelis: Did you intend for the above three functions to be inline? (or perhaps even constexpr)

vinniefalco:
See: https://github.com/vinniefalco/url

vinniefalco:
the main repo for Boost.URL is https://github.com/CPPAlliance/url

PR #1237, opened 2018-08-28
add certificate verifying in ssl client examples

PeterW3:
Verify the remote servers' certificates in the ssl examples.


vinniefalco:
Why are removing that one certificate?

PR #1238, opened 2018-08-28
Fix Contributing Link in README

ricejasonf:
All of your contributors missed this one apparently. :stuck_out_tongue:


vinniefalco:
Yep, thanks! Also the directory layout in the README is a bit out of date and could use an update (reminder to myself)

PR #1240, opened 2018-09-02
fixed timer issue after wss upgrade

JarleStrand:
Fixes the timer bug mentioned in Issue 1091 for the advanced_server_flex example.

The bug is related to a timer that is not cancelled after an upgrade to wss. 

Instead of cancelling the timer, this solution ensures that the timer expires immediately, and that the call-back detects that no actions should be performed.

### DIFF
--- a/example/advanced/server-flex/advanced_server_flex.cpp
+++ b/example/advanced/server-flex/advanced_server_flex.cpp
@@ -811,6 +811,10 @@ public:
         if(ec && ec != boost::asio::error::operation_aborted)
             return fail(ec, "timer");
 
+        // Check if this has been upgraded to Websocket
+        if(timer_.expires_at() == (std::chrono::steady_clock::time_point::min)())
+            return;
+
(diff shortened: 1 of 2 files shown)


vinniefalco:
Well I had some plans for a more general fix but this pull request is looking better and better to me. One thing, can we please also fix the advanced-server example? It is here:
https://github.com/boostorg/beast/blob/develop/example/advanced/server/advanced_server.cpp

This is what I was thinking for the general solution:
https://github.com/vinniefalco/beast/blob/timed_stream/include/boost/beast/experimental/core/timed_socket.hpp#L25

It needs to be adapted for WebSocket so you can send a ping, probably I will make the "on timer" action customizable.

JarleStrand:
Hi, I updated the advanced-server example as well with the exact same changes. It seems to work ok.
I agree with you that a more general fix would certainly be better, but at least this will fix some annoying bugs in major examples.

PR #1241, opened 2018-09-05
Replace boost string_view with std string_view if macro is set.

### LINKED ISSUE #1133: std::string_view option
It should be possible for users to define BOOST_BEAST_USE_STD_STRING_VIEW or some such, and have beast use `std::string_view` (and `std::basic_string_view`) in `<boost/beast/core/string.hpp>`

do-m-en:
Closes #1133 

The solution uses a naive approach of expecting the user knows whether compiler implements std::string_view (not using/defining any compatibility check macros)  and also ignoring the <experimental/string_view> version.

### DIFF
--- a/include/boost/beast/core/string.hpp
+++ b/include/boost/beast/core/string.hpp
@@ -12,19 +12,35 @@
 
 #include <boost/beast/core/detail/config.hpp>
 #include <boost/version.hpp>
+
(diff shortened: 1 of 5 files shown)


vinniefalco:
LOL is that it?! I did most of the heavy lifting and didn't seem to cross the finish line...

Issue #1245, opened 2018-09-14
Invalid UTF8 payload error for valid Unicode strings


bsergeev:
### Version of Beast
1.68.0 in Windows 10 and 1.67.0 in Ubuntu 16

### Steps necessary to reproduce the problem
Run the code below with `#define SCENARIO 0`
Other SCENARIO values demonstrate successful operation.

### All relevant compiler information
Tested with Visual Studio 15.8.4 in Windows 10 and with GCC 6.4.0 in Ubuntu 16.04

### Code
The following code is based on Boost Beast synchronous WebSocket client sample, https://www.boost.org/doc/libs/1_68_0/libs/beast/example/websocket/client/sync-ssl/websocket_client_sync_ssl.cpp, with two differences:
- the string to send to an echo server is hard-coded and controlled by SCENARIO define
- CA certificates aren't loaded and `ws.next_layer().set_verify_mode(ssl::verify_none)` is executed instead.

```
// Show unexpected websocket::error::bad_frame_payload when receiving test_string
// from a WS echo server.
// Different values of SCENARIO lead to:
//   0 - complete string => ERROR
//   1 - only Bengali characters => success
//   2 - all, except for Bengali characters => success
// Just combining the Unicode strings, without enclosing ASCII (in raw string
// literals below) is not enough to break UTF8 check. Only the complete string,
// SCENARIO == 0, causes the error.
//
// Sending the complete string to an echo server from a web page
//  https://websocket.org/echo.html
// works fine, i.e. browser doesn't consider this Unicode string invalid
#define SCENARIO 0

#include <boost/beast/core.hpp>
#include <boost/beast/websocket.hpp>
#include <boost/beast/websocket/ssl.hpp>
#include <boost/asio/connect.hpp>
#include <boost/asio/ip/tcp.hpp>
#include <boost/asio/ssl/stream.hpp>
#include <cstdlib>
#include <iostream>
#include <string>

using tcp = boost::asio::ip::tcp;               // from <boost/asio/ip/tcp.hpp>
namespace ssl = boost::asio::ssl;               // from <boost/asio/ssl.hpp>
namespace websocket = boost::beast::websocket;  // from <boost/beast/websocket.hpp>

// Sends a WebSocket message and prints the response
int main(int, char**)
{
  // Characters copied from http://www.ltg.ed.ac.uk/~richard/unicode-sample.html
  // then converted to HEX at https://r12a.github.io/app-conversion/
  static const std::string text_japanese =
[...]

PR #1249, opened 2018-09-18
Change filler value for temp buffer to pass valid() check

bsergeev:
This is fix for issue https://github.com/boostorg/beast/issues/1245.

The problem happened when read_some buffer ends in the middle of 3-byte symbol (e.g. Bengali). In my example, the first `read_some()` call read 1532 characters, which included the 1st character of the 3-byte sequence, `0xE0`, as the last byte read.  Then, `fail_fast()` was called and filled the last 3 bytes of  `utf8_checker_t::cp_` with `0x81` values, then called `valid()`. Since `{0xE0, 0x81, 0x81, 0x81}` sequence doesn't pass the _overlong_ check for 3-byte values
```
(p[0] == 0xe0 && (p[1] & 0xe0) == 0x80) // overlong
```
`valid()` returned false and an error was thrown.

Changing the filler values from `0x81` to `0xA0` allows all checks to pass, for 2-, 3-, and probably 4-byte code points (I haven't tested the 4-byte ones, but believe they'll work too).

### DIFF
--- a/include/boost/beast/websocket/detail/utf8_checker.hpp
+++ b/include/boost/beast/websocket/detail/utf8_checker.hpp
@@ -153,13 +153,13 @@ write(std::uint8_t const* in, std::size_t size)
                 BOOST_ASSERT(false);
                 BOOST_FALLTHROUGH;
             case 1:
-                cp_[1] = 0x81;
+                cp_[1] = 0xA0;
                 BOOST_FALLTHROUGH;
             case 2:
-                cp_[2] = 0x81;
+                cp_[2] = 0xA0;
                 BOOST_FALLTHROUGH;
             case 3:
-                cp_[3] = 0x81;
+                cp_[3] = 0xA0;
                 break;
             }
             std::uint8_t const* p = cp_;
--- a/include/boost/beast/websocket/detail/utf8_checker.hpp
+++ b/include/boost/beast/websocket/detail/utf8_checker.hpp
@@ -112,7 +112,7 @@ write(std::uint8_t const* in, std::size_t size)
             if((p[0] & 0xe0) == 0xc0)
             {
                 if( (p[1] & 0xc0) != 0x80 ||
-                    (p[0] & 0xfe) == 0xc0)  // overlong
+                    (p[0] & 0x1e) == 0)  // overlong
                     return false;
                 p += 2;
                 return true;
@@ -121,8 +121,8 @@ write(std::uint8_t const* in, std::size_t size)
             {
                 if(    (p[1] & 0xc0) != 0x80
                     || (p[2] & 0xc0) != 0x80
-                    || (p[0] == 0xe0 && (p[1] & 0xe0) == 0x80) // overlong
-                    || (p[0] == 0xed && (p[1] & 0xe0) == 0xa0) // surrogate
+                    || (p[0] == 0xe0 && (p[1] & 0x20) == 0) // overlong
+                    || (p[0] == 0xed && (p[1] & 0x3C) >= 0x3C) // surrogate
                     //|| (p[0] == 0xef && p[1] == 0xbf && (p[2] & 0xfe) == 0xbe) // U+FFFE or U+FFFF
                     )
                     return false;
@@ -134,7 +134,7 @@ write(std::uint8_t const* in, std::size_t size)
                 if(    (p[1] & 0xc0) != 0x80
                     || (p[2] & 0xc0) != 0x80
                     || (p[3] & 0xc0) != 0x80
-                    || (p[0] == 0xf0 && (p[1] & 0xf0) == 0x80) // overlong
+                    || (p[0] == 0xf0 && (p[1] & 0x30) == 0) // overlong
                     || (p[0] == 0xf4 && p[1] > 0x8f) || p[0] > 0xf4 // > U+10FFFF
                     )
                     return false;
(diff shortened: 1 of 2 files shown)


vinniefalco:
Hah, so it was the `utf8_checker` all along

vinniefalco:
Do you know if this still passes the Autobahn|Testsuite?

bsergeev:
> Do you know if this still passes the Autobahn|Testsuite?
I don't know how to run them... You might have mentioned how, but I haven't tried that yet.
Doesn't your CD/CD run these tests?

I've just realized that the problem only happens if the message is pretty large, so that the first `read_some()` doesn't read it all, and the partial read buffer ends after the 1st byte of 3-byte code point, while ending after the 2nd byte works fine.

vinniefalco:
Okay I ran the Autobahn|Testsuite UTF8 test cases, and they almost all passed. There were two failures. I am linking the successful runs from the previous version of Beast:

https://vinniefalco.github.io/BeastAssets/reports/autobahn/index.html#case_desc_6_4_2
https://vinniefalco.github.io/BeastAssets/reports/autobahn/index.html#case_desc_6_4_4

These two cases (6.4.2 and 6.4.4) fail with the change to the fill character.  This is the diagnostic output for the failed test cases:

6.4.2 Case Outcome
Actual events differ from any expected.
Expected:
{'NON-STRICT': [('timeout', 'A'), ('timeout', 'B')], 'OK': [('timeout', 'A')]}
Observed:
[]

6.4.4 Case Outcome
Actual events differ from any expected.
Expected:
{'NON-STRICT': [('timeout', 'A'), ('timeout', 'B')], 'OK': [('timeout', 'A')]}
Observed:
[]

I am attaching the HTML for the two failed cases inside a ZIP archive:

[testcases.zip](https://github.com/boostorg/beast/files/2394912/testcases.zip)

bsergeev:
Looking into the Autobahn failures...

Travis also reported build failures, but its build output is hard to interpret (I've never dealt with Travis).
Will take care of the two failing Autobahn tests, then will see what Travis is whining about.

vinniefalco:
> Travis also reported build failures

Those are not your fault, I wouldn't worry about it.

bsergeev:
I've added 3 tests to _utf8_checker.cpp_: one to reproduce the problem I've stumbled upon and two for Autobahn tests I've broken (they both failed on 4-byte sequences that I didn't test).

While running _utf8_checker.cpp_, I've noticed some checks that don't make sense to me... For example, why is 224 (0xE0), which is the 1st character in 3-byte sequence, expected to fail:
https://github.com/boostorg/beast/blob/develop/test/beast/websocket/utf8_checker.cpp#L64
Shouldn't the beginning of a valid sequence be considered valid?

vinniefalco:
> Shouldn't the beginning of a valid sequence be considered valid?

It is possible that the utf8_checker.cpp tests are wrong. This thing was a pain to write and debug.

bsergeev:
Ok, so, Autobahn tests are the ultimate truth?
How can I run them?

bsergeev:
I'm ready to run Autobahn tests, "websocket-server-fast" compiled and running.

But you haven't mentioned what `wtest` is:
> wstest -m fuzzingclient

Is it the one described in Autobahn|Testsuite README (reading it now)?
https://github.com/crossbario/autobahn-testsuite 
It seems to be a Python script, right?
Can I run it without their Docker image (I don't have Docker installed on my personal laptop)?

include/boost/beast/websocket/detail/utf8_checker.hpp
```diff
@@ -121,8 +121,8 @@ write(std::uint8_t const* in, std::size_t size)
             {
                 if(    (p[1] & 0xc0) != 0x80
                     || (p[2] & 0xc0) != 0x80
-                    || (p[0] == 0xe0 && (p[1] & 0xe0) == 0x80) // overlong
-                    || (p[0] == 0xed && (p[1] & 0xe0) == 0xa0) // surrogate
+                    || (p[0] == 0xe0 && (p[1] & 0x20) == 0) // overlong
+                    || (p[0] == 0xed && (p[1] & 0x3C) >= 0x3C) // surrogate
```
bsergeev: The old surrogate check was wrong...
The check for UTF-16 surrogate pairs should be to verify that the values after 0xED do not fall in the range 0x800...0xFFF. Since, in the second byte of 3-byte sequence, 0x8 is shifted left by 2 bits, the mask becomes 0x3C.

bsergeev:
Attempted to install Autobahn|Testsuite (with Python 2, as they warned that 3 won't work) with
```
pip install autobahntestsuite
```
but installation failed with the following error:
> error: Microsoft Visual C++ 9.0 is required

Could you, please, run Autobahn tests for me?

vinniefalco:
Well these tests you wrote look very nice! I'm glad you understand what is going on with the code points, because I sure don't. Do you have a link to a good reference that clearly explains the process for determining if a code point is valid? If yes then you could include it as a comment in the source file.

bsergeev:
**All Autobahn 6.* tests are succeeding!**

I had to edit a couple of places in the unit test, as the first byte `0xF0` alone is not a failure, as it can start valid 4-byte sequences. It can become a problem, only if the following byte has zeros in last 6 bits, leading to overlong. But this condition is checked separately, so `0xF0` should not be deemed an invalid byte.

I've worked with UTF-8 before, so, it wasn't totally new to me. To refresh my memory, I read only Wikipedia, https://en.wikipedia.org/wiki/UTF-8, which has sufficient information, but perhaps not well structured.

vinniefalco:
Actually the Travis failures are due to a compilation error:
https://travis-ci.org/boostorg/beast/jobs/430871487#L1431

vinniefalco:
Great job on this Boris, and thank you very much for taking care of it!

bsergeev:
My pleasure! Thank you for the great library!

Are you in coming to Registration Reception in 20 minutes?

vinniefalco:
I'm coming to the Speaker's dinner on Thursday night and then giving my talk Friday afternoon :)

bsergeev:
See you then!

PR #1255, opened 2018-09-23
beast: misc-typos

luzpaz:
Found via `codespell -q 3 -L uint,te,tim --skip="./test/extern/zlib*,./example/http/client/crawl*"`


vinniefalco:
Good finds! One little detail though, the files http_parser.c and http_parser.h should be left alone since they are external code and it is important that the source files are exactly the same as they are in the foreign repository, even with the spelling errors.

luzpaz:
@vinniefalco oops...tried to avoid changing any 3rd party code. Reverted in 1e96ee0

PR #1267, opened 2018-10-14
Fix parse_hex algorithm

compmaniak:
Current `parse_hex` implementation doesn't work at the extreme case when input is "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff...". The function consumes all of such chars but returns maximum value for the specified type. This happens because of bounds checking via integer overflow.

### DIFF
--- a/include/boost/beast/http/detail/basic_parser.hpp
+++ b/include/boost/beast/http/detail/basic_parser.hpp
@@ -35,6 +35,12 @@ struct basic_parser_base
     //
     static std::size_t constexpr max_obs_fold = 4096;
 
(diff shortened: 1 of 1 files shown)


vinniefalco:
Good find! I will add a test to confirm. Can you please give me more details? What is the exact string, is it `"fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"`? How many `f` are in there? And what is the "specified type" you are using?

vinniefalco:
I understand the problem now. Good work

PR #1271, opened 2018-10-19
Update the build guide with cxxstd

Kojoley:


vinniefalco:
Hmm...this looks like an improvement!!

PR #1281, opened 2018-10-30
Fixes issue produced by exchanges that send non-printable characters …

konradstrachan:
…mid-way through the header (for example, as part of the cookie)

### DIFF
--- a/include/boost/beast/http/detail/basic_parser.hpp
+++ b/include/boost/beast/http/detail/basic_parser.hpp
@@ -267,42 +267,37 @@ struct basic_parser_base
                 return p;
             }
             if(BOOST_UNLIKELY(! is_print(*p)))
-                if((BOOST_LIKELY(static_cast<
-                        unsigned char>(*p) < '\040') &&
-                    BOOST_LIKELY(*p != '\011')) ||
-                    BOOST_UNLIKELY(*p == '\177'))
-                    goto found_control;
-        }
-    found_control:
-        if(BOOST_LIKELY(*p == '\r'))
-        {
-            if(++p >= last)
             {
-                ec = error::need_more;
-                return last;
+                if(BOOST_LIKELY(*p == '\r'))
+                {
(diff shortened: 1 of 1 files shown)


vinniefalco:
I appreciate the contribution, but this would make the parser accept invalid HTTP messages. Expecting a library to account for all ways in the world that servers can send malformed data is not only unrealistic, but would negatively impact the security of the library (see:  https://vinniefalco.github.io/BeastAssets/Beast%20-%20Hybrid%20Application%20Assessment%202017%20-%20Assessment%20Report%20-%2020171114.pdf).

This issue has been raised before, so I am thinking that a better way forward might be to add a hook to the parser's derived class which allows it to optionally rewrite each line of the header. Then you can make a parser instance that handles the specific problem you are trying to solve, and it won't affect any downstream Beast users.

konradstrachan:
> I appreciate the contribution, but this would make the parser accept invalid HTTP messages. Expecting a library to account for all ways in the world that servers can send malformed data is not only unrealistic, but would negatively impact the security of the library (see: https://vinniefalco.github.io/BeastAssets/Beast%20-%20Hybrid%20Application%20Assessment%202017%20-%20Assessment%20Report%20-%2020171114.pdf).
> 
> This issue has been raised before, so I am thinking that a better way forward might be to add a hook to the parser's derived class which allows it to optionally rewrite each line of the header. Then you can make a parser instance that handles the specific problem you are trying to solve, and it won't affect any downstream Beast users.

Hi Vinnie, thanks for the quick response. As a temporary measure, how about we #if 0 it as per your work-around for '\n' terminator? Failing that, do you have an idea for when such a parser change might be ready?

vinniefalco:
Heads up, I haven't forgotten about this by any means, but currently I am in the middle of big revamping of websocket for 1.70.

vinniefalco:
Can't do this, non-compliant parsing is out of scope.

PR #1293, opened 2018-11-09
Compile ASIO into a static lib

djarek:


tools/build-and-test.sh
```diff
@@ -17,7 +17,9 @@ if [[ $TRAVIS_BRANCH == "master" || $TRAVIS_BRANCH == "develop" ]]; then
     MAIN_BRANCH="1"
 fi
 
-if [[ "${TRAVIS}" == "true" ]]; then
+if [[ "${BEAST_RETRY}" == "true" ]]; then
+  JOBS=1
+elif [[ "${TRAVIS}" == "true" ]]; then
```
vinniefalco: So... on the first try you use 2 jobs, on subsequent tries 1 job? That's rather...brilliant :)

djarek: Yeah, couldn't figure a better way.

PR #1303, opened 2018-11-14
Properly cancel the timer in timeout_service

djarek:
The lambda was never invoked, because at that point the Executor was already stopped.

### DIFF
--- a/include/boost/beast/experimental/core/detail/impl/timeout_service.ipp
+++ b/include/boost/beast/experimental/core/detail/impl/timeout_service.ipp
@@ -165,13 +165,9 @@ void
 timeout_service::
 shutdown() noexcept
 {
-    boost::asio::post(
-        boost::asio::bind_executor(
-            strand_,
-            [this]()
-            {
-                timer_.cancel();
-            }));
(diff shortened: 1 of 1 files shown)


include/boost/beast/experimental/core/detail/impl/timeout_service.ipp
```diff
@@ -165,13 +165,9 @@ void
...
-    boost::asio::post(
-        boost::asio::bind_executor(
-            strand_,
-            [this]()
-            {
-                timer_.cancel();
-            }));
+    // When shutdown() is called the ExecutionContext must already be in a
+    // stopped state, so it's safe to just cancel() without synchronization.
+    timer_.cancel();
```
vinniefalco: LOL... whoops :)

vinniefalco:
So does this mean that there is no race?

djarek:
I believe there's still a data race on the timer here: https://github.com/boostorg/beast/pull/1303/files#diff-793414e6f06bada9618b29c958450224L69

vinniefalco:
This needs proper analysis for the race

vinniefalco:
I think it is fixed now

PR #1310, opened 2018-11-21
Remove redundant use of bind_handler

djarek:
bind_handler with 0 arguments is equivalent to passing directly to post().


include/boost/beast/http/impl/write.ipp
```diff
@@ -293,7 +293,7 @@ operator()(
             BOOST_ASIO_CORO_YIELD
             boost::asio::post(
                 s_.get_executor(),
-                bind_handler(std::move(*this)));
+                std::move(*this));
```
vinniefalco: I don't know why I did this, looks like a blind copy/paste/edit.

PR #1312, opened 2018-11-21
Lean tuple

djarek:
Intended to replace `std::tuple` for simple cases where a variadic list of data members is needed.

### DIFF
--- a/include/boost/beast/core/detail/lean_tuple.hpp
+++ b/include/boost/beast/core/detail/lean_tuple.hpp
@@ -1,0 +1,86 @@
+//
+// Copyright (c) 2018 Damian Jarek (damian dot jarek93 at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+
+#ifndef BOOST_BEAST_DETAIL_LEAN_TUPLE_HPP
+#define BOOST_BEAST_DETAIL_LEAN_TUPLE_HPP
+
+#include <boost/mp11/integer_sequence.hpp>
+#include <boost/mp11/algorithm.hpp>
+
+#include <boost/type_traits/remove_cv.hpp>
+#include <boost/type_traits/copy_cv.hpp>
+
+namespace boost
+{
+namespace beast
+{
+namespace detail
+{
+
+template<std::size_t I, typename T>
+struct lean_tuple_element
+{
+    T t_;
+};
+
+template<typename... Ts>
+struct lean_tuple_impl;
+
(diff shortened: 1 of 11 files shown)


include/boost/beast/core/detail/lean_tuple.hpp
```diff
@@ -0,0 +1,156 @@
...
+// Official repository: https://github.com/boostorg/beast
+//
+
+
+#ifndef BOOST_BEAST_DETAIL_LEAN_TUPLE_HPP
+#define BOOST_BEAST_DETAIL_LEAN_TUPLE_HPP
+
+#include <boost/mp11/integer_sequence.hpp>
+
+#if __cplusplus > 201703L
```
glenfe: This is wrong. You cannot assume this for *any* C++ language feature or C++ library feature. This is why we have Boost.Config. Because C++ implementations that have  `__cpluplus`  greater than 201703L (e.g. in C++2a standards mode) still do not have this feature implemented.

test/beast/core/detail/lean_tuple.cpp
```diff
@@ -0,0 +1,57 @@
+//
+// Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
```
vinniefalco: Damian?

test/beast/core/detail/lean_tuple.cpp
```diff
@@ -0,0 +1,57 @@
+//
+// Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
```
vinniefalco: Your name?

include/boost/beast/core/detail/lean_tuple.hpp
```diff
@@ -0,0 +1,86 @@
...
+// Copyright (c) 2018 Damian Jarek (damian dot jarek93 at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+
+#ifndef BOOST_BEAST_DETAIL_LEAN_TUPLE_HPP
```
glenfe: Shouldn't this be `BOOST_BEAST_CORE_DETAIL_LEAN_TUPLE_HPP`?

vinniefalco: I pretend like the "core" directory doesn't actually exist. Because it shouldn't exist. The actual include file should be `<beast/detail/lean_tuple.hpp>`. Unfortunately if I were to do that then the Visual Studio project would have a really long, annoying list of header files that I could not collapse and still see other virtual folders.

PR #1321, opened 2018-11-24
Tidy test::stream

### LINKED ISSUE #1308: Reduce dependencies of test::stream
`test::stream` has a dependency on a small part of websockets and also transitively pulls in `tcp.hpp` from ASIO. I believe this can be resolved with forward declarations.

djarek:
test::stream will no longer include asio::tcp::ip transitively.
Resolves #1308


vinniefalco:
I'm not entirely sure this is right. Doesn't the compiler have to see the primary function template first, before a specialization?

djarek:
I don't believe these are specializations, they are regular overloads. https://en.cppreference.com/w/cpp/language/template_specialization

PR #1322, opened 2018-11-25
Fix test::stream async ops for custom result types

djarek:
Async initiating functions must return init.result.get(). Returning the result of post caused a compilation failure for custom completion tokens.

### DIFF
--- a/include/boost/beast/_experimental/test/impl/stream.ipp
+++ b/include/boost/beast/_experimental/test/impl/stream.ipp
@@ -223,7 +223,7 @@ read_some(MutableBufferSequence const& buffers,
         });
     std::size_t bytes_transferred;
     if(in_->b.size() > 0)
-    {   
+    {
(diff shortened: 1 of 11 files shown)


include/boost/beast/_experimental/test/impl/stream.ipp
```diff
@@ -257,17 +257,18 @@ async_read_some(
...
-        if(in_->fc->fail(ec))
-            return boost::asio::post(
-                in_->ioc.get_executor(),
-                bind_front_handler(
-                    std::move(init.completion_handler),
-                    ec,
-                    0));
+        boost::asio::post(
+            in_->ioc.get_executor(),
+            bind_front_handler(
```
vinniefalco: needs qualification, e.g. `beast::bind_front_handler`

djarek: Done.

PR #1327, opened 2018-11-27
Fix use of destroyed buffer in test

djarek:
b was accessed by a move constructor of the unexecuted websocket read operation.


vinniefalco:
Needs a colon on the first line (since there is continuation text). I will amend the commit message when I merge it.

PR #1328, opened 2018-11-27
ssl_stream enhancements

octobanana:
* Use std::make_unique instead of new in construction of std::unique_ptr.
* Explicitly make the object non-copyable by deleting the copy constructor and copy assignment operator.
* Remove the non-owning pointer to the ssl::context private member variable, due to it being unused after constructing the underlying stream.
* Fix a few style inconsistencies.
* Use the default move constructor and default move assignment operator.

### DIFF
--- a/include/boost/beast/_experimental/core/ssl_stream.hpp
+++ b/include/boost/beast/_experimental/core/ssl_stream.hpp
@@ -108,19 +108,6 @@ public:
     {
     }
 
-    /// Move Constructor
-    ssl_stream(ssl_stream&& other)
-        : p_(std::move(other.p_))
-    {
-    }
-
-    /// Move Assignment
-    ssl_stream& operator=(ssl_stream&& other)
-    {
-        p_ = std::move(other.p_);
-        return *this;
-    }
-
     /** Get the executor associated with the object.
 
         This function may be used to obtain the executor object that the stream
--- a/include/boost/beast/_experimental/core/ssl_stream.hpp
+++ b/include/boost/beast/_experimental/core/ssl_stream.hpp
@@ -72,7 +72,6 @@ class ssl_stream
     using stream_type = boost::beast::flat_stream<ssl_stream_type>;
 
     std::unique_ptr<stream_type> p_;
-    boost::asio::ssl::context* ctx_;
 
 public:
     /// The native handle type of the SSL stream.
@@ -106,14 +105,12 @@ public:
         boost::asio::ssl::context& ctx)
         : p_(new stream_type{
             std::forward<Arg>(arg), ctx})
-        , ctx_(&ctx)
     {
     }
 
     /// Move Constructor
     ssl_stream(ssl_stream&& other)
         : p_(std::move(other.p_))
-        , ctx_(other.ctx_)
     {
     }
 
@@ -121,7 +118,6 @@ public:
     ssl_stream& operator=(ssl_stream&& other)
     {
         p_ = std::move(other.p_);
-        ctx_ = other.ctx_;
         return *this;
     }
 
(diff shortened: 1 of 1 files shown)


octobanana:
Thoughts?

include/boost/beast/_experimental/core/ssl_stream.hpp
```diff
@@ -104,8 +104,8 @@ class ssl_stream
     ssl_stream(
         Arg&& arg,
         boost::asio::ssl::context& ctx)
-        : p_(new stream_type{
-            std::forward<Arg>(arg), ctx})
+        : p_(std::make_unique<stream_type>(
```
vinniefalco: This is available in C++14 and later, while Beast only requires C++11

include/boost/beast/_experimental/core/ssl_stream.hpp
```diff
@@ -125,6 +131,9 @@ class ssl_stream
         return *this;
     }
 
+    /// Default Deconstructor
+    ~ssl_stream() = default;
```
vinniefalco: hmm...I'm not sure that this actually changes anything, the copy members were already implicitly disabled because they are not available in the underlying stream. And the destructor was already implicitly defaulted. What was the intent here?

vinniefalco:
It looks like the only meaningful changes are the removal of the unused pointer (great!) and fixing the whitespace...

include/boost/beast/_experimental/core/ssl_stream.hpp
octobanana: Oops, correct, I forgot it was a 14 and up feature.

include/boost/beast/_experimental/core/ssl_stream.hpp
octobanana: To be more explicit, but you're right, it's redundant since they are not available in the underlying stream.

vinniefalco: The less code the better, especially because defaulted special members can sometimes confuse the coverage reports.

octobanana: So do you prefer having the move ctor and move assignment to not use the '= default'?

vinniefalco: I think it is better to just leave them unspecified and let the compiler implicitly delete them (or generate them, whatever the case may be). The rules for what the compiler does in each case are in this presentation from @HowardHinnant https://stackoverflow.com/questions/24342941/what-are-the-rules-for-automatic-generation-of-move-operations

octobanana: That diagram is helpful, thanks.

include/boost/beast/_experimental/core/ssl_stream.hpp
```diff
@@ -106,24 +105,14 @@ class ssl_stream
         boost::asio::ssl::context& ctx)
         : p_(new stream_type{
             std::forward<Arg>(arg), ctx})
-        , ctx_(&ctx)
     {
     }
 
     /// Move Constructor
-    ssl_stream(ssl_stream&& other)
```
octobanana: Since the move constructor and move assignment are implicitly generated, this could be removed.

include/boost/beast/_experimental/core/ssl_stream.hpp
vinniefalco: I tend to forget so don't be surprised if you find several places in beast where I have unnecessary explicitly defaulted special members.

include/boost/beast/_experimental/core/ssl_stream.hpp
djarek: Note that this changes the behavior: `.get_executor()` on a moved-from stream will now be UB (null-deref). Perhaps the behavior on move of this stream should be documented? (it differs from how I/O objects in ASIO behave).

djarek: Ehh, I just noticed `.get_executor()` already behaves like that, so no change. Sorry for the noise!

PR #1331, opened 2018-11-30
Version 194

vinniefalco:
* http::async_read returns the right byte count on error
* Add net namespace alias
* Don't use-after-free in test
* Tidy up ssl_stream (experimental)
* Dynamic buffer improvements

### DIFF
--- a/include/boost/beast/core/flat_buffer.hpp
+++ b/include/boost/beast/core/flat_buffer.hpp
@@ -20,17 +20,29 @@
 namespace boost {
 namespace beast {
 
-/** A linear dynamic buffer.
+/** A dynamic buffer providing buffer sequences of length one.
+
+    A dynamic buffer encapsulates memory storage that may be
+    automatically resized as required, where the memory is
+    divided into two regions: readable bytes followed by
+    writable bytes. These memory regions are internal to
+    the dynamic buffer, but direct access to the elements
+    is provided to permit them to be efficiently used with
+    I/O operations.
 
     Objects of this type meet the requirements of @b DynamicBuffer
-    and offer additional invariants:
-    
-    @li Buffer sequences returned by @ref data and @ref prepare
-    will always be of length one.
+    and have the following additional properties:
+
+    @li A mutable buffer sequence representing the readable
+    bytes is returned by @ref data when `this` is non-const.
 
     @li A configurable maximum buffer size may be set upon
     construction. Attempts to exceed the buffer size will throw
     `std::length_error`.
+
+    @li Buffer sequences representing the readable and writable
+    bytes, returned by @ref data and @ref prepare, will have
+    length one.
 
     Upon construction, a maximum size for the buffer may be
     specified. If this limit is exceeded, the `std::length_error`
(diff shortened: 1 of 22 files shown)


include/boost/beast/core/multi_buffer.hpp
```diff
@@ -82,82 +110,103 @@ class basic_multi_buffer
...
 
-        After the move, `*this` will have an empty output sequence.
+        Buffer sequences previously obtained using @ref data
+        or @ref prepare are not invalidated after the move.
 
         @param other The object to move from. After the move,
-        The object's state will be as if constructed using
-        its current allocator and limit.
+        the moved-from object's state will be as if default
+        constructed using its current allocator and limit.
```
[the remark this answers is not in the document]
vinniefalco: @glenfe ?

include/boost/beast/core/flat_buffer.hpp
```diff
@@ -206,131 +245,159 @@ class basic_flat_buffer
...
 
-        @note All previous buffers sequences obtained from
-        calls to @ref data or @ref prepare are invalidated.
+        @param n The number of bytes to append. If this number
+        is greater than the number of writable bytes, all
+        writable bytes are appended.
     */
     void
-    commit(std::size_t n)
+    commit(std::size_t n) noexcept
```
glenfe: Why is this noexcept?

include/boost/beast/core/impl/flat_buffer.ipp
```diff
@@ -284,7 +321,7 @@ prepare(std::size_t n) ->
 template<class Allocator>
 void
 basic_flat_buffer<Allocator>::
-consume(std::size_t n)
+consume(std::size_t n) noexcept
```
glenfe: Also a questionable that all consumes are noexcept

include/boost/beast/core/multi_buffer.hpp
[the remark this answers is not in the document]
glenfe: Since C++20 moving an allocator of the same type should have the same postconditions as copying it.

include/boost/beast/core/static_buffer.hpp
```diff
@@ -46,18 +60,98 @@ class static_buffer_base
...
+        // workaround for buffers_iterator bug
+        using value_type =
+            boost::asio::mutable_buffer;
+
+        mutable_buffer_pair() = default;
+        mutable_buffer_pair(
+            mutable_buffer_pair const&) = default;
+
+        boost::asio::mutable_buffer&
+        operator[](int i) noexcept
```
glenfe: And these are also questionably noexcept.

include/boost/beast/core/flat_buffer.hpp
vinniefalco: it only performs a possible deallocation

vinniefalco: "moving" bytes from the writable to the readable area merely adjusts a few pointers

include/boost/beast/core/impl/flat_buffer.ipp
vinniefalco: It only performs a possible deallocation. Can deallocate throw?

include/boost/beast/core/static_buffer.hpp
vinniefalco: Why wouldn't it be `noexcept`?

PR #1336, opened 2018-12-03
Set SO_REUSEADDR in echo_op example

### LINKED ISSUE #1335: echo-op example does not set SO_REUSEADDR
The echo-op example does not set the SO_REUSEADDR socket option. It would match the other examples well if it also set it.

octobanana:
fix #1335


vinniefalco:
This is in https://github.com/boostorg/beast/pull/1334 now, waiting on CI

vinniefalco:
Hmm.....if we are going to make a bunch of changes to echo-op I think it would be best to put it in a single commit, and we can give it a couple of days to settle would that be OK?

octobanana:
Sure, I can close the echo-op related pull requests and submit a single one with all the changes.

vinniefalco:
I think that would be best

PR #1337, opened 2018-12-03
Remove unused type_traits in detail

djarek:


vinniefalco:
I wrong this for something...what was it??

PR #1340, opened 2018-12-03
Use built-in union in handler_ptr

djarek:
Replace aligned_union in handler_ptr with the built-in one to reduce the number of template instantiations and remove the reinterpret_casts that were required.

### DIFF
--- a/include/boost/beast/core/impl/handler_ptr.ipp
+++ b/include/boost/beast/core/impl/handler_ptr.ipp
@@ -41,7 +41,7 @@ handler_ptr<T, Handler>::
     if(t_)
     {
         clear();
-        handler().~Handler();
+        h_.~Handler();
     }
 }
 
@@ -52,8 +52,9 @@ handler_ptr(handler_ptr&& other)
 {
     if(other.t_)
     {
-        new(&h_) Handler(std::move(other.handler()));
-        other.handler().~Handler();
+        ::new(static_cast<void*>(std::addressof(h_)))
+            Handler(std::move(other.h_));
+        other.h_.~Handler();
         other.t_ = nullptr;
     }
 }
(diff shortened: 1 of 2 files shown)


include/boost/beast/core/impl/handler_ptr.ipp
```diff
@@ -52,8 +52,8 @@ handler_ptr(handler_ptr&& other)
 {
     if(other.t_)
     {
-        new(&h_) Handler(std::move(other.handler()));
-        other.handler().~Handler();
+        new(&h_) Handler(std::move(other.h_));
```
glenfe: No `addressof(h)`? Before you didn't have to worry about `T` having overloaded `&`, but now you do.

vinniefalco:
Reducing instantiations **and** removing `reinterpret_cast`? Great!!

include/boost/beast/core/impl/handler_ptr.ipp
djarek: Damn, forgot about that. Done.

vinniefalco: Also fully-qualify `::new`?

include/boost/beast/core/impl/handler_ptr.ipp
```diff
@@ -52,8 +52,9 @@ handler_ptr(handler_ptr&& other)
 {
     if(other.t_)
     {
-        new(&h_) Handler(std::move(other.handler()));
-        other.handler().~Handler();
+        ::new((void*)std::addressof(h_))
```
vinniefalco: I would prefer to leave out C-style casts completely, because it is almost impossible to search for them

include/boost/beast/core/impl/handler_ptr.ipp
djarek: Done.

include/boost/beast/core/impl/handler_ptr.ipp
djarek: Done.

vinniefalco:
merged

Issue #1275, opened 2018-10-24
Add sync echo algorithm to example


vinniefalco:
This would be nice for symmetry

vinniefalco:
Not needed. Sync APIs are not for production

PR #1347, opened 2018-12-03
echo_op example

octobanana:
Set the `reuse_address` socket option to match the style of the other examples.

---

Add the ability to set the listening address and port through command line arguments. Print out usage and return error exit status on invalid or missing arguments.

---

Currently, the __echo-op__ example uses the carriage return `\r` delimiter for `async_read_until`. Running the server and sending a message to it with curl or netcat:
```sh
printf 'Hello, World!\r\n' | nc 127.0.0.1 8080
```
leads to some potentially confusing behaviour. The `async_read_until` function reads up to and including the first `\r`. When the message is received by the client and printed to stdout, the trailing `\r` ends up moving the cursor back to the start of the line, resulting in the prompt overwriting the message. It appears as if nothing has happened.

If `async_read_until` uses the newline `\n` as the delimiter, the echoed output will not be overwritten, regardless of whether the line ends in `\n` or `\r\n`.

---

Had a go at adding a synchronous echo op function #1275.

### DIFF
--- a/example/echo-op/echo_op.cpp
+++ b/example/echo-op/echo_op.cpp
@@ -301,6 +301,7 @@ int main(int, char** argv)
     boost::asio::ip::tcp::acceptor acceptor{ioc};
     endpoint_type ep{boost::asio::ip::make_address("0.0.0.0"), 0};
     acceptor.open(ep.protocol());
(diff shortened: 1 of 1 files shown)


vinniefalco:
The sync echo-op needs exposition and documentation...

Issue #1285, opened 2018-11-01
Add multiple io_context example


vinniefalco:
We could use a third advanced server that has multiple threads, one io_context per thread.

vinniefalco:
This can be done with a custom executor now

benstadin:
I just found this after trying your example at [1] with some quick and dirty hacks to get it compiling with the current version of Beast. However, it crashes after a few requests. My guess is it's related to reusing the context, but it's not clear to me how to reset or guard the context correctly. 

I've created a gist here using the advanced server example with hacks to the context pool:
https://gist.github.com/benstadin/4898c4a57bc55cc4ba2d483104335890

The crash I see after a few requests is like:
`Assertion failed: (value > 0 && value < 100), function version, file /usr/local/include/boost/beast/http/message.hpp, line 342.`

I'm looking to improve performance with multiple CPU cores, and looking at cpu time spent in the profiler looks like this could help significantly. At the moment requests / second settle near 60k (not bad at all though), using 8 or 16 threads has only minor effect on my 8 core i9 CPU (like from 54k to 60k rps). 

[1] https://github.com/vinniefalco/beast/blob/multi-io/example/advanced/server-multi-io-context/advanced_server_multi_io_context.cpp

benstadin:
Edit: I've updated a gist. It doesn't crash anymore (issue was behind the monitor). 

But would you consider this approach a good design in general, or should this rather be done with a custom executor now? 

Some measurements show roughly a 50% improvement in requests/s when using multiple io contexts (running on OS X 10.15.2, 8-core Intel i9):

1) Using advanced server example (changed to return immediately a string "Test" to avoid disk io).

wrk -t8 -c100 -d10s -H 'Connection: keep-alive' --latency --timeout 60 http://127.0.0.1:8080
Running 10s test @ http://127.0.0.1:8080
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.52ms  247.35us   3.77ms   71.85%
    Req/Sec     7.88k   661.47     9.67k    62.62%
  Latency Distribution
     50%    1.47ms
     75%    1.63ms
     90%    1.90ms
     99%    2.24ms
  633900 requests in 10.10s, 68.31MB read
Requests/sec:  62759.35
Transfer/sec:      6.76MB

2) Using multiple io contexts

[...]

PR #1348, opened 2018-12-03
Add multiple io_context advanced server example

octobanana:
I had a go at #1285, an advanced server example that uses a single io_context per thread. It is based off the advanced server example. It selects the next io_context per socket in a round robin fashion. If you diff it to the advanced server example, you can see the changes that were made.

Is this kind of what you had in mind?

### DIFF
--- a/example/advanced/server-io/CMakeLists.txt
+++ b/example/advanced/server-io/CMakeLists.txt
@@ -1,0 +1,19 @@
+#
+# Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/boostorg/beast
+#
+
+GroupSources(include/boost/beast beast)
+GroupSources(example/advanced/server-io "/")
+
+add_executable (advanced-server-io
+    ${BOOST_BEAST_FILES}
+    Jamfile
+    advanced_server_io.cpp
+)
+
+set_property(TARGET advanced-server-io PROPERTY FOLDER "example-advanced-server")
--- a/example/advanced/server-io/advanced_server_io.cpp
+++ b/example/advanced/server-io/advanced_server_io.cpp
@@ -1,0 +1,914 @@
+//
+// Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+//------------------------------------------------------------------------------
+//
+// Example: Advanced server, one io_context per thread
+//
+//------------------------------------------------------------------------------
+
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
+#include <boost/beast/websocket.hpp>
+#include <boost/beast/version.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/signal_set.hpp>
+#include <boost/asio/steady_timer.hpp>
+#include <boost/make_unique.hpp>
+#include <boost/bind.hpp>
+#include <boost/config.hpp>
+#include <algorithm>
+#include <cstdlib>
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <thread>
+#include <vector>
+
+using tcp = boost::asio::ip::tcp;               // from <boost/asio/ip/tcp.hpp>
+namespace http = boost::beast::http;            // from <boost/beast/http.hpp>
+namespace websocket = boost::beast::websocket;  // from <boost/beast/websocket.hpp>
+
+class round_robin
+{
+    using io_ptr = std::shared_ptr<boost::asio::io_context>;
+    using io_work = boost::asio::executor_work_guard<
+        boost::asio::io_context::executor_type>;
+
+    struct ctx
+    {
+        ctx(io_ptr io, io_work work)
+            : io {io}
+            , work {work}
+        {
+        }
+
+        ctx(ctx const&) = delete;
+        ctx& operator=(ctx const&) = delete;
+        ctx(ctx&&) = default;
+        ctx& operator=(ctx&&) = default;
+        ~ctx() = default;
+
+        io_ptr io;
+        io_work work;
+    }; // struct ctx
+
+    std::vector<ctx> ctx_;
+    std::size_t next_ {0};
+
+public:
+
+    round_robin(std::size_t size = 1)
+    {
+        size = std::max<std::size_t>(1, size);
+
+        for (std::size_t i = 0; i < size; ++i)
+        {
+            io_ptr io {std::make_shared<boost::asio::io_context>()};
+            ctx_.emplace_back(io, boost::asio::make_work_guard(*io));
+        }
+    }
+
+    round_robin(round_robin const&) = delete;
+    round_robin& operator=(round_robin const&) = delete;
+    round_robin(round_robin&&) = default;
+    round_robin& operator=(round_robin&&) = default;
+    ~round_robin() = default;
+
+    void
+    run()
+    {
+        std::vector<std::thread> threads;
+        for (auto& e : ctx_)
+        {
+            threads.emplace_back(
+                boost::bind(&boost::asio::io_context::run, e.io));
+        }
+
+        for (auto& e : threads)
+        {
+            e.join();
+        }
+    }
+
+    void
+    stop()
+    {
+        for (auto& e : ctx_)
+        {
+            e.io->stop();
+        }
+    }
+
+    boost::asio::io_context&
+    get()
+    {
+        auto& io = *ctx_.at(next_).io;
+        if (++next_ == ctx_.size())
+        {
+            next_ = 0;
+        }
+        return io;
+    }
+}; // class round_robin
+
+// Return a reasonable mime type based on the extension of a file.
+boost::beast::string_view
+mime_type(boost::beast::string_view path)
+{
+    using boost::beast::iequals;
+    auto const ext = [&path]
+    {
+        auto const pos = path.rfind(".");
+        if(pos == boost::beast::string_view::npos)
+            return boost::beast::string_view{};
+        return path.substr(pos);
+    }();
+    if(iequals(ext, ".htm"))  return "text/html";
+    if(iequals(ext, ".html")) return "text/html";
+    if(iequals(ext, ".php"))  return "text/html";
+    if(iequals(ext, ".css"))  return "text/css";
+    if(iequals(ext, ".txt"))  return "text/plain";
+    if(iequals(ext, ".js"))   return "application/javascript";
+    if(iequals(ext, ".json")) return "application/json";
+    if(iequals(ext, ".xml"))  return "application/xml";
+    if(iequals(ext, ".swf"))  return "application/x-shockwave-flash";
+    if(iequals(ext, ".flv"))  return "video/x-flv";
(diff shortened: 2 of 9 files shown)


example/advanced/server-io/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+#
+# Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
```
vinniefalco: Should add your name

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+#include <thread>
+#include <vector>
+
+using tcp = boost::asio::ip::tcp;               // from <boost/asio/ip/tcp.hpp>
+namespace http = boost::beast::http;            // from <boost/beast/http.hpp>
+namespace websocket = boost::beast::websocket;  // from <boost/beast/websocket.hpp>
+
+class round_robin
+{
+    using io_ptr = std::shared_ptr<boost::asio::io_context>;
```
vinniefalco: I prefer to just repeat the type, it is more clear

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+        io_ptr io;
+        io_work work;
+    }; // struct ctx
+
+    std::vector<ctx> ctx_;
+    std::size_t next_ {0};
+
+public:
+
+    round_robin(std::size_t size = 1)
```
vinniefalco: Should be `explicit`

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+    get()
+    {
+        auto& io = *ctx_.at(next_).io;
+        if (++next_ == ctx_.size())
+        {
+            next_ = 0;
+        }
+        return io;
+    }
+}; // class round_robin
```
vinniefalco: I would put this class in its own header and rename it to `multi_io_context`. This is a candidate for the _experimental directory.

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+            // to return immediately, eventually destroying the
+            // `io_context` and all of the sockets in it.
+            io_pool.stop();
+        });
+
+    // Run the I/O services on the requested number of threads
+    io_pool.run();
+
+    // (If we get here, it means we got a SIGINT or SIGTERM)
+
```
vinniefalco: I think it would be nice to explicitly call `io_pool.join()` this way we can place a comment after that says "at this point, no more threads are running" or similar.

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+        return EXIT_FAILURE;
+    }
+    auto const address = boost::asio::ip::make_address(argv[1]);
+    auto const port = static_cast<unsigned short>(std::atoi(argv[2]));
+    auto const doc_root = std::make_shared<std::string>(argv[3]);
+    auto const threads = std::max<std::size_t>(1, std::atoi(argv[4]));
+
+    // The io_context is required for all I/O
+    // Use one io_context per thread round robin style
+    round_robin io_pool {threads};
```
vinniefalco: I think this would look better like this:
```
multi_io_context ioc{threads};
```

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+    stop()
+    {
+        for (auto& e : ctx_)
+        {
+            e.io->stop();
+        }
+    }
+
+    boost::asio::io_context&
+    get()
```
vinniefalco: Consider calling this function `next_context` to make it clear that 1. a choice is being made, and 2. the state of the object is changing

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+    ~round_robin() = default;
+
+    void
+    run()
+    {
+        std::vector<std::thread> threads;
+        for (auto& e : ctx_)
+        {
+            threads.emplace_back(
+                boost::bind(&boost::asio::io_context::run, e.io));
```
vinniefalco: I think we can use `std::bind`. Better yet, this should probably be a lambda since it will be more clear to readers.

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+#include <memory>
+#include <string>
+#include <thread>
+#include <vector>
+
+using tcp = boost::asio::ip::tcp;               // from <boost/asio/ip/tcp.hpp>
+namespace http = boost::beast::http;            // from <boost/beast/http.hpp>
+namespace websocket = boost::beast::websocket;  // from <boost/beast/websocket.hpp>
+
+class round_robin
```
vinniefalco: Something to consider for the future, not important now:
```
template<class SelectionPolicy = round_robin>
class multi_io_context;
```

example/advanced/server-io/advanced_server_io.cpp
vinniefalco: We shouldn't need `shared_ptr` here it should be possible to simply use `vector<io_context>`.

djarek: `vector` won't work because `io_context` is noncopyable.

example/advanced/server-io/advanced_server_io.cpp
djarek: In general you should avoid taking the address of member functions of classes you don't own. (Especially if they're in they're standard/"semistandard"(

example/advanced/server-io/advanced_server_io.cpp
vinniefalco: Ah yes you're right. `deque` works though.

octobanana: I'll try it out using a `deque`. Since it doesn't need to share ownership, a `vector<unique_ptr<io_context>>`  also works.

example/advanced/server-io/advanced_server_io.cpp
octobanana: Right, I'll add `explicit`

example/advanced/server-io/advanced_server_io.cpp
octobanana: Alright, I'll move it to its own header and rename it.

example/advanced/server-io/advanced_server_io.cpp
octobanana: That makes sense. I'll move the thread join calls from `run()` to it's own `join()` function.

example/advanced/server-io/advanced_server_io.cpp
octobanana: Sure

example/advanced/server-io/advanced_server_io.cpp
octobanana: Using `next_context()` should make it more clear in what it does.

example/advanced/server-io/advanced_server_io.cpp
octobanana: I'll change it to use a lambda. Noted on avoiding taking the address of member functions of non-owned objects.

example/advanced/server-io/advanced_server_io.cpp
octobanana: That would be neat. Having a `least_connected` selection policy would be interesting. Give each io_context a weight based on the current number of sockets it's being used with, then next_context would return the io_context with the lowest weight.

example/advanced/server-io/advanced_server_io.cpp
```diff
@@ -0,0 +1,914 @@
...
+
+public:
+
+    round_robin(std::size_t size = 1)
+    {
+        size = std::max<std::size_t>(1, size);
+
+        for (std::size_t i = 0; i < size; ++i)
+        {
+            io_ptr io {std::make_shared<boost::asio::io_context>()};
```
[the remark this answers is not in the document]
octobanana: Ah good idea!

octobanana:
Thanks for the review and tips! I'll have a go at the changes.

example/advanced/server-io/advanced_server_io.cpp
vinniefalco: That is ugly though, extra allocation and indirection per element.

example/advanced/server-io/advanced_server_io.cpp
vinniefalco: Hmm good idea! That requires counting the sockets but we could build some RAII counter thingy that you put in your session object. Here's another idea, in the io_context thread we call `run_one` instead of `run`, in a loop, and we can count I/O and compare how much time each thread is spending as a cheap proxy for utilization.

vinniefalco:
There needs to be a way for the caller to control the thread, or else how would they put a try/catch around the call to `run`? Or what if they want to do some fancy stuff? Or what if they don't even want to use `std::thread` but rather some other system API? We don't have to address this now but the constructor probably needs to accept a "factory" lambda which, when invoked, spawns the thread and invokes the appropriate io_context run or poll functions. Then the multi_io_context calls the factory for each requested io_context.

example/advanced/server-io/advanced_server_io.cpp
djarek: there's no need to resize, so I think `make_unique<io_context[]>(n)` is fine

vinniefalco: `make_unique` is not C++11

vinniefalco: How do you pass constructor arguments?

djarek: There's no point of having more than 1 thread per `io_context` if you have load balancing across multiple `io_context`s., so a hardcoded concurrency hint of 1 is fine.

example/advanced/server-io/advanced_server_io.cpp
djarek: Note that `run_one` may execute more than one CompletionHandler if the `io_context`'s Executor is wrapped with another Executor, e.g. `boost::asio::strand`.

example/advanced/server-io/advanced_server_io.cpp
octobanana: Would there ever be a scenario where having several threads per multiple `io_context`'s would be beneficial? Say 2 `io_context`'s and 4 `thread`'s, 2 per context?

vinniefalco: I don't think so. And if there is, then it is such a niche use-case that I don't think it is worth addressing in this particular class.

vinniefalco:
Is there an update?

octobanana:
There is! Sorry about the delay. I'll push the updated request tomorrow.

vinniefalco:
There's no rush. I am about to update all the examples to use `multi_socket`.

octobanana:
After trying out a couple different implementations for `multi_io_context`, it seemed that it would make more sense if it didn't manage the threads. Removing the threads from the class fixes the lack of access to the `io_context` at thread construction, and allows any thread implementation to be used. What do you think?

There are currently four ways to access the io_contexts, `multi_io_context::get_context()`, `multi_io_context::next_context()`, `multi_io_context::bump_context()`, and `multi_io_context::contexts()`. For the ability to iterate over the `io_contexts`, such as in the on signal handler, I added `multi_io_context::contexts()`, which returns a `std::vector<boost::asio::io_context*> const`. Maybe this could be done better?

Adding `operator[]` and `at()` for random access and `begin()`, `end()` type functions for range-based for loop compatibility would be nice to create a standard-like interface.

octobanana:
Hey, just checking in to see if you've had the chance to look over the changes made since the initial review.

vinniefalco:
ermmm... yes, and I rebased it (see https://github.com/vinniefalco/beast/tree/multi-io)

PR #1353, opened 2018-12-06
Use a struct instead of pair in flat_stream

djarek:
No need to instantiate a pair and a custom struct has better member names.
Doesn't affect public APIs.


vinniefalco:
Seems reasonable

vinniefalco:
Merged in 197

PR #1370, opened 2018-12-13
fix typo

Tocic:


vinniefalco:
This will go into version 200, thank you!

PR #1378, opened 2018-12-19
Use SaxonHE in reference generation

djarek:
Depends on: https://github.com/boostorg/build/pull/378

### DIFF
--- a/doc/docca/include/docca/doxygen.xsl
+++ b/doc/docca/include/docca/doxygen.xsl
@@ -865,14 +865,14 @@
             </xsl:otherwise>
           </xsl:choose>
           <xsl:text> `</xsl:text>
-          <xsl:value-of name="text" select="$dox-name"/>
+          <xsl:value-of select="$dox-name"/>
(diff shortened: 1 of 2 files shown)


doc/docca/include/docca/doxygen.xsl
```diff
@@ -1209,7 +1209,7 @@
       $private &gt; 0">
     <xsl:text>[heading Private Data Members]&#xd;</xsl:text>
     <xsl:text>[table [[Name][Description]]&#xd;</xsl:text>
-    <xsl:for-each select="sectiondef[@kind='private-attrib' or @kind='private-static-attrib']/memberdef" mode="class-table">
+    <xsl:for-each select="sectiondef[@kind='private-attrib' or @kind='private-static-attrib']/memberdef"  >
```
vinniefalco: There seems to be extra spacing before the `>` ?

djarek: Removed.

vinniefalco:
This should wait until after 1.70.

PR #1402, opened 2019-01-05
Example/CppCon2018, Include error_code to acceptor_.set_option() call

johron:
Looking at the example code, it seems like this should be included in the call in order to update the content of the error_code variable.

If that's not the intention, the test for the error_code on the line after is likely unnecessary.

### DIFF
--- a/example/cppcon2018/listener.cpp
+++ b/example/cppcon2018/listener.cpp
@@ -31,7 +31,7 @@ listener(
     }
 
     // Allow address reuse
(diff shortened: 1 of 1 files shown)


vinniefalco:
Yep! Might need that here too:
https://github.com/vinniefalco/CppCon2018

johron:
I'll add another PR for that repo too.

vinniefalco:
This was merged to 203 https://github.com/boostorg/beast/commit/be800436ea40c2d0633e3c1f82583b94e45f6d1f Thanks!

PR #1412, opened 2019-01-13
Version 203

vinniefalco:
* Update networking refresher doc
* Include error code in call to set_option
* saved_handler is a public interface
* Use new saved_handler in websocket
* session_alloc is thread-safe
* examples use flat_buffer
* parse_until is not static


doc/qbk/03_core/1_refresher.qbk
```diff
@@ -0,0 +1,200 @@
...
+it does allocate a copy of the contents.
+
+Buffers described thus far may not be resized. The __DynamicBuffer__
+concept defines a storage type whose size can change. Beast uses
+dynamic buffers in interfaces where the amount of storage required
+to complete an operation is not known ahead of time.
+
+
+
+    [/ [heading Synchronous I/O] ]
```
djarek: Is this heading supposed to be here?

include/boost/beast/core/impl/saved_handler.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+    ebo_pair v_;
+    net::executor_work_guard<
+        net::associated_executor_t<Handler>> wg_;
+
+public:
+    template<class Handler_>
+    impl(alloc_type const& a, Handler_&& h)
+        : v_(a, std::forward<Handler_>(h))
+        , wg_(net::get_associated_executor(v_.h))
```
djarek: This breaks composed operations that use the non-intrusive executor hook, because if the operation has no associated executor, then this will select `system_executor` rather than the I/O executor.

include/boost/beast/core/impl/saved_handler.hpp
```diff
@@ -0,0 +1,193 @@
...
+template<class Handler, class Allocator>
+void
+saved_handler::
+emplace(Handler&& handler, Allocator const& alloc)
+{
+    // Can't delete a handler before invoking
+    BOOST_ASSERT(! has_value());
+    using alloc_type = typename
+        beast::detail::allocator_traits<Allocator>::
+            template rebind_alloc<impl<Handler, Allocator>>;
```
djarek: The handler type should probably be decayed, just in case the user does something out of the ordinary here.

doc/qbk/03_core/1_refresher.qbk
vinniefalco: That's a commented-out heading, I am going to continue with these docs later

include/boost/beast/core/impl/saved_handler.hpp
vinniefalco: I believe that should be okay, because the caller is already required to maintain a work guard for the operation which "owns" this saved handler?

djarek: If a user writes an async algorithm that performs a single operation, there's no need to maintain the work guard.

vinniefalco: the previous version had the guard, you think it wasn't needed?

djarek: I'm not saying the guard is not needed, just that the executor is not selected properly. [async.reqmts.async.work] indicates we need guards for `ex1` and `ex2`, but here, `ex1` is implicitly somewhere else, so we don't actually need it. `ex2` is `get_associated_executor(completion_handler,
ex1)`

PR #1414, opened 2019-01-14
Remove linking with Boost.System

Lastique:
Since Boost.System is header-only now, no need to link with the library.

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -96,7 +96,6 @@ project /boost/beast
     <implicit-dependency>/boost//headers
     <include>.
     <include>./test/extras/include
-    <library>/boost/system//boost_system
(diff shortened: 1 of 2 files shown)


vinniefalco:
I can't really find any fault with this :)

Lastique:
Linking with the library adds a runtime dependency on the Boost.System binary, which is no longer needed. Since you're not building a shared library, it may not seem like a big deal, except for users that use Boost.Build to build their projects with Boost.Beast.

There is no reason to link with Boost.System, and IMO ultimately we should move towards removing the library.

vinniefalco:
Note to self: README.md needs to be updated before merge

Lastique:
I've added a commit updating the readme as well.

PR #1436, opened 2019-02-07
fixed extra semicolon

bebuch:


vinniefalco:
This will go into 211

PR #1447, opened 2019-02-10
V213

vinniefalco:
* Fix posix_file::close handling of EINTR
* basic_stream subsumes stranded_stream
* Use timeouts in HTTP server examples
* Use timeouts in HTTP client examples
* Use tcp_stream in WebSocket client examples
* Use tcp_stream in WebSocket server examples
* Use tcp_stream, HTTP timeouts in advanced servers
* Skip timer if no timeout is set


example/advanced/server/advanced_server.cpp
```diff
@@ -581,79 +561,38 @@ class http_session : public std::enable_shared_from_this<http_session>
     run()
     {
         // Make sure we run on the strand
-        if(! strand_.running_in_this_thread())
+        if(! stream_.get_executor().running_in_this_thread())
```
djarek: Isn't this condition always true?

example/http/client/coro/http_client_coro.cpp
```diff
@@ -52,15 +54,18 @@ do_session(
 
     // These objects perform our I/O
     tcp::resolver resolver{ioc};
-    tcp::socket socket{ioc};
+    stream_type stream(ioc);
```
djarek: The inconsistency of initialization methods might be confusing to some readers.

vinniefalco: Yeah I need to go through all the examples and clean that up

example/advanced/server/advanced_server.cpp
vinniefalco: It is in theory true for the example but I want to make these components bullet-proof so if they are copied and pasted they will have a higher chance of working in adverse conditions. More can be done on this front.

djarek: Because this is `io_context::strand` that condition may not be true - 2 strands may share implementations (`net::strand<E>` exhibits different behavior).

PR #1451, opened 2019-02-13
Version 214

vinniefalco:
* Handler binders use the associated allocator
* Add detail::bind_continuation
* Rewrite the echo-op example

### DIFF
--- a/example/echo-op/echo_op.cpp
+++ b/example/echo-op/echo_op.cpp
@@ -14,233 +14,86 @@
 #include <memory>
 #include <utility>
 
+namespace net = boost::asio;
+namespace beast = boost::beast;
+
 //[example_core_echo_op_1
 
 template<
     class AsyncStream,
+    class DynamicBuffer,
     class CompletionToken>
 auto
-async_echo (AsyncStream& stream, CompletionToken&& token)
+async_echo (AsyncStream& stream, DynamicBuffer& buffer, CompletionToken&& token)
 
 //]
-    -> BOOST_ASIO_INITFN_RESULT_TYPE(CompletionToken, void(boost::beast::error_code));
+    ->
+    typename net::async_result<
+        typename std::decay<CompletionToken>::type,
+        void(beast::error_code)>::return_type;
+
+//------------------------------------------------------------------------------
 
 //[example_core_echo_op_2
 
 /** Asynchronously read a line and echo it back.
 
     This function is used to asynchronously read a line ending
-    in a newline ("LF") from the stream, and then write
-    it back. The function call always returns immediately. The
-    asynchronous operation will continue until one of the
-    following conditions is true:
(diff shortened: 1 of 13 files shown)


example/echo-op/echo_op.cpp
```diff
@@ -14,276 +14,324 @@
...
+
+                        // Keep looping until we get the newline
+                        continue;
+                    }
+
+                    // We have our newline, so send the first `pos` bytes of the
+                    // buffers. The function `buffers_prefix` returns the front part
+                    // of the buffers we want.
+
+                    yield net::async_write(stream_,
```
djarek: Can't this use `net::async_write(stream_, buffer.data(), net::transfer_exactly(pos), std::move(*this)`?

example/echo-op/echo_op.cpp
```diff
@@ -14,276 +14,324 @@
...
+
+                    // The loop terminates here, and we will either deliver a
+                    // successful result or an error to the caller's completion handler.
+
+                    break;
+                }
+
+                // When a composed operation completes immediately, it must not
+                // directly invoke the completion handler otherwise it could
+                // lead to unfairness, starvation, or stack overflow. Therefore,
```
djarek: I'd just scare people that it's UB. If we're dealing with coroutines, calling the handler directly would resume a coroutine that is running.

example/echo-op/echo_op.cpp
```diff
@@ -14,276 +14,324 @@
...
+                // if cont == false (meaning, that the call stack still includes
+                // the frame of the initiating function) then use `net::post`
+                // to cause us to be called again after the initiating function
+                // returns. The function `bind_handler` works similarly to
+                // `std::bind`, allowing bound arguments to be passed to our
+                // completion handler during the dispatch, but also takes care
+                // of forwarding the allocator and executor associations in
```
djarek: "associations" => "customization points"

test/beast/core/test_executor.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    void
+    on_work_started() const noexcept
+    {
+        ++sp_->info_.work;
+    }
+
+    void
+    on_work_finished() const noexcept
+    {
```
djarek: Should this decrement the work count?

vinniefalco: No, the purpose of the counter is just to know that it was called

example/echo-op/echo_op.cpp
vinniefalco: I copied Chris' wording from the executors paper

example/echo-op/echo_op.cpp
vinniefalco: In theory yes, but its nice to see as many beast functions as possible in one example to show their use

PR #1475, opened 2019-02-23
Cleanup in test::stream internals:

djarek:
* avoid one instantiation of `bind_handler_front`
* avoid going through the 2-arg post unnecessarily in `read_op`

### DIFF
--- a/include/boost/beast/_experimental/test/impl/stream.hpp
+++ b/include/boost/beast/_experimental/test/impl/stream.hpp
@@ -29,14 +29,13 @@ class stream::read_op : public stream::read_op_base
     using ex2_type
         = net::associated_executor_t<Handler, ex1_type>;
 
-    class lambda
+    struct lambda
     {
         Handler h_;
         state& s_;
         Buffers b_;
         net::executor_work_guard<ex2_type> wg2_;
 
-    public:
         lambda(lambda&&) = default;
         lambda(lambda const&) = default;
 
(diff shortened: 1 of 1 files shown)


vinniefalco:
Does this work? Do we want it?

include/boost/beast/_experimental/test/impl/stream.hpp
```diff
@@ -99,11 +98,10 @@ class stream::read_op : public stream::read_op_base
...
     {
-        net::post(
-            wg1_.get_executor(),
-            bind_handler(
-                std::move(fn_),
-                ec));
+
+        auto alloc = net::get_associated_allocator(fn_.h_);
+        wg1_.get_executor().post(
+            beast::bind_front_handler(std::move(fn_), ec), alloc);
```
vinniefalco: This doesn't really do anything, it all gets inlined anyway...

include/boost/beast/_experimental/test/impl/stream.hpp
```diff
@@ -162,47 +160,36 @@ struct stream::run_write_op
             "WriteHandler type requirements not met");
 
         ++in_->nwrite;
+        auto const complete_op = [&](error_code ec, std::size_t n)
```
vinniefalco: This should be renamed to `upcall`, I think it would be better (I'll do it)

include/boost/beast/_experimental/test/impl/stream.hpp
djarek: There's a significant difference between this and a 2-arg post. The 2-arg post does something like this, not counting hooks and work guards:
```
auto alloc = net::get_associated_allocator(fn_.h_);
        wg1_.get_executor().post(
          [f = std::move(f)](){
              auto alloc = net::get_associated_allocator(f);
              net::get_associated_executor(f).dispatch(std::move(f));
          }, alloc);
```

vinniefalco:
merged, thank you

PR #1491, opened 2019-03-01
Allow the use of string_body and vector_body with -fno-exceptions

djarek:
`string_body` and `vector_body` will no longer translate all exceptions
to "buffer_overflow" error code. `buffer_overflow` error can now only
occur if the Body's max_size() is exceeded.

Changes required:
Code that relies on exceptions thrown from value_type's reserve/resize
being translated into an error code must implement a mechanism to catch
the exception.


djarek:
Merged in 226.

PR #1493, opened 2019-03-02
Version 226

vinniefalco:
* Support -fno-exceptions
* make_strand is in net::
* Fix HTTP parser static string calculation
* Move parser definitions to .ipp


appveyor.yml
```diff
@@ -103,8 +119,10 @@ install:
 build: off
 
 test_script:
-  - b2 variant=release cxxstd=11 address-model=64 libs/beast/example
-  - b2 variant=release cxxstd=11 address-model=64 --verbose-test libs/beast/test//run-fat-tests
+  - if not "%CXXSTD%" == "" set CXXSTD=cxxstd=%CXXSTD%
+  - if not "%ADDRMD%" == "" set ADDRMD=address-model=%ADDRMD%
```
djarek: Adding a 32 bit build will fail because this setting would also be needed to be provided to vcpkg, when building OpenSSL, example: https://github.com/djarek/certify/blob/master/appveyor.yml#L13

PR #1498, opened 2019-03-04
Fix UB in decorator:

djarek:
- don't assume layout or size overhead of classes with
  virtual members (use split vtable)
- don't use SOO for types with throwing move


djarek:
Merged in 228.

Issue #1506, opened 2019-03-07
master's websocket::stream::async_read() gets confused when handlers with similar signatures are available


RedDwarf69:
Using Boost 1.69 this example compiles. But using Boost `b2e37d03b4aa23f8add2ce3dd37d3041c19591d2` it fails with:

```
/include/boost/beast/core/async_base.hpp:351:13: error: call of ‘(echo_op<boost::beast::websocket::stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp> >, main(int, char**)::<lambda(boost::beast::error_code)> >) (boost::asio::error::basic_errors, int)’ is ambiguous
             h_(std::forward<Args>(args)...);
             ^~
echo_op.cpp:199:6: note: candidate: ‘void echo_op<AsyncStream, Handler>::operator()(boost::beast::error_code, std::size_t) [with AsyncStream = boost::beast::websocket::stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp> >; Handler = main(int, char**)::<lambda(boost::beast::error_code)>; boost::beast::error_code = boost::system::error_code; std::size_t = long unsigned int]’
 void echo_op<AsyncStream, Handler>::
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
echo_op.cpp:188:10: note: candidate: ‘void echo_op<AsyncStream, Handler>::operator()(boost::beast::error_code, bool) [with AsyncStream = boost::beast::websocket::stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp> >; Handler = main(int, char**)::<lambda(boost::beast::error_code)>; boost::beast::error_code = boost::system::error_code]’
     void operator()(boost::beast::error_code ec, bool test)
```

Diff from echo_op example from Boost 1.69
```
@@ -8,6 +8,7 @@
 //
 
 #include <boost/beast/core.hpp>
+#include <boost/beast/websocket/stream.hpp>
 #include <boost/asio.hpp>
 #include <cstddef>
 #include <iostream>
@@ -184,6 +185,7 @@
     // as our intermediate operations complete. Definition below.
     //
     void operator()(boost::beast::error_code ec, std::size_t bytes_transferred);
+    void operator()(boost::beast::error_code ec, bool test);
 };
 
 //]
@@ -209,15 +211,16 @@
         case 0:
             // read up to the first newline
             p.step = 1;
-            return boost::asio::async_read_until(p.stream, p.buffer, "\r", std::move(*this));
[...]

PR #1508, opened 2019-03-07
Fix completion handler invocation signatures

djarek:
Completion handlers should be called with the exact same signature as provided to `async_result`.

Fixes: #1506


djarek:
Merged in 230.

PR #1510, opened 2019-03-07
Remove dependency on lexical_cast in CI

djarek:
lexical_cast is not used in Beast.


djarek:
Doesn't work on Linux.

PR #1514, opened 2019-03-10
fix README.md docs link

nunojpg:


vinniefalco:
This needs to be fixed in the qbk file to keep the section name stable, I will submit a patch

PR #1518, opened 2019-03-14
Android: Use compiler pre-defined __ANDROID__

gjasny:
The ANDROID define us usually originationg from he build system
(like CMake, ndk-build, etc.). But some buildsystems like Bazel
do not set it which brakes Android platform detection.

Better use __ANDROID__ (like boost predef does) which is
predefined by the compiler if targeting Android.

### DIFF
--- a/include/boost/beast/core/impl/file_posix.ipp
+++ b/include/boost/beast/core/impl/file_posix.ipp
@@ -24,7 +24,7 @@
 #include <limits.h>
 
 #if ! defined(BOOST_BEAST_NO_POSIX_FADVISE)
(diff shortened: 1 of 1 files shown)


gjasny:
Do you know how I can re-trigger travis to get a green build? I isn't the red build a merge blocker?

djarek:
Vinnie has his own workflow which doesn't register with github, your PR is already merged into develop:
https://github.com/boostorg/beast/pull/1521/commits

vinniefalco:
Yep, this got merged already - thanks!

Issue #1519, opened 2019-03-14
Disable vulnerable versions of SSL/TLS


djarek:
Anything below TLS1.1 should be considered unsafe:
https://www.globalsign.com/en/blog/disable-tls-10-and-all-ssl-versions/

Usages:
https://github.com/boostorg/beast/search?q=sslv23&unscoped_q=sslv23

vinniefalco:
Pull request?

djarek:
Merged in V233

PR #1520, opened 2019-03-14
Disable vulnerable versions of TLS/SSL

djarek:
TLS1.2 can be used instead, it is available in all currently supported versions of OpenSSL.

Close: #1519


djarek:
Merged in V233

PR #1537, opened 2019-03-22
root_certificates.hpp:  brought in the server certificate

tlanc007:
This allows for the self-sign ssl to work.  For example: [beast issue 1534](https://github.com/boostorg/beast/issues/1534l)


vinniefalco:
Nice! It looks like your local `git config`  doesn't have the email and user set though (I can fix it) note how your avatar is the plain octocat...

PR #1556, opened 2019-04-04
Set parser status and flags even if body_limit_ has been reached

medelh:
Parsing the header of a request that is larger than 1 MB prevents the setting of body_limit and reading the request body

### DIFF
--- a/include/boost/beast/http/impl/basic_parser.ipp
+++ b/include/boost/beast/http/impl/basic_parser.ipp
@@ -508,15 +508,16 @@ finish_header(error_code& ec, std::false_type)
     }
     else if(f_ & flagContentLength)
     {
-        if(len_ > body_limit_)
-        {
-            ec = error::body_limit;
-            return;
-        }
         if(len_ > 0)
         {
             f_ |= flagHasBody;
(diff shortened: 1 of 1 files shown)


vinniefalco:
Seems reasonable, thanks! This will have to go into the develop branch for now, as the Boost 1.70 release is closed (and about to ship). After the release I can merge it to master, but it will not be available in the official Boost distribution until Boost 1.71.

vinniefalco:
@medelh your username and email are not configured locally (using `git config`) so the commit is not linked to your GitHub account, is this intended?

medelh:
> @medelh your username and email are not configured locally (using `git config`) so the commit is not linked to your GitHub account, is this intended?

not really, it was by mistake. Can it still be set?

vinniefalco:
> not really, it was by mistake. Can it still be set?

Yes, it can - and I would prefer that you are properly credited for the contribution. You just need to use `git config` to set your username and email to match the credentials on your GitHub account, amend the commit locally, and then force-push it to your fork - this will update the pull request automatically.

medelh:
Many thanks!

PR #1557, opened 2019-04-05
Fix async_base immediate completion

djarek:
`complete(false, ...)` used the wrong executor when an async operation completed immediately, potentially executing the completion handler on the IO executor in some cases.


djarek:
Already merged.

PR #1568, opened 2019-04-11
Don't use tribool unnecessarily in the examples.

Eelis:


vinniefalco:
we need to search for `tribool` (including in comments) and make sure each occurrence is justified

PR #1579, opened 2019-04-19
Relax requirements for vector_body

djarek:
std::byte is not an arithmetic type, which caused compilation to fail when it was used in vector_body.

### DIFF
--- a/test/beast/http/vector_body.cpp
+++ b/test/beast/http/vector_body.cpp
@@ -9,6 +9,7 @@
 
 // Test that header file is self-contained.
 #include <boost/beast/http/vector_body.hpp>
+#include <cstddef>
 
 namespace boost {
 namespace beast {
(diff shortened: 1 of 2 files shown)


test/beast/http/vector_body.cpp
```diff
@@ -18,6 +19,12 @@ BOOST_STATIC_ASSERT(is_body<vector_body<char>>::value);
 BOOST_STATIC_ASSERT(is_body_writer<vector_body<char>>::value);
 BOOST_STATIC_ASSERT(is_body_reader<vector_body<char>>::value);
 
+#if __cpp_lib_byte >= 201603
```
vinniefalco: is there a boost macro for this? and if so, should we use it instead? ping @glenfe

vinniefalco: I'm hearing that it is okay, this will go into v253.

vinniefalco:
Merged in v253

PR #1587, opened 2019-04-23
Fix uninitalized memory use in deflate_stream

djarek:
Reference in original zlib:
https://github.com/madler/zlib/pull/393


Neustradamus:
@djarek: Please look upstream.

PR #1595, opened 2019-04-29
Fix UB in websocket close tests

djarek:
Extend the lifetime of buffers in tests to avoid use-after-free.


test/beast/websocket/close.cpp
```diff
@@ -473,6 +473,8 @@ class close_test : public websocket_test_suite
         doFailLoop([&](test::fail_count& fc)
         {
             echo_server es{log};
+             std::string const s = "Hello, world!";
```
Eelis: Extra indentation

PR #1611, opened 2019-05-18
Make chat websocket javascript client more user friendly

podsvirov:
One image instead of 1000 words:

make-chat-client-more-user-friendly

And I saved 60 lines of code.


vinniefalco:
Very nice, thanks!

PR #1614, opened 2019-05-19
Version 256

djarek:
* Remove uses of the deprecated `buffers` function
* Remove uses of deprecated methods in websocket tests
* Remove redundant use of `static_string`
* Remove redundant template in service_base
* Expand CI matrix using Azure Pipelines
* Make chat websocket javascript client more user friendly
* `allocator_traits::construct` is used for user-defined types
* Add 1-element specialization for `buffers_cat`
* Fix `buffers_cat` iterator tests
* Don't pessimize-move
* Use steady_timer type
* Preserve operation_aborted on partial message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Version 256:
 * `allocator_traits::construct` is used for user-defined types
 * Add 1-element specialization for `buffers_cat`
 * Fix `buffers_cat` iterator tests
+* Don't pessimize-move
 
 --------------------------------------------------------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Version 256:
 * Remove redundant use of `static_string`
 * Remove redundant template in service_base
 * Expand CI matrix using Azure Pipelines
+* Make chat websocket javascript client more user friendly
 
 --------------------------------------------------------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Version 256:
 * Make chat websocket javascript client more user friendly
 * `allocator_traits::construct` is used for user-defined types
 * Add 1-element specialization for `buffers_cat`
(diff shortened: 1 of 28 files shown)


include/boost/beast/core/impl/basic_stream.hpp
```diff
@@ -161,12 +160,21 @@ close()
 //------------------------------------------------------------------------------
 
 template<class Protocol, class Executor, class RatePolicy>
+template<class Executor2>
```
vinniefalco: I'm not generally opposed to this change, but aren't we just trading one instantiation for another?

include/boost/beast/core/detail/service_base.hpp
```diff
@@ -16,15 +16,10 @@ namespace boost {
 namespace beast {
 namespace detail {
 
-template<class T>
-struct service_id : net::execution_context::id
```
vinniefalco: Doesn't asio do it this way? I thought it was a workaround for a link problem... are you sure this change will be correct everywhere?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Version 256:
 * Remove redundant template in service_base
 * Expand CI matrix using Azure Pipelines
 * Make chat websocket javascript client more user friendly
+* Don't call `allocator_traits<T>::construct` unnecessarily
```
vinniefalco: The change message could be "allocator_traits<T>::construct is for user-defined types"

include/boost/beast/core/impl/basic_stream.hpp
djarek: `bind_executor` has to instantiate at least 3 additional classes (due to support for binding executors for CompletionTokens) and adds a few hundred lines of header material.

include/boost/beast/core/detail/service_base.hpp
djarek: It's possibly a workaround for a problem in a toolset that is not supported by beast anyway.

PR #1620, opened 2019-05-24
Clean up typo in chat websocket javascript client

podsvirov:
It so happened that the code works, because both variables exist,
but the code was not clean.


podsvirov:
Excellent!

PR #1628, opened 2019-06-03
Version 259

djarek:
* Reduce the number of instantiations of filter_token_list
* Remove the use of `static_string` from `http::fields`
* Add gcc-9 to AzP CI test matrix
* Enable split compilation in http::basic_fields
* Remove redundant instation of `static_string` in websocket
* Remove redundant use of `asio::coroutine` in `flat_stream`
* Remove unused includes from `test::stream`
* Move `char_buffer` into a separate file
* Fix coverage collection in AzP CI
* Improve performance of `http::string_to_verb`

### DIFF
--- a/include/boost/beast/core/detail/char_buffer.hpp
+++ b/include/boost/beast/core/detail/char_buffer.hpp
@@ -1,0 +1,78 @@
+//
+// Copyright (c) 2019 Damian Jarek(damian.jarek93@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_CORE_DETAIL_CHAR_BUFFER_HPP
+#define BOOST_BEAST_CORE_DETAIL_CHAR_BUFFER_HPP
+
+#include <boost/config.hpp>
+#include <cstddef>
+#include <cstring>
+#include <cstdint>
+
+namespace boost {
+namespace beast {
+namespace detail {
+
+template <std::size_t N>
+class char_buffer
+{
+public:
+    bool try_push_back(char c)
+    {
+        if (size_ == N)
+            return false;
+        buf_[size_++] = c;
+        return true;
+    }
+
+    bool try_append(char const* first, char const* last)
+    {
+        std::size_t const n = last - first;
+        if (n > N - size_)
+            return false;
+        std::memmove(&buf_[size_], first, n);
+        size_ += n;
+        return true;
+    }
+
+    void clear() noexcept
+    {
+        size_ = 0;
+    }
+
+    char* data() noexcept
+    {
+        return buf_;
+    }
+
+    char const* data() const noexcept
+    {
+        return buf_;
+    }
+
+    std::size_t size() const noexcept
+    {
+        return size_;
+    }
+
+    bool empty() const noexcept
+    {
+        return size_ == 0;
+    }
+
+private:
+    std::size_t size_= 0;
+    char buf_[N];
+};
+
+} // detail
+} // beast
+} // boost
+
+#endif
--- a/include/boost/beast/core/detail/impl/temporary_buffer.ipp
+++ b/include/boost/beast/core/detail/impl/temporary_buffer.ipp
@@ -1,0 +1,67 @@
+//
+// Copyright (c) 2019 Damian Jarek(damian.jarek93@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_DETAIL_IMPL_TEMPORARY_BUFFER_IPP
+#define BOOST_BEAST_DETAIL_IMPL_TEMPORARY_BUFFER_IPP
+
+#include <boost/beast/core/detail/temporary_buffer.hpp>
+#include <boost/beast/core/detail/clamp.hpp>
+#include <boost/core/exchange.hpp>
+
+#include <memory>
+#include <cstring>
+
+namespace boost {
+namespace beast {
+namespace detail {
+
+
+void
+temporary_buffer::append(string_view sv)
+{
+    grow(sv.size());
+    unchecked_append(sv);
+}
+
+void
+temporary_buffer::append(string_view sv1, string_view sv2)
+{
+    grow(sv1.size() + sv2.size());
+    unchecked_append(sv1);
+    unchecked_append(sv2);
+}
+
+void
+temporary_buffer::unchecked_append(string_view sv)
+{
+    auto n = sv.size();
+    std::memcpy(&data_[size_], sv.data(), n);
+    size_ += n;
+}
+
+void
+temporary_buffer::grow(std::size_t sv_size)
+{
+    if (capacity_ - size_ >= sv_size)
+    {
+        return;
+    }
+
+    auto const new_cap = (sv_size + size_) * 2u;
+    BOOST_ASSERT(!detail::sum_exceeds(sv_size, size_, new_cap));
+    char* const p = new char[new_cap];
+    std::memcpy(p, data_, size_);
+    deallocate(boost::exchange(data_, p));
+    capacity_ = new_cap;
+}
+} // detail
+} // beast
+} // boost
+
+#endif
--- a/include/boost/beast/core/string.hpp
+++ b/include/boost/beast/core/string.hpp
@@ -10,82 +10,17 @@
 #ifndef BOOST_BEAST_STRING_HPP
 #define BOOST_BEAST_STRING_HPP
 
-#include <boost/beast/core/detail/config.hpp>
-#include <boost/version.hpp>
-
-#if defined(BOOST_BEAST_USE_STD_STRING_VIEW)
-#include <string_view>
-#else
-#include <boost/utility/string_view.hpp>
-#endif
-
-#include <algorithm>
+#include <boost/beast/core/detail/string.hpp>
 
 namespace boost {
 namespace beast {
 
-#if defined(BOOST_BEAST_USE_STD_STRING_VIEW)
-  /// The type of string view used by the library
-  using string_view = std::string_view;
+/// The type of string view used by the library
+using detail::string_view;
 
-  /// The type of basic string view used by the library
-  template<class CharT, class Traits>
-  using basic_string_view =
-      std::basic_string_view<CharT, Traits>;
-#else
-  /// The type of string view used by the library
-  using string_view = boost::string_view;
-
-  /// The type of basic string view used by the library
-  template<class CharT, class Traits>
-  using basic_string_view =
-      boost::basic_string_view<CharT, Traits>;
-#endif
-
-namespace detail {
-
-inline
-char
-ascii_tolower(char c)
-{
-    return ((static_cast<unsigned>(c) - 65U) < 26) ?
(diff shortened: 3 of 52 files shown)


include/boost/beast/http/impl/fields.hpp
```diff
@@ -751,12 +750,91 @@ equal_range(string_view name) const ->
...
+    ~temporary_buffer() noexcept
+    {
+        deallocate(data_);
+    }
+
+    void append(string_view sv)
+    {
+        auto const sv_size = sv.size();
+        if (max_size() - size_ < sv_size)
+            boost::throw_exception(
```
vinniefalco: These need to use the `BOOST_THROW_EXCEPTION` macro

include/boost/beast/http/impl/fields.hpp
```diff
@@ -990,6 +1068,9 @@ void
 basic_fields<Allocator>::
 set_chunked_impl(bool value)
 {
+    using A = typename beast::detail::allocator_traits<
+        Allocator>::template rebind_alloc<char>;
+    detail::temporary_buffer<A> buf{A{this->get()}};
```
vinniefalco: We should not use the allocator of the `basic_fields` for the temporary string. Instead, we should simply use the system allocator (`std::allocator<char>`).

include/boost/beast/http/impl/fields.hpp
```diff
@@ -1011,78 +1092,25 @@ set_chunked_impl(bool value)
...
-                A> s{A{this->get()}};
-        #endif
-            s.reserve(it->value().size() + 9);
-            s.append(it->value().data(), it->value().size());
-            s.append(", chunked", 9);
-            set(field::transfer_encoding, s);
-        }
+
+        buf.append(it->value());
+        buf.append(", chunked");
```
vinniefalco: We could have a 2-arg `append` which performs only one reallocation / resize.

test/beast/http/fields.cpp
```diff
@@ -24,6 +24,9 @@ namespace http {
 class fields_test : public beast::unit_test::suite
 {
 public:
+    static constexpr std::size_t max_static_buffer =
+        sizeof(http::detail::temporary_buffer<std::allocator<char>>);
```
vinniefalco: This looks like a cheat

djarek: The object can't hide `sizeof + 1` bytes in it, so it's good enough, without requiring a definition visible to user code.

include/boost/beast/http/impl/fields.hpp
```diff
@@ -751,12 +750,99 @@ equal_range(string_view name) const ->
...
+            char* const p = new char[new_cap];
+            std::memcpy(p, data_, size_);
+            deallocate(boost::exchange(data_, p));
+            capacity_ = new_cap;
+        }
+    }
+
+    void deallocate(char* data) noexcept
+    {
+        if (data != buffer_)
```
vinniefalco: LOL nice hack

include/boost/beast/http/impl/fields.hpp
```diff
@@ -751,12 +750,99 @@ equal_range(string_view name) const ->
...
+
+private:
+    void unsafe_append(string_view sv)
+    {
+        auto n = sv.size();
+        std::memcpy(&data_[size_], sv.data(), n);
+        size_ += n;
+    }
+
+    void reserve(std::size_t sv_size)
```
vinniefalco: should we move these function definitions into the .ipp file?

include/boost/beast/core/detail/char_buffer.hpp
```diff
@@ -0,0 +1,78 @@
+//
+// Copyright (c) 2016-2019 Vinnie Falco (vinnie dot falco at gmail dot com)
```
vinniefalco: This one is yours, I would put your name on it.

include/boost/beast/http/impl/verb.ipp
```diff
@@ -135,7 +120,7 @@ string_to_verb(string_view v)
         break;
 
     case 'B':
-        if(v == "IND")
+        if(v =="IND")
```
vinniefalco: How do you know this improves the performance? After the change it constructs a `string_view` from `char const*` which causes the string to be iterated twice? Doesn't it depend on which stdlib is being used (e.g. constexpr constructor)?

djarek: Replaced it with an explicit conversion via a UDL to guarantee that `strlen` is not called.

include/boost/beast/core/string.hpp
```diff
@@ -44,6 +44,11 @@ namespace beast {
 
 namespace detail {
 
+inline string_view operator ""_sv(char const* p, std::size_t n)
```
vinniefalco: I like this

include/boost/beast/http/impl/verb.ipp
```diff
@@ -270,35 +257,35 @@ string_to_verb(string_view v)
...
+        if(v == "BIND"_sv)
             return verb::unbind;
-        if(eq(v, "LINK"))
+        if(v == "LINK"_sv)
             return verb::unlink;
-        if(eq(v, "LOCK"))
+        if(v == "LOCK"_sv)
             return verb::unlock;
-        if(eq(v, "SUBSCRIBE"))
+        if(v == "SUBSCRIBE"_sv)
```
vinniefalco: This is great! It went from the black sheep of the family to the belle of the ball.

include/boost/beast/core/string.hpp
vinniefalco: This definition needs to go in **<boost/beast/core/detail/string.hpp>** (create if it doesn't exist). I use the physical organization of header files to help me know where something is, and also as a clue that something exists. I know that it is just one tiny symbol but the practice of putting an implementation detail in a public header, no matter how small, is not sustainable. Over time, there would be an accumulation of small symbols which are not indexed or locatable by a predictable physical orgnanization, leading to reduced maintainability.

include/boost/beast/http/impl/fields.hpp
```diff
@@ -751,12 +750,99 @@ equal_range(string_view name) const ->
 
 namespace detail {
 
+struct temporary_buffer
```
vinniefalco: This needs to be in its own file

include/boost/beast/http/impl/fields.hpp
vinniefalco: Did you change the meaning of the parameter? Is it the new capacity?

include/boost/beast/core/detail/impl/temporary_buffer.ipp
```diff
@@ -0,0 +1,73 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef BOOST_BEAST_DETAIL_IMPL_TEMPORARY_BUFFER_IPP
+#define BOOST_BEAST_DETAIL_IMPL_TEMPORARY_BUFFER_IPP
+
+#include <boost/beast/core/detail/temporary_buffer.hpp>
```
vinniefalco: I see, this got moved in the later commit

include/boost/beast/core/detail/impl/temporary_buffer.ipp
```diff
@@ -0,0 +1,73 @@
...
+
+namespace boost {
+namespace beast {
+namespace detail {
+
+
+void
+temporary_buffer::append(string_view sv)
+{
+    reserve_additional(sv.size());
```
vinniefalco: I would have gone with `grow` but I guess you like typing a lot

include/boost/beast/core/detail/impl/temporary_buffer.ipp
```diff
@@ -0,0 +1,73 @@
...
+void
+temporary_buffer::append(string_view sv1, string_view sv2)
+{
+    reserve_additional(sv1.size() + sv2.size());
+    unsafe_append(sv1);
+    unsafe_append(sv2);
+}
+
+void
+temporary_buffer::unsafe_append(string_view sv)
```
vinniefalco: I prefer the term `unchecked_append`. This is a marketing issue: I rather not have people saying that beast has "unsafe" functions. But this is a minor nitpick.

include/boost/beast/core/detail/impl/temporary_buffer.ipp
```diff
@@ -0,0 +1,73 @@
...
+void
+temporary_buffer::reserve_additional(std::size_t sv_size)
+{
+    if (capacity_ - size_ >= sv_size)
+    {
+        return;
+    }
+
+    auto constexpr limit = (std::numeric_limits<std::size_t>::max)();
+    if (beast::detail::sum_exceeds(capacity_, capacity_, limit) ||
```
vinniefalco: This works but it would be better if it clamped the new capacity to `limit` instead of just throwing an exception when `capacity*2` is too large. I admit I tried to rearrange the code to achieve this but the results were all messy.

include/boost/beast/websocket/detail/hybi13.ipp
```diff
@@ -47,11 +47,11 @@ make_sec_ws_accept(
     string_view key)
 {
     BOOST_ASSERT(key.size() <= sec_ws_key_type::max_size_n);
-    static_string<sec_ws_key_type::max_size_n + 36> m(key);
-    m.append("258EAFA5-E914-47DA-95CA-C5AB0DC85B11");
+    auto const& guid = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
```
vinniefalco: well you might as well use `_sv` since you have it

vinniefalco:
Please, no more new commits to this PR...they can go into the next one. I'm almost done reviewing.

vinniefalco:
Just a few tweaks and this is ready

include/boost/beast/core/string.hpp
djarek: Should I move the other things in `detail` in this file as well?

vinniefalco: Yes, and for exactly the reason I described. I forgot that `detail::iequals` was even in the file. Looking it over, those other routines need a tidying. They use the outdated "gratuituous template" (`<class=void>`) when they should just use `BOOST_BEAST_DECL` / inline like everything else. `iequals` can probably go in **core/detail/string.ipp** and all of that jazz. `beast::iequals` can be a declaration instead of a one-line inline function that calls the implementation.

include/boost/beast/core/detail/impl/temporary_buffer.ipp
djarek: I reworked the logic a bit to simplify the overflow checks. I also noticed that `basic_fields` currently assumes `std::size_t` is larger than `std::uint16_t`, so I've removed the overflow check in `temporary_buffer`, since it can never overflow, because the input we filter out fits in `std::uint16_t` and the output never exceeds `2*std::numeric_limits<std::uint16_t>::max() + small constant`.

include/boost/beast/core/string.hpp
```diff
@@ -10,82 +10,17 @@
...
-        if(ascii_tolower(a) != ascii_tolower(b))
-            return false;
-        a = *p1++;
-        b = *p2++;
-    }
-    while(n--);
-    return true;
-}
+/// The type of string view used by the library
+using detail::string_view;
```
vinniefalco: The docs will show `detail::string_view` since that's what you put there, which is not ideal. Instead, the aliases should go into **core/string_type.hpp** instead of **core/detail/string.hpp**. I'll make this change before I merge

djarek: wouldn't `using string_view = detail::string_view` work?

vinniefalco:
Top two commits have my changes, I will merge this shortly
https://github.com/vinniefalco/beast/commit/39b0f482096efcad8005c654ed0353f942f3646f

PR #1641, opened 2019-06-23
Version 260

djarek:
* More split compilation in rfc7230.hpp
* Qualify calls to `beast::iequals` in basic_parser.ipp
* More split compilation in websocket/detail/mask.hpp
* Cleanup transitive includes in beast/core/detail/type_traits.hpp
* Simplify generation of sec-websocket-key

### DIFF
--- a/test/beast/core/_detail_base64.cpp
+++ b/test/beast/core/_detail_base64.cpp
@@ -11,6 +11,7 @@
 #include <boost/beast/core/detail/base64.hpp>
 
 #include <boost/beast/_experimental/unit_test/suite.hpp>
+#include <string>
 
 namespace boost {
 namespace beast {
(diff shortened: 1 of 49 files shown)


vinniefalco:
Looks good

test/beast/core/_detail_base64.cpp
```diff
@@ -19,6 +20,35 @@ namespace detail {
 class base64_test : public beast::unit_test::suite
 {
 public:
+    std::string
+    base64_encode(
```
vinniefalco: minor nit this could have been static, but I'm not losing sleep over it.

vinniefalco:
17>C:\Users\vinnie\src\boost\libs\beast\include\boost/beast/http/impl/file_body_win32.hpp(458): error C2760: syntax error: unexpected token 'void', expected 'expression' (compiling source file C:\Users\vinnie\src\boost\libs\beast\test\bench\parser\bench_parser.cpp)

vinniefalco:
Merged

PR #1663, opened 2019-07-22
checking zs.total_out against its value before calling inflate, rathe…

dmorilha-twilio:
…r than 0

### DIFF
--- a/include/boost/beast/websocket/impl/read.hpp
+++ b/include/boost/beast/websocket/impl/read.hpp
@@ -492,11 +492,12 @@ public:
                             empty_block[4] = { 0x00, 0x00, 0xff, 0xff };
                         zs.next_in = empty_block;
                         zs.avail_in = sizeof(empty_block);
+                        auto const total_out = zs.total_out;
                         impl.inflate(zs, zlib::Flush::sync, ec);
                         if(! ec)
                         {
                             // https://github.com/madler/zlib/issues/280
-                            if(zs.total_out > 0)
+                            if(zs.total_out > total_out)
                                 ec = error::partial_deflate_block;
                         }
                         if(impl.check_stop_now(ec))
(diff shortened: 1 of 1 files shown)


vinniefalco:
Any chance you could put together a test?

dmorilha-twilio:
I would love to, but I don't know where to start... any pointers?

vinniefalco:
Yes, have a look at how the existing tests are written:
https://github.com/boostorg/beast/blob/develop/test/beast/websocket/read1.cpp#L63

vinniefalco:
@djarek is a test possible for this?

djarek:
Is there a way to force Beast's websocket stream to send an incomplete fragmented message (without sending the final frame) followed by a close frame?

vinniefalco:
> Is there a way to force Beast's websocket stream to send an incomplete fragmented message

Yeah, I think you just call `write_some` without indicating that it is the final fragment. Then call close.

include/boost/beast/websocket/impl/read.hpp
```diff
@@ -492,11 +492,12 @@ class stream<NextLayer, deflateSupported>::read_some_op
                             empty_block[4] = { 0x00, 0x00, 0xff, 0xff };
                         zs.next_in = empty_block;
                         zs.avail_in = sizeof(empty_block);
+                        auto const total_out = zs.total_out;
                         impl.inflate(zs, zlib::Flush::sync, ec);
                         if(! ec)
                         {
                             // https://github.com/madler/zlib/issues/280
-                            if(zs.total_out > 0)
+                            if(zs.total_out > total_out)
```
vinniefalco: I'm not sure this does anything, isn't `zs.total_out` always 0 at line 495? `zs` is always newly constructed on line 465, which sets `total_out` to `0`. Or am I missing something?

vinniefalco:
This doesn't look like a bug to me, can you please address my comment?

include/boost/beast/websocket/impl/read.hpp
AeroStun: You're not missing anything, it is always initially 0; confirmed not a bug.

vinniefalco:
I'm closing this, since it seems like it is not a bug - feel free to reopen if we have new information (or if I was mistaken in closing it). Thanks!

PR #1664, opened 2019-07-23
Add default dtors to satisfy -Wnon-virtual-dtor

mellery451:
I tested with `b2 -j2 libs/beast/test cxxstd=11 cxxflags="-Wnon-virtual-dtor"`, but please advise if other scenarios need to be verified.

### DIFF
--- a/include/boost/beast/core/async_base.hpp
+++ b/include/boost/beast/core/async_base.hpp
@@ -249,6 +249,10 @@ public:
     /// Move Constructor
     async_base(async_base&& other) = default;
 
+    virtual ~async_base() = default;
+    async_base(async_base const&) = delete;
+    async_base& operator=(async_base const&) = delete;
+
     /** The type of allocator associated with this object.
 
         If a class derived from @ref async_base is a completion
--- a/include/boost/beast/_experimental/unit_test/suite.hpp
+++ b/include/boost/beast/_experimental/unit_test/suite.hpp
@@ -193,6 +193,10 @@ public:
     {
     }
 
+    virtual ~suite() = default;
+    suite(suite const&) = delete;
+    suite& operator=(suite const&) = delete;
+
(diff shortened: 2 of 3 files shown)


include/boost/beast/core/async_base.hpp
```diff
@@ -249,6 +249,10 @@ class async_base
     /// Move Constructor
     async_base(async_base&& other) = default;
 
+    virtual ~async_base() = default;
```
djarek: `virtual` is wrong here, because this class has no `virtual` member functions. Considering the fact that the move constructor is only for exposition in documentation, might as well apply the rule of 0 and wrap it into an ifdef:

```
#if BOOST_BEAST_DOXYGEN
    /// Move Constructor
    async_base(async_base&& other) = default;
#endif
```
This should silence the warning in regular builds.

mellery451: isn't `before_invoke_hook` virtual? If I use your suggestion, I still get the warning for `async_base`.

djarek: Hmm, I missed that. The destructor should be non-virtual(since this class isn't supposed to be used like a typical polymorphic base class) and `protected`. copy/move operations should be `default`ed.

vinniefalco: Yeah but making the destructor virtual has almost no cost, and satisfies the pedants.

mellery451: I think I only really need `suite.hpp` changes in order to be able to use the unit_test stuff in our project, so I'd be happy to omit or modify the `async_base` however you see fit. Thanks.

vinniefalco: I think we should keep the destructor virtual. The class already has a virtual member, and making the destructor virtual will silence possible warnings. The cost is so miniscule as to be virtually nil.

PR #1687, opened 2019-08-23
Fix signed/unsigned mismatch in file_stdio::seek

achartier:
This fixes a C4388 warning when compiling with MSVC.

### DIFF
--- a/include/boost/beast/core/impl/file_stdio.ipp
+++ b/include/boost/beast/core/impl/file_stdio.ipp
@@ -241,7 +241,7 @@ seek(std::uint64_t offset, error_code& ec)
         ec = make_error_code(errc::bad_file_descriptor);
         return;
     }
-    if(offset > (std::numeric_limits<long>::max)())
+    if(offset > static_cast<std::uint64_t>(std::numeric_limits<long>::max()))
     {
         ec = make_error_code(errc::invalid_seek);
         return;


include/boost/beast/core/impl/file_stdio.ipp
```diff
@@ -241,7 +241,7 @@ seek(std::uint64_t offset, error_code& ec)
         ec = make_error_code(errc::bad_file_descriptor);
         return;
     }
-    if(offset > (std::numeric_limits<long>::max)())
+    if(offset > (std::numeric_limits<unsigned long>::max)())
```
basiliscos: I think it should be `std::uint64_t` instead of `unsigned long`

vinniefalco: Hmm... I don't think so. stdio (e.g. `FILE*`) only supports 32 bit offsets: https://en.cppreference.com/w/cpp/io/c/fseek
If you want support for large files you have to use one of the platform-specific implementations.

basiliscos: But if the input argument type to seek is `std::uint64_t` why should we care? 

The warning is quite annoying with `-Wall` flag.

achartier: I misunderstood that check earlier. Updated the change to cast the `std::numeric_limits` value into a `std::uint64_t` instead.

include/boost/beast/core/impl/file_stdio.ipp
```diff
@@ -241,7 +241,7 @@ seek(std::uint64_t offset, error_code& ec)
         ec = make_error_code(errc::bad_file_descriptor);
         return;
     }
-    if(offset > (std::numeric_limits<long>::max)())
+    if(offset > static_cast<std::uint64_t>(std::numeric_limits<long>::max()))
```
vinniefalco: This looks right now, thanks

PR #1689, opened 2019-08-28
cmake: check policy first

wasphin:
fix unkown policy with higher version cmake, e.g. 3.11.4
>  Policy "CMP0074" is not known to this version of CMake.


vinniefalco:
This is merged, thanks!

Issue #1734, opened 2019-10-18
Buffer leak in basic_parser::put


inetic:
The version of [`basic_parser::put` that takes a buffer _sequence_ as an argument](https://github.com/boostorg/beast/blob/045dfeb47f6750a940ced9202af0a6331bdcb085/include/boost/beast/http/impl/basic_parser.hpp#L25) has a bug in the [fallback code path](https://github.com/boostorg/beast/blob/045dfeb47f6750a940ced9202af0a6331bdcb085/include/boost/beast/http/impl/basic_parser.hpp#L53-L56) (the one that is not optimized for single buffer sequences nor those whose size is smaller than `max_stack_buffer`).

The issue is that while `buf_.get()` has size `buf_len_`, it has only `size` number of _valid_ octets. This causes some of the invalid octets (those with index >=`size` and <`buf_len_`) to "leak" from previous calls to `put` into the current one.

The [fix](https://github.com/inetic/beast/commit/aff4f4f3f9dc5aa0b7bf85e4b27a2d842c856f62) is simple. I'm attempting to write a unit test, but that's a bit more challenging due to the fact that the buffers fed into the `put` function need to be crafted to avoid the optimized paths.

We noticed this bug in Boost 1.69, but it's still present in current 1.71 and the [`develop`](https://github.com/boostorg/beast/blob/045dfeb47f6750a940ced9202af0a6331bdcb085) branch.

vinniefalco:
This is a case of me being too clever for my own good. The parser should simply require that the input is in a single contiguous buffer. In a future version of Beast I will likely just build the dynamic buffer into the parser itself and keep it flat to hide this implementation detail.

inetic:
So for completeness, here is a code that reproduces the issue. It's quite ugly because it assumes a particular implementation of `basic_parser::put(ConstBufferSequence const&,...)` and thus I think it's best to just keep it here and not commit to unit tests(?).

```c++
    void
    testIssue1734()
    {
        // Ensure more than one buffer, this is to avoid an optimized path in
        // basic_parser::put(ConstBufferSequence const&,...) which avoids
        // buffer flattening.
        auto multibufs = [](auto buffers) {
            std::vector<net::const_buffer> bs;
            for (auto b : buffers) bs.push_back(b);
            while (std::distance(bs.begin(), bs.end()) < 2) {
                bs.push_back({});
            }
            return bs;
        };

        // XXX: This must be the same as basic_parser<...>::max_stack_buffer.
[...]

PR #1736, opened 2019-10-18
Fix buffer leak in basic_parser::put

inetic:
Fixes issue #1734


include/boost/beast/http/impl/basic_parser.hpp
```diff
@@ -51,9 +51,9 @@ put(ConstBufferSequence const& buffers,
     }
     // flatten
     net::buffer_copy(net::buffer(
-        buf_.get(), buf_len_), buffers);
+        buf_.get(), size), buffers);
     return put(net::const_buffer{
-        buf_.get(), buf_len_}, ec);
+        buf_.get(), size}, ec);
```
vinniefalco: _"That which is untested, contains bugs."_

Issue #1719, opened 2019-10-05
Allow gzip in zlib port (or steps to implement it)


AeroStun:
I am currently writing an API giving back JSON data that may be highly redundant, and I wish to offer users the standard `Content-Encoding`s available.

Thanks to the Zlib C++ port included now in Beast, I have successfully allowed `Content-Encoding: deflate`.
However, I wish to also offer `Content-Encoding: gzip`, especially since not every web client allows deflate.
So my real question boils down to this: What would it take to port the gzip part of zlib into the Inflate/Deflate streams (beside the obvious window bits +16) ?

This is mainly for my personal project, but I would argue that such a facility would be great in Beast so that people who wish to serve compressed HTTP bodies could do so in gzip just as easily as in deflate, given the C Zlib has support for both already.

vinniefalco:
Great question. The reason I pulled out the gzip, is because I did not have tests for it, and because the websocket stream doesn't need it. Better to avoid having untested, unused code that would eventually fall to bit-rot.

I agree that having gzip would be useful and it should be fairly straightforward to put it back, once you spend a little time understanding how I ported zlib to relatively modern, header-only C++. But first, we need to write tests for what is currently there, and it needs to have good a code coverage percentage. Otherwise, we will not know if we broke anything when adding gzip back (and having tests is its own reward).

@djarek did some work on this: https://github.com/djarek/beast/commits/v265 

I think a good course of action is:

1. Review and integrate Damian's changes
2. Fill in any gaps in the existing test coverage
3. Try adding gzip support back in, using the original zlib code as a guide
4. Write tests for gzip

It might also be necessary to apply the more recent mainline zlib commits to beast, this requires porting that C code over to the converted C++ that beast uses. I can answer questions.

AeroStun:
Now onto step n°2; I'm searching for existing zlib tests (so far I found the golang one), as well as places in Beast's zlib that need some test cases.
If you have any tips, I'll gladly take them.

Also, you specified that we may need to apply more recent upstream zlib commits; did you meant that we should do that as part of adding gzip back in, or as a preparation step ?

vinniefalco:
I don't have any tips other than... it might help to get a deep-dive understanding of how huffman coding works. It would be nice to refactor the tests to abstract that compressor/decompressor, so the same tests could be run against zlib.

As for the more recent upstream zlib commits, they should be a preparatory step - ideally after we have a bit better test coverage. Let codecov be the guide:

[...]

PR #1744, opened 2019-10-23
Run upstream zlib against our tests to ensure compatibility

AeroStun:
From #1719 
Comments and reviews are very welcome

[A Travis-CI build simulating this pull request](https://travis-ci.org/AeroStun/beast/builds/601784832) has already been run and did yield a positive outcome

### DIFF
--- a/test/beast/zlib/deflate_stream.cpp
+++ b/test/beast/zlib/deflate_stream.cpp
@@ -283,6 +283,7 @@ public:
 
     using self = deflate_stream_test;
     typedef void(self::*pmf_t)(
+        ICompressor& c,
         int level, int windowBits, int memLevel,
         int strategy, std::string const&);
 
@@ -303,29 +304,26 @@ public:
 
     void
     doDeflate1_beast(
+        ICompressor& c,
         int level, int windowBits, int memLevel,
         int strategy, std::string const& check)
     {
         std::string out;
-        z_params zs;
-        deflate_stream ds;
-        ds.reset(
+        c.init(
             level,
             windowBits,
             memLevel,
-            toStrategy(strategy));
-        out.resize(ds.upper_bound(
-            static_cast<uLong>(check.size())));
-        zs.next_in = (Bytef*)check.data();
-        zs.avail_in = static_cast<uInt>(check.size());
-        zs.next_out = (Bytef*)out.data();
-        zs.avail_out = static_cast<uInt>(out.size());
+            strategy);
+        out.resize(c.bound(check.size()));
+        c.next_in(check.data());
+        c.avail_in(check.size());
+        c.next_out((void*)out.data());
+        c.avail_out(out.size());
         {
             bool progress = true;
             for(;;)
             {
-                error_code ec;
-                ds.write(zs, Flush::full, ec);
+                error_code ec = c.write(Flush::full);
                 if( ec == error::need_buffers ||
                     ec == error::end_of_stream) // per zlib FAQ
                     goto fin;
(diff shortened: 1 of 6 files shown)


test/beast/zlib/deflate_stream.cpp
```diff
@@ -433,6 +433,7 @@ class deflate_stream_test : public beast::unit_test::suite
     {
         doMatrix(c, "Hello, world!", &self::doDeflate1_beast);
         doMatrix(c, "Hello, world!", &self::doDeflate2_beast);
+        log << "no-silence keepalive" << std::endl;
```
vinniefalco: oh no... why?

test/beast/zlib/deflate_stream.cpp
```diff
@@ -25,6 +25,138 @@ namespace zlib {
 
 class deflate_stream_test : public beast::unit_test::suite
 {
+    struct ICompressor {
```
vinniefalco: These are uuuuuugly but I guess that's a consequence of the zlib api...

test/beast/zlib/deflate_stream.cpp
AeroStun: The Travis build using valgrind times out after 10 minutes of silence
Since we now test both Beast.zlib and upstream zlib, the build exceeds the silence limit

test/beast/zlib/deflate_stream.cpp
AeroStun: There's even a `const_cast` in there, because zlib wants its inputs to be non-const, although it doesn't modify them, while Beast rightfully requests a non-const version.
If any improvements come to your mind to make this slightly better, I'll happily apply them

test/beast/zlib/deflate_stream.cpp
AeroStun: See [this Travis job](https://travis-ci.org/AeroStun/beast/jobs/601715039)

test/beast/zlib/deflate_stream.cpp
vinniefalco: I want to believe that there is a more modern abstraction still waiting to be discovered which streamlines the inflate/deflate API and makes it easier to document and understand. Eventually I will investigate and hopefully find it. In the long term, the zlib part of Beast needs to be moved into its own repository and Boost library. In order to pass the Boost review though, some cleaning up will be needed (this doesn't have to happen now). Having it in a separate repository will make things easier on Travis for Beast and Zlib.

AeroStun:
Are there any changes/additions you'd like to see added to that PR, or do you consider it staged for the next version ?

AeroStun:
Rebased on version 275

vinniefalco:
This introduces no changes to the public interfaces correct?

AeroStun:
It does, in the form of a typo correction in an enum name
https://github.com/boostorg/beast/pull/1744/commits/6b5eb041f1cc5559cb008fad766d904a33eae469

vinniefalco:
In **theory** we should use
```
    invalid_code_lengths,
#ifndef BOOST_BEAST_NO_DEPRECATED
    invalid_code_lenths = invalid_code_lengths,
#endif
```

in practice...it doesn't matter, because no one is using these APIs. In fact 99% of users probably don't know they exist...

AeroStun:
Okay
Still thanks for showing me the right way of doing it; will definitely save me time if I encounter a similar issue again.

AeroStun:
Just added your snippet, to do the right thing
No reason for Travis to break.

PR #1751, opened 2019-10-31
https_get: request should send Host header, not Server header

RockinRoel:
If you try the https_get example with the Server header, servers will
respond with 400 Bad Request. This example works as it should when it
uses the Host header.


vinniefalco:
Can't argue with this, lol - thanks!

Issue #1729, opened 2019-10-09
Assert triggered when calling websocket::stream::async_close


RedDwarf69:
Running this code

```
#include <boost/asio.hpp>
#include <boost/beast.hpp>
#include <iostream>

namespace asio = boost::asio;
namespace ip = asio::ip;
using tcp = ip::tcp;

namespace beast = boost::beast;
namespace websocket = beast::websocket;

static auto tcp_socketpair(asio::io_context &ioc)
    -> std::tuple<tcp::socket, tcp::socket> {
  auto acceptor = tcp::acceptor{ioc, tcp::endpoint{tcp::v4(), 5000}};

  auto client_socket = tcp::socket{ioc};
  auto server_socket = tcp::socket{ioc};

  acceptor.async_accept(server_socket, [](const boost::system::error_code &) {
    std::cout << "TCP connection accepted" << std::endl;
  });
  client_socket.async_connect(
      tcp::endpoint{ip::address::from_string("127.0.0.1"), 5000},
      [](const boost::system::error_code &) {
        std::cout << "TCP connection connected" << std::endl;
      });

  ioc.poll();
  ioc.reset();

  return std::make_tuple(std::move(client_socket), std::move(server_socket));
}

int main() {
  auto ioc = asio::io_context{};

  auto [client_socket, server_socket] = tcp_socketpair(ioc);

  auto ws_client = websocket::stream<tcp::socket>{std::move(client_socket)};
  auto ws_server = websocket::stream<tcp::socket>{std::move(server_socket)};

  auto ws_timeout = websocket::stream_base::timeout{
      websocket::stream_base::none(), // handshake timeout
      std::chrono::seconds(30),       // idle timeout
      true};
  ws_server.set_option(ws_timeout);

  ws_server.async_accept([](const beast::error_code &) {
    std::cout << "WS connection accepted" << std::endl;
  });
  ws_client.async_handshake("lala.com", "/", [](const beast::error_code &) {
    std::cout << "WS connection connected" << std::endl;
  });

  ioc.poll();
  ioc.reset();

  beast::flat_buffer buffer;
  ws_server.async_read(buffer, asio::detached);
  ws_server.async_close(websocket::close_reason{}, asio::detached);
}
```
with either Boost 1.71 or develop HEAD results in:

```
TCP connection accepted
TCP connection connected
WS connection accepted
WS connection connected
[...]

PR #1754, opened 2019-11-01
Fix async_close error code when async_read times out

RedDwarf69:
My understanding of the Beast codebase is limited, but AFAIK this works fine. The rationale is explained in https://github.com/boostorg/beast/issues/1729#issuecomment-547392533.


vinniefalco:
seems legit, is there a test?

RedDwarf69:
Actually I pushed the wrong code and the example from 1729 now still doesn't work... sorry, I pushed in a rush because I have to go, will try to fix it tomorrow.

RedDwarf69:
It was actually working, but I was building the test manually against the wrong Beast version.
Anyway, I have also added an unit test now.

PR #1759, opened 2019-11-07
simplify docs generation

grisumbras:


vinniefalco:
I had to roll back the doc Jamfile because of unrelated failures, this will be revisited.

vinniefalco:
What is the status of this?

grisumbras:
This is outdated. We have a docca jam module now and it it used in Beast.

PR #1781, opened 2019-12-14
Fix build on linux

madmongo1:
Minimal changes necessary to get the cmake project to build in a linux environment.


vinniefalco:
The cmake changes were merged in https://github.com/boostorg/beast/commit/7bf58e28e4e6119e6345d780c0da6c568717da26

PR #1783, opened 2019-12-18
Resubmission of the socks example code

madmongo1:


example/contrib/socks-client/socks/detail/char_type.hpp
```diff
@@ -0,0 +1,193 @@
...
+        c == ')')
+        return true;
+    return false;
+}
+
+inline
+string_view
+to_hex(unsigned char c) noexcept
+{
+    static const char* hexstring[] = {
```
vinniefalco: This table could be made 1/3rd smaller by not inserting nulls, e.g.:
```
static char const hex[] = "00" "01" "02" "03" ...
```
(Note how there are no commas). I'm not sure I care enough about it to change it but I just wanted to point it out. If and when we make this a public API then yeah I would change it.

example/contrib/socks-client/socks/handshake.hpp
```diff
@@ -0,0 +1,53 @@
...
+/** Perform the SOCKS v4 handshake in the client role.
+*/
+template<
+    typename AsyncStream,
+    typename Handler
+>
+BOOST_BEAST_ASYNC_RESULT1(Handler)
+async_handshake_v4(
+    AsyncStream& stream,
+    const std::string& hostname,
```
vinniefalco: should use `string_view` instead of `std::string const&`

example/contrib/socks-client/socks/handshake.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <boost/asio/async_result.hpp>
+#include <string>
+
+namespace socks {
+
+/** Perform the SOCKS v4 handshake in the client role.
+*/
+template<
+    typename AsyncStream,
+    typename Handler
```
vinniefalco: This is missing the default-completion-token stuff

example/contrib/socks-client/socks/impl/error.ipp
```diff
@@ -0,0 +1,97 @@
...
+#ifndef SOCKS_IMPL_ERROR_IPP
+#define SOCKS_IMPL_ERROR_IPP
+
+#include <socks/error.hpp>
+
+namespace socks {
+
+namespace detail {
+
+class error_codes : public error_category
```
vinniefalco: There's a better way to structure this code:
https://github.com/vinniefalco/json/blob/develop/include/boost/json/impl/error.ipp#L19
You tuck it all into the function. It isn't urgent to change this now but FYI.

example/contrib/socks-client/socks/impl/error.ipp
```diff
@@ -0,0 +1,97 @@
...
+{
+public:
+    const char*
+    name() const noexcept override
+    {
+        return "SOCKS";
+    }
+
+    BOOST_BEAST_DECL
```
vinniefalco: I would actually like to know if derived overrides need the `__declspec(DLLEXPORT)` (or whatever).

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+#ifndef SOCKS_IMPL_HANDSHAKE_HPP
+#define SOCKS_IMPL_HANDSHAKE_HPP
+
+#include <socks/detail/protocol.hpp>
+#include <boost/beast/core/async_base.hpp>
+#include <boost/beast/core/detail/is_invocable.hpp>
```
vinniefalco: Questionable, since it is a private include. No change advised but keep an eye on this.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+#include <boost/config.hpp> // for BOOST_FALLTHROUGH
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/streambuf.hpp>
```
vinniefalco: asio streambuf is meh

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+type read(source& p)
+{
+    type ret = 0;
+    for (std::size_t i = 0; i < sizeof(type); i++)
+        ret = (ret << 8) | (static_cast<unsigned char>(*p++));
+    return ret;
+}
+
+template<typename type, typename target>
+void write(type v, target& p)
```
vinniefalco: These unconstrained read/write functions are dangerous.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+        *p = static_cast<unsigned char>((v >> (i * 8)) & 0xff);
+}
+
+} // detail
+
+template<
+    class Stream,
+    class Handler,
+    class Buffer,
+    class base_type = boost::beast::async_base<
```
vinniefalco: This `base_type` parameter is unnecessary

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+        Handler, typename Stream::executor_type>>
+class socks4_op : public base_type
+{
+public:
+    socks4_op(socks4_op&&) = default;
+    socks4_op(socks4_op const&) = default;
+
+    socks4_op(Stream& stream, Handler& handler,
+        const std::string& hostname, unsigned short port,
```
vinniefalco: Should be `string_view`

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+        using detail::read;
+
+        auto& response = *response_;
+        auto& request = *request_;
+
+        switch (ec ? 3 : step_)
+        {
+        case 0:
+        {
+            step_ = 1;
```
vinniefalco: All these things could have been done in the constructor instead, and then we could have fewer data members.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    {
+        boost::ignore_unused(bytes_transferred);
+
+        using detail::write;
+        using detail::read;
+
+        auto& response = *response_;
+        auto& request = *request_;
+
+        switch (ec ? 3 : step_)
```
vinniefalco: This needs to use `boost::asio::coroutine` instead of a hand-rolled state machine

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+            break;
+        }
+
+        this->complete_now(ec);
+    }
+
+private:
+    Stream& stream_;
+
+    using BufferPtr = std::unique_ptr<Buffer>;
```
vinniefalco: This is such a straightforward implementation, I don't see the utility in letting `Buffer` be a configurable type. It doesn't need to go into a unique_ptr.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+                std::cout << "* SOCKS remote host: " << domain << ":" << port << std::endl;
+            }
+
+            if (atyp == detail::SOCKS5_ATYP_IPV4)
+            {
+                net::ip::tcp::endpoint remote_endp(
+                    net::ip::address_v4(read<uint32_t>(resp)),
+                    read<uint16_t>(resp));
+
+                std::cout << "* SOCKS remote host: " << remote_endp.address().to_string()
```
vinniefalco: \me raises eyebrows

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+                response,
+                net::transfer_exactly(bytes_to_read),
+                std::move(*this));
+        }
+        case 7:
+        {
+            step_ = 8;
+
+            auto resp = static_cast<const char*>(response.data().data());
+            auto ver = read<uint8_t>(resp);
```
vinniefalco: All this code and other business logic needs to be factored out into individual implementation-specific functions.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+            int atyp = read<uint8_t>(resp);
+
+            if (atyp == detail::SOCKS5_ATYP_DOMAINNAME)
+            {
+                auto domain_length = read<uint8_t>(resp);
+
+                std::string domain;
+                for (int i = 0; i < domain_length; i++)
+                    domain.push_back(read<uint8_t>(resp));
+                auto port = read<uint16_t>(resp);
```
vinniefalco: I don't understand, this code does nothing with `domain` afterwards?

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+                    req += 16;
+                    write<uint16_t>(port_, req);
+                    bytes_to_write = 22;
+                }
+            }
+
+            request.commit(bytes_to_write);
+            return net::async_write(
+                stream_,
+                request,
```
vinniefalco: Are these using the Beast dynamic buffers? They have different semantics from the Asio ones, we need to be careful here.

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    net::async_completion<Handler, void(error_code)> init{ handler };
+    using HandlerType = typename std::decay<decltype(init.completion_handler)>::type;
+
+    static_assert(boost::beast::detail::is_invocable<HandlerType,
+        void(error_code)>::value, "Handler type requirements not met");
+
+    using Buffer = net::basic_streambuf<typename std::allocator_traits<
+        net::associated_allocator_t<HandlerType>>:: template rebind_alloc<char> >;
+
+    socks5_op<AsyncStream, HandlerType, Buffer>(stream, init.completion_handler,
```
vinniefalco: This needs to use `async_initiate`

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+    net::async_completion<Handler, void(error_code)> init{ handler };
+    using HandlerType = typename std::decay<decltype(init.completion_handler)>::type;
+
+    static_assert(boost::beast::detail::is_invocable<HandlerType,
+        void(error_code)>::value, "Handler type requirements not met");
+
+    using Buffer = net::basic_streambuf<typename std::allocator_traits<
+        net::associated_allocator_t<HandlerType>>:: template rebind_alloc<char> >;
+
+    socks4_op<AsyncStream, HandlerType, Buffer>(stream, init.completion_handler,
```
vinniefalco: This needs to use `async_initiate`

example/contrib/socks-client/socks/impl/handshake.hpp
```diff
@@ -0,0 +1,598 @@
...
+template<
+    typename AsyncStream,
+    typename Handler
+>
+BOOST_BEAST_ASYNC_RESULT1(Handler)
+async_handshake_v4(
+    AsyncStream& stream,
+    const std::string& hostname,
+    unsigned short port,
+    std::string const& username,
```
vinniefalco: `string_view`

madmongo1: Great feedback. All actionable. I'll get on it.

vinniefalco: It looks like there's stuff missing. Based on what I'm seeing, we will probably need to make several passes over it.

example/contrib/socks-client/socks/detail/char_type.hpp
madmongo1: label: comment-1

2underscores-vic: Or `const char hexstring[][2] = {` with commas

madmongo1: b08272525 (HEAD -> socks-example, origin/socks-example) comment-1 : detail::to_hex now implemented in terms of continuous string. Added a test

example/contrib/socks-client/socks/handshake.hpp
madmongo1: Interesting. The implementation stores a string in any case. From a stylstic persepctive I would support an argument for "pass string by value, move into place". In reality it will make no difference but conceptually it can avoid copying a long string.

In any case, will change as advised.

madmongo1: label: comment-2

vinniefalco: I'm still not convinced it needs to store any string at all

madmongo1: @vinniefalco Strictly correct for socks 4, socks 5 does not submit the hostname until after authentication, so the impl would need to either cache the supplied hostname or pre-build the buffer for submission in state 5. Similarly, auth credentials are not submitted in socks5 until state 2 so some buffering required there.

Storing a string is minutely inefficient, the tradeoff being maintainability (and the ability to add better debugging/loging logic in future if desired).

In production code I'd store the string, as I know at some point I'm going to want to add some logging or diagnostics. In library code where we are shipping shrink-wrapped guaranteed-correct black boxes, the 'correct' approach is probably a matter of opinion.

I will defer to your decision. You have a lot more experience in the maintenance of public libraries.

vinniefalco: Hmm...didn't know that about SOCKSv5. But just to keep things spicy, if `std::string` uses the small buffer optimization then passing it as a buffer to I/O routines from a composed operation could result in undefined behavior if the string is moved. Up to you to decide what you want to do. I would say, go for whatever is expedient. It is more important we get this merged than to make it "perfect."

example/contrib/socks-client/socks/handshake.hpp
madmongo1: label: comment-3

madmongo1:
@vinniefalco 

Breaking: I think there is a design error in the handshake functions. 
The internal operation creates and discards its own read buffer.
This opens up the possibility of the internal op performing a read which acquires more bytes than is needed to satisfy the response from the socks server, and then throwing away the excess bytes.
This is a logic error, since those bytes may have come from the upstream server.

example/contrib/socks-client/socks/handshake.hpp
```diff
@@ -13,7 +13,7 @@
 #include <socks/config.hpp>
 #include <socks/error.hpp>
 #include <boost/asio/async_result.hpp>
-#include <string>
+#include <boost/utility/string_view.hpp>
```
vinniefalco: `<boost/beast/core/string_type.hpp>` instead

example/contrib/socks-client/socks/impl/async_handshake_v5.hpp
```diff
@@ -36,7 +37,7 @@ template<
     class Buffer,
     class base_type = boost::beast::async_base<
         Handler, typename Stream::executor_type>>
-class socks5_op : public base_type
+class socks5_op : public base_type, public net::coroutine
```
vinniefalco: `boost::asio::coroutine`, because fauxroutines will certainly never make it into the stdlib :)

example/contrib/socks-client/socks/impl/async_handshake_v4.hpp
```diff
@@ -32,77 +32,73 @@
...
+  class Handler,
+  class base_type = boost::beast::stable_async_base
+                    <
+                      Handler, 
+                      typename Stream::executor_type
+                    >
+>
+struct socks4_op 
+: base_type
+, net::coroutine
```
vinniefalco: `boost::asio::coroutine`

example/contrib/socks-client/socks/impl/error.ipp
madmongo1: so done.

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: updated.

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: done

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: done for socks 4. in progress for socks5

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: resolved

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: will be removed in the refactor.

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: done in socks 4. pending for socks 5

example/contrib/socks-client/socks/impl/handshake.hpp
madmongo1: all buffers being replaced with one re-usable buffer string

example/contrib/socks-client/socks/impl/async_handshake_v5.hpp
madmongo1: doesn't this tie the implementation to boost asio? 

Should beast have its own fauxroutine?

madmongo1:
@vinniefalco Having read the SOCKS 4 and 5 RFCs carefully, My view is that the SOCKS4 handshake handler should have an extra parameter of type `ip::tcp::endpoint&` in order to receive the endpoint chosen by the socks server at its client address when connecting to the downstream host.

This data is returned by the socks server as part of the protocol.

This will become particularly acute when imlementing the `BIND` method in socks4, where the socks server is acting as an an accepting host on behalf of the client. The client will need to know the external bound address in order to communitcate it to people who want to make inbound connections.

The more I look at these protocols, the more I take the view that in ASIO terms, SOCKS is a stream layer, in the same way that SSL is a stream layer.

madmongo1:
For socks 5, the stream approach is unavoidable, for reasons I have posted in #beast in slack.

example/contrib/socks-client/socks/impl/async_handshake_v5.hpp
```diff
@@ -397,6 +404,35 @@ class socks5_op : public base_type, public net::coroutine
     int step_ = 0;
 };
 
+struct async_handshake_v5_initiator
+{
+  template<class HandlerType, class AsyncStream>
+  void operator()(
+    HandlerType&& handler,
+    AsyncStream& stream,
```
vinniefalco: I'm not certain, but I believe references need special treatment. You might ask in #general . I think Chris passes them as pointers to avoid some ambiguity (although, this could be because he tries to be C++03 compatible).

example/contrib/socks-client/socks/impl/async_handshake_v5.hpp
```diff
@@ -397,6 +404,35 @@ class socks5_op : public base_type, public net::coroutine
     int step_ = 0;
 };
 
+struct async_handshake_v5_initiator
```
vinniefalco: I think in the latest iteration, the initiating function object has a nested `executor_type` which enables some special thing.

example/contrib/socks-client/socks/impl/socks5_username_password_authentication.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+  auto
+  build_request(string_view username, string_view password)
+  -> error_code
+  {
+    auto error = error_code();
+
+    if (username.size() > 255 or password.size() > 255)
+    {
+      error = net::error::invalid_argument;
```
vinniefalco: I would argue we need to throw `std::length_error` here

example/contrib/socks-client/socks/impl/async_handshake_v5.hpp
madmongo1: I had a look in the asio code. Chris performs a BOOST_ASIO_MOVE_ARG on each argument. In c++11 or better, this will of course ccast all arguments to r-values which is fine because the composed operation will cast them back to l-value references.
In the obsolete pre-c++11 world there might be some boost-move-related issue, but this is no concern of ours.

example/contrib/socks-client/socks/impl/socks5_username_password_authentication.hpp
madmongo1: When we have coroutines, maybe. Remember that this is a composed operation designed to be called from async code within an outer composed operation. The long username and password might have been legitimately passed in by a user who expected that GSSAPI authentication would happen, but instead the socks server mandated username/password. Adding exception handling in an asio fauxroutine would be a horror show.

In any case, it's being hived off to a different branch for "later consideration"

example/contrib/socks-client/socks/impl/async_handshake_v5.hpp
vinniefalco: Well, beast is tied to boost::asio at the moment, so there's no getting around that. I am very confident that Networking-TS / C++ will *never* get fauxroutines so the `net::` prefix is misleading here. This should read `boost::asio::coroutine`. Eventually we might incorporate it into some standalone thing but it can never be in net::.

example/contrib/socks-client/socks/impl/socks5_username_password_authentication.hpp
vinniefalco: Yeah that makes sense.

PR #1792, opened 2019-12-24
Pr1791 travis fix

madmongo1:
Extends PR1791 with a fix to prevent compiler error under clang due to narrowing of chars.


vinniefalco:
This needs to be squashed down to either 1 or 2 commits, with the original author preserved. If you want authorship credit then the 2nd commit should have all of your fixes.

vinniefalco:
Suggested commit messages:
> file_win32 supports UTF-8 paths
> file_stdio supports unicode paths

vinniefalco:
Actually there are 2 or 3 commits here, the file_win32 and file_stdio should be individual commits, and then if you want authorship, all of your fixes in a 3rd commit.

PR #1793, opened 2019-12-24
Pr1791 narrowing fix

### LINKED ISSUE #793: utf8 file names
Implementations of **File** need to support utf-8 as promised.

madmongo1:
Ammends PR1791 By Mike Fischer:

Original Comment:
==============
This brings file_win32 in sync with the documentation.
Previously, the path passed to open worked if encoded in the system
codepage (which is almost never UTF-8).
Now, the path must be encoded as UTF-8, as stated in the
documentation.

Adapt file tests so that for file_win32 all paths include a unicorn
character.

Note that file_stdio on Windows is still broken as before.

Partially fixes #793

Further Edits:
==========
* Fix narrowing conversion compile error only evidenced in clang build
* Amend Change Log and Version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+Version 282:
+
+* file_win32 supports UTF-8 paths
+* file_stdio supports unicode paths
+
 Version 281:
 
 * Travis builds docs
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ endfunction()
 #
 #-------------------------------------------------------------------------------
 
-project (Beast VERSION 281)
+project (Beast VERSION 282)
 
 set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 option (Beast_BUILD_EXAMPLES "Build examples" ON)
--- a/include/boost/beast/core/detail/win32_unicode_path.hpp
+++ b/include/boost/beast/core/detail/win32_unicode_path.hpp
@@ -1,0 +1,82 @@
+//
+// Copyright (c) 2019 Mika Fischer (mika.fischer@zoopnet.de)
+//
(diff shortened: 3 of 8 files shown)


vinniefalco:
Suggested revision:
```
file_win32 supports UTF-8 paths:

fix #793, close #1791, close #1793

This brings file_win32 in sync with the documentation.
Previously, the path passed to open worked if encoded in the system
codepage (which is almost never UTF-8).
Now, the path must be encoded as UTF-8, as stated in the
documentation.

Adapt file tests so that for file_win32 all paths include a unicorn
character.
```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+Version 282:
+
+* file_win32 supports UTF-8 paths
+* file_stdio supports unicode paths
```
Flamefire: More correct: `file_stdio supports UTF-8 paths on MSVC` :/

vinniefalco: I guess that is more accurate..are you saying that other toolchains on windows don't use the dinkum stdlib or whatever it is that msvc uses? I am ignorant about such things (I pretty much stay in visual studio always).

Flamefire: No, it really is the MSVC compiler only due to usage of `BOOST_MSVC`: 

> Defined if the compiler is really Microsoft Visual C++, as opposed to one of the many other compilers that also define _MSC_VER. Has the same value as _MSC_VER. 

from https://www.boost.org/doc/libs/master/libs/config/doc/html/boost_config/boost_macro_reference.html

In #1791 I proposed to use `_wfopen` if on windows but not on MSVC and adjust the checks accordingly. Then it will either fail to compile (unlikely as AFAIK all windows compiler libs have `_wfopen`) or work as advertised which is a desirable outcome

Issue #1394, opened 2018-12-29
Example of explicit, graceful shutdown


vinniefalco:
A new example, or the advanced examples, should demonstrate a clean shutdown with a grace period. It could be triggered by the first SIGINT, or perhaps by an HTTP GET request from the loopback address to a special URL (e.g. "GET /cgi-bin/stop HTTP/1.1").

See: https://github.com/ripple/rippled/commit/9ad2b9be45e768dc2bb088cf9fae71eaf11a7349

benstadin:
I'm interested in an example to gracefully shutdown a Beast based server. Background (long story short) for my case is that I use catch2 in a cross platform project, on OS X with XCode this requires some workarounds and force me to start / stop the server for test cases.

vinniefalco:
Here's one example
https://github.com/vinniefalco/BeastLounge/blob/develop/server/server.cpp#L365

benstadin:
Thanks for the quick reply! Starting to check port this. Though it would be nice if that behavior was part of the default implementation.

benstadin:
It works for me now, but I intentionally did not implement the server_impl class for maintenance reasons. My current implementation is based on the http async server example. I added a close method to the listener like below to shutdown the io context and the acceptor:

```
    void close() {
        if (!_stop) {
            _stop = true;
            ioc_.stop();
            acceptor_.close();
        }
    }
```

The server is handling the shutdown similar to your example, but calls to close the listener like this:

```
// Block the main thread until stop() is called
    {
        std::unique_lock<std::mutex> lock(mutex_);
        cv_.wait(lock, [] {
            return stop_.load();
        });
    }
    
    listenerInstance->close();
    
    // services must be kept alive until after
    // all executor threads are joined.
    
    // If we get here, then the server has
    // stopped, so join the threads before
    // destroying them.
    
    for(auto& t : v)
        t.join();
```

The stop() method looks same as in your example, just without the signal handling. 

Is this ok to do it this way?

vinniefalco:
> Is this ok to do it this way?

I/O objects are not thread safe. You have to post to the strand or executor if you are calling `acceptor_.close()`, and `acceptor_` is being accessed from another thread.

However, since you are calling `ioc_.stop();`, you don't need to close `acceptor_` at all, just let it get destroyed normally.

PR #1795, opened 2019-12-31
Add example for the clean shutdown for an async HTTP server

serhii-muchychka:
I slightly supplemented the example of an async HTTP server

https://github.com/boostorg/beast/issues/1394

### DIFF
--- a/example/http/server/CMakeLists.txt
+++ b/example/http/server/CMakeLists.txt
@@ -8,6 +8,7 @@
 #
 
 add_subdirectory (async)
+add_subdirectory (clean-shutdown)
 add_subdirectory (coro)
 add_subdirectory (fast)
 add_subdirectory (small)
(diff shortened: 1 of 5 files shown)


vinniefalco:
How is this different from the advanced server examples, which already implement shutdown on SIGINT?
https://github.com/boostorg/beast/blob/develop/example/advanced/server/advanced_server.cpp#L653

serhii-muchychka:
It supports shutdown via special command from the loopback

vinniefalco:
> It supports shutdown via special command from the loopback

There are some problems here. First of all, you are calling `ioc.stop()` which won't wait for the HTTP response to be delivered. The program could exit `main` before the socket is finished sending, giving the peer a partial response or no response.

Second, the `handle_request` function is designed to be the same in every example program. Modifying it to respond specially to a particular URL breaks the design pattern. If you want to do this in your own programs, that's fine - but it is not something that I can support in a Beast example.

PR #1807, opened 2020-01-18
V2 transition

madmongo1:
Story so far


include/boost/beast/core/detail/impl/read.hpp
```diff
@@ -17,6 +17,7 @@
 #include <boost/asio/basic_stream_socket.hpp>
 #include <boost/asio/coroutine.hpp>
 #include <boost/throw_exception.hpp>
+#include <boost/beast/core/detail/dynamic_buffer_handle.hpp>
```
vinniefalco: beast includes come first

PR #1814, opened 2020-01-21
Add accessor function to File member of basic_file_body

### LINKED ISSUE #1260: Changing file position is advised in http::file_body but no methods provided for it
http::file_body has a good and convenient way for downloading large files by opening them with file_mode = append_existing. In the documentation (file_base.hpp) It is also advised to change file position:
"The current file position shall be set to the end of
        the file prior to each write."
But I can't see any methods for accomplish this using basic_file_body<File>::value_type class. Moreover, I see that "File file_" is private (not protected) - so I cannot add this functionality by overriding the class. I have to completely rewrite the class to accomplish the desired behavior.

Please add the necessary method (or an additional file_mode) for conveniently downloading large files.

mika-fischer:
Fixes #1260

### DIFF
--- a/include/boost/beast/http/basic_file_body.hpp
+++ b/include/boost/beast/http/basic_file_body.hpp
@@ -111,6 +111,12 @@ public:
     /// Move assignment
     value_type& operator=(value_type&& other) = default;
 
+    /// Return the file
+    File& file()
+    {
+        return file_;
+    }
+
     /// Returns `true` if the file is open
     bool
     is_open() const
--- a/include/boost/beast/http/impl/file_body_win32.hpp
+++ b/include/boost/beast/http/impl/file_body_win32.hpp
@@ -78,6 +78,11 @@ struct basic_file_body<file_win32>
         value_type(value_type&& other) = default;
         value_type& operator=(value_type&& other) = default;
 
+        file_win32& file()
+        {
+            return file_;
+        }
+
         bool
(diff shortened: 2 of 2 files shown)


include/boost/beast/http/basic_file_body.hpp
```diff
@@ -111,6 +111,12 @@ class basic_file_body<File>::value_type
     /// Move assignment
     value_type& operator=(value_type&& other) = default;
 
+    /// Accessor to contained File instance
```
vinniefalco: Should read
```
/// Return the file
```

include/boost/beast/http/impl/file_body_win32.hpp
```diff
@@ -78,6 +78,12 @@ struct basic_file_body<file_win32>
         value_type(value_type&& other) = default;
         value_type& operator=(value_type&& other) = default;
 
+        /// Accessor to contained File instance
```
vinniefalco: Should read
```
/// Return the file
```

mika-fischer: I think that one does not generate docs and has no other doc comments, so I just removed it

include/boost/beast/http/basic_file_body.hpp
mika-fischer: Changed

Issue #1815, opened 2020-01-21
file_body_win32 causes system errors


mika-fischer:
ASIO takes some care to translate system errors to asio error codes.

When transfering a basic_file_body<file_win32> (i.e. using TransmitFile), and there is an error, I receive Win32 system errors. In particular, I've observed

- ERROR_NETNAME_DELETED
- WSAEINTR

It's probably a good idea to map these errors like ASIO does it (see for instance https://github.com/boostorg/asio/blob/6534af41b471288091ae39f9ab801594189b6fc9/include/boost/asio/detail/impl/socket_ops.ipp#L842-L857)

mika-fischer:
How do you mean? This does not translate anything, it just returns a system_error with the Windows-specific error code.

Check my link above to see how ASIO takes care to re-write these Windows-specific error codes to something platform-agnostic...

vinniefalco:
Ah yes I see what you mean.

madmongo1:
@mika-fischer Thanks for taking the time to let us know. I'm making a start on this issue today.

madmongo1:
Before making any changes I felt it best to refer back to the orignal documents describing the intent of `boost::system::error_code`

# Introduction

Quoting from the Introduction page of boost.system:

> Boost.System is part of the C++11 Standard Library. A number of changes, particularly to names, were made by the C++ committee during standardization. The Boost implementation has been tracking those changes.

From this it seems reasonable to take descriptions from the boost documentation as canonical (since we are in the boost domain, and the boost implementation is tracking the intention of the standard).

From the documentation from `boost::system::system_category()`:

> The class error_code describes an object used to hold error code values, such as those **originating from the operating system** or other low-level application program interfaces.

Now from the documentation of `boost::system::error_category`:

> The class error_category defines the base class for types used to identify the source and encoding of a particular category of error code.

> Classes may be derived from error_category to support categories of errors **in addition to those defined in Boost.System**.

# Behaviour in ASIO

It is noteworthy that ASIO does define its own error categories (although these are not documented) and does *sometimes* perform translations from `system_category` errors to asio domain errors.

[...]

PR #1821, opened 2020-01-24
file read sets eof error in case of eof on read:

### LINKED ISSUE #1818: Premature end-of-file leads to infinite loop
- All implementations of `file::read` communicate EOF by returning 0 and no error AFAICT.
- [`basic_file_body<File>::writer::get`](https://github.com/boostorg/beast/blob/46ac848fa46ce5b81f2bb5b221774d15c0cdd128/include/boost/beast/http/basic_file_body.hpp#L324-L380) [only has a `BOOST_ASSERT`](https://github.com/boostorg/beast/blob/46ac848fa46ce5b81f2bb5b221774d15c0cdd128/include/boost/beast/http/basic_file_body.hpp#L361) that fails in the EOF case, but is ignored in release mode. In which case [an empty buffer and `more == true` is returned](https://github.com/boostorg/beast/blob/46ac848fa46ce5b81f2bb5b221774d15c0cdd128/include/boost/beast/http/basic_file_body.hpp#L376-L379).
- This leads the serializer to try reading forever and never getting any data.
- I'm not sure whether the `file_win32` specialization that uses `TransmitFile` has the same issue
- I also don't know whether this (that a file size of an opened file might not be accurate or change over time) can happen on Windows at all.

I would argue that encountering EOF before the expected size is reached to be an important runtime error that should be reported. An assertion is not enough.

Premature EOF can always happen (at least on Linux) if the file is truncated after determining the file size. On Linux it's also possible to have files that report an incorrect size (e.g. 0 for pipes or 4K for sysfs files), and it might be nice to support those as well (or at the very least report an error instead of causing an infinite loop).

madmongo1:
fixes #1818

### DIFF
--- a/include/boost/beast/http/impl/file_body_win32.hpp
+++ b/include/boost/beast/http/impl/file_body_win32.hpp
@@ -17,6 +17,7 @@
 #include <boost/beast/core/buffers_range.hpp>
 #include <boost/beast/core/detail/clamp.hpp>
 #include <boost/beast/core/detail/is_invocable.hpp>
+#include <boost/beast/http/error.hpp>
 #include <boost/beast/http/write.hpp>
 #include <boost/beast/http/serializer.hpp>
 #include <boost/asio/async_result.hpp>
@@ -133,14 +134,17 @@ struct basic_file_body<file_win32>
         template<bool isRequest, class Fields>
         writer(header<isRequest, Fields>&, value_type& b)
             : body_(b)
+            , pos_(body_.first_)
         {
+            // The file must already be open
+            BOOST_ASSERT(body_.file_.is_open());
         }
 
         void
-        init(error_code&)
+        init(error_code& ec)
         {
             BOOST_ASSERT(body_.file_.is_open());
-            pos_ = body_.first_;
+            ec.clear();
         }
 
         boost::optional<std::pair<const_buffers_type, bool>>
@@ -156,6 +160,11 @@ struct basic_file_body<file_win32>
             auto const nread = body_.file_.read(buf_, n, ec);
             if(ec)
                 return boost::none;
+            if (nread == 0)
+            {
+                ec = error::short_read;
+                return boost::none;
+            }
             BOOST_ASSERT(nread != 0);
             pos_ += nread;
             ec = {};
--- a/test/beast/http/error.cpp
+++ b/test/beast/http/error.cpp
@@ -67,6 +67,7 @@ public:
         check("beast.http", error::bad_obs_fold);
 
         check("beast.http", error::stale_parser);
+        check("beast.http", error::short_read);
     }
 };
 
--- a/test/beast/http/file_body.cpp
+++ b/test/beast/http/file_body.cpp
@@ -18,6 +18,7 @@
 #include <boost/beast/http/serializer.hpp>
 #include <boost/beast/_experimental/unit_test/suite.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/asio/error.hpp>
 
 namespace boost {
 namespace beast {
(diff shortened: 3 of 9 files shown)


madmongo1:
It might make sense to use asio file descriptor primitves to model file access, since we're in an asio context.

mika-fischer:
Note that the same problem exists in `file_stdio` and `file_win32`, so this fix is only partial.

test/beast/http/file_body.cpp
```diff
@@ -193,7 +196,10 @@ class file_body_test : public beast::unit_test::suite
     void
     run() override
     {
-        fileBodyUnexpectedEofOnGet();
+#if BOOST_BEAST_USE_POSIX_FILE
+        fileBodyUnexpectedEofOnGet<file_stdio>();
```
mika-fischer: I think the stdio test can always run, no?

include/boost/beast/core/impl/file_posix.ipp
```diff
@@ -284,6 +284,7 @@ read(void* buffer, std::size_t n, error_code& ec) const
         if(result == 0)
         {
             // short read
+            ec = net::error::eof;
```
vinniefalco: I don't think we should return a `net::error` from any "...file" classes. These are supposed to be independent of networking, in fact at some point it should be factored out into its own library because it is generally useful. If you think about it, returning this error could confuse an HTTP read with a file body into thinking the peer shut down the socket, when it didn't.

include/boost/beast/core/impl/file_stdio.ipp
```diff
@@ -298,6 +298,10 @@ read(void* buffer, std::size_t n, error_code& ec) const
         ec.assign(errno, generic_category());
         return 0;
     }
+    if (nread == 0 && std::feof(f_))
+    {
+        ec = net::error::eof;
```
vinniefalco: same thing, a `net::error` is not the right error

test/beast/http/file_body.cpp
```diff
@@ -98,9 +136,89 @@ class file_body_test : public beast::unit_test::suite
         boost::filesystem::remove(temp, ec);
         BEAST_EXPECTS(! ec, ec.message());
     }
+
+#if BOOST_BEAST_USE_POSIX_FILE
+    static void move_to_eof(file_posix& file)
+    {
+        ::lseek(file.native_handle(), 0, SEEK_END);
```
vinniefalco: This could be done portably using `File::pos` and `File::seek`

test/beast/http/file_body.cpp
```diff
@@ -98,9 +136,89 @@ class file_body_test : public beast::unit_test::suite
...
+    }
+
+#if BOOST_BEAST_USE_WIN32_FILE
+    static void move_to_eof(file_win32& file)
+    {
+        ::SetFilePointer(file.native_handle(), 0, 0, FILE_END);
+    }
+#endif
+
+    template<class FileType>
```
vinniefalco: `class File` would be better (Since `File` is a named requirement)

test/beast/http/file_body.cpp
```diff
@@ -98,9 +136,89 @@ class file_body_test : public beast::unit_test::suite
...
     void
     run() override
     {
+        fileBodyUnexpectedEofOnGet<file_stdio>();
+    #if BOOST_BEAST_USE_POSIX_FILE
+        fileBodyUnexpectedEofOnGet<file_posix>();
+    #endif
+    #if BOOST_BEAST_USE_WIN32_FILE
+        fileBodyUnexpectedEofOnGet<file_win32>();
+    #endif
```
vinniefalco: Please move these calls to below the calls to `doTestFileBody` so that the order of invocation matches the order of definitions in the source file.

include/boost/beast/http/impl/file_body_win32.hpp
```diff
@@ -156,6 +160,11 @@ struct basic_file_body<file_win32>
             auto const nread = body_.file_.read(buf_, n, ec);
             if(ec)
                 return boost::none;
+            if (nread == 0)
```
vinniefalco: what if `nread > 0` but `nread` is also less than what we expect in terms of reading everything in the file?

test/beast/http/error.cpp
```diff
@@ -67,6 +67,7 @@ class error_test : public unit_test::suite
         check("beast.http", error::bad_obs_fold);
 
         check("beast.http", error::stale_parser);
+        check("beast.http", error::short_read);
```
vinniefalco: don't we already have a short readerror?

vinniefalco: seems not

include/boost/beast/http/impl/file_body_win32.hpp
mika-fischer: Then we'll get the short read error in the next call to get.

This behavior is similar to POSIX read and fread.

madmongo1: agree. no special handling required here.

madmongo1:
fixes #1818 
fixes #1815

vinniefalco:
"file body reports short_read on eof"

PR #1823, opened 2020-01-26
Fix issue 1822

madmongo1:


example/http/client/async-ssl/http_client_async_ssl.cpp
```diff
@@ -216,9 +228,12 @@ int main(int argc, char** argv)
     auto const target = argv[3];
     int version = argc == 5 && !std::strcmp("1.0", argv[4]) ? 10 : 11;
 
-    // The io_context is required for all I/O
+    // The io_context is one possible executor context
```
vinniefalco: You can't just change something like this in one example, it has to be changed in every example, and we need to agree upon the wording. I agree that it is a bit out of date with Chris' changes to I/O objects.

example/http/client/async-ssl/http_client_async_ssl.cpp
```diff
@@ -195,6 +195,18 @@ class session : public std::enable_shared_from_this<session>
 
         // If we get here then the connection is closed gracefully
     }
+
+private:
+    // private constructor called by public construct after strand has been created on the given
+    // note that the same strand is passed as the executor to the constructors of both
+    // resolver_ and stream_.
```
vinniefalco: The enormous size of this comment is a red flag that the code is wrong

example/http/client/async-ssl/http_client_async_ssl.cpp
```diff
@@ -195,6 +195,18 @@ class session : public std::enable_shared_from_this<session>
...
+    explicit
+    session(net::strand<net::executor> strnd, ssl::context& ctx)
```
vinniefalco: Go big or go small, nothing in between. So, `strand`, or `ex` (for executor) but not `strnd`. I would go with `ex` since that's what is used everywhere else.

example/http/client/async-ssl/http_client_async_ssl.cpp
```diff
@@ -54,9 +54,9 @@ class session : public std::enable_shared_from_this<session>
     // Objects are constructed with a strand to
     // ensure that handlers do not execute concurrently.
     explicit
-    session(net::io_context& ioc, ssl::context& ctx)
-        : resolver_(net::make_strand(ioc))
-        , stream_(net::make_strand(ioc), ctx)
+    session(net::executor exec, ssl::context& ctx)
+    // note: defer to private constructor after having created the strand
+    : session(net::make_strand(exec), ctx)
```
vinniefalco: Anyway what is the necessity for being so verbose? Why not just call `make_strand` at the call site of `make_shared<session>`?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+Version XXX:
+
+* updated root certificates in examples
```
vinniefalco: This should have gone into the other commit

example/http/client/async-ssl/http_client_async_ssl.cpp
```diff
@@ -229,7 +244,7 @@ int main(int argc, char** argv)
     ctx.set_verify_mode(ssl::verify_peer);
 
     // Launch the asynchronous operation
-    std::make_shared<session>(ioc, ctx)->run(host, port, target, version);
+    std::make_shared<session>(exec, ctx)->run(host, port, target, version);
```
vinniefalco: Wouldn't this make more sense as
```
std::make_shared<session>(
    net::make_strand(ioc.get_executor()),
    ctx
        )->run(host, port, target, version);
```

example/http/client/async-ssl/http_client_async_ssl.cpp
```diff
@@ -216,9 +228,12 @@ int main(int argc, char** argv)
     auto const target = argv[3];
     int version = argc == 5 && !std::strcmp("1.0", argv[4]) ? 10 : 11;
 
-    // The io_context is required for all I/O
+    // The io_context is one possible executor context
     net::io_context ioc;
 
+    // All handlers will be invoked through this executor
+    auto exec = ioc.get_executor();
```
vinniefalco: This executor doesn't need to be named

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: The way I would structure this by choice is something like this:
```
asio::io_context ioc;

...

start(ioc.get_executor(), ...);

ioc.run();
```

This ends the dependency on the io_context type as early as possible in the program.

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: Calling make_strand at the call site passes responsibility for thread-safety to the caller. I find this philosophically concerning as it's leaking information.

To be honest, the program does not actually require strands at all. There are no concurrent async operations in play.

I wonder whether we'd be better off removing all mention of strands altogether.

CHANGELOG.md
madmongo1: agree

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: I think it would to you. To me it looks like a horrid mess :-)

example/http/client/async-ssl/http_client_async_ssl.cpp
vinniefalco: > Calling make_strand at the call site passes responsibility for thread-safety to the caller.

Yes, that's exactly right. The caller is the one with the information on whether or not the strand is necessary, since only the caller knows how many threads are calling `io_context::run`.

> the program does not actually require strands at all.

It doesn't, but the philosophy of the examples is to always include them so that users have more robust code in the face of changes. 90% of crashing bug reports are caused by missing strands in multi-threaded contexts.

example/http/client/async-ssl/http_client_async_ssl.cpp
vinniefalco: How is 
```
asio::io_context ioc;
...
start(ioc.get_executor(), ...);
ioc.run();
```
any different from
```
    net::io_context ioc;
    auto exec = ioc.get_executor();
    std::make_shared<session>(exec, ctx)->run(host, port, target, version);
```
except that it adds one unnecessary function? `ioc` is still referenced the same number of times (twice, one for launching the client and the other for calling `run`).

example/http/client/async-ssl/http_client_async_ssl.cpp
vinniefalco: Which part is the mess, the form or the substance?

example/http/client/async-ssl/http_client_async_ssl.cpp
vinniefalco: If anything, we should either include a very brief comment on why we're using strands, and/or include a few sentences in the documentation about the use of strands in the examples. And we should audit every example and make sure they are using strands consistently. I agree that we want to transition from mentioning the `io_context` executor to the `strand` as early as possible.

example/http/client/async-ssl/http_client_async_ssl.cpp
vinniefalco: ...but if it did _need_ a name (which it doesn't, here) then the canonical name for an object of _Executor_ type when only one such instance exists in the scope is `ex`.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+* correct strand use in example
```
vinniefalco: "examples use strands correctly"

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: I guess it's a matter of personal style. It's no big deal.

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: OK that makes sense.

To be more honest, I don't think this particular example requires a shared_ptr either, since there are no concurrent outstanding async operations and the lifetime of the session must necessarily end before the lifetime of the io_context.

Whether this example is the place to teach this is another matter.

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: for me, the form. But then as you know I like to line up my soldiers.

example/http/client/async-ssl/http_client_async_ssl.cpp
madmongo1: For me, the application's initialisation (starting initiatl operations on the executor) is a separate concern to the pumping of messages through the io_context.

I generally prefer to separate them completely.

PR #1827, opened 2020-01-27
CircleCI sandbox

madmongo1:
First stab at getting circleci integration working
runs tests and compiles examples using latest gcc with latest c++ version

this is not a request to pull, just an informational

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,59 @@
+version: 2.0
+jobs:
+    build:
+      docker:
+        - image: madmongo1/fedora-devel:latest
+
+      working_directory: /tmp
+
+      steps:
+        - run:
+            name: fix ssh keys
+            working_directory: ~
+            command: |
+              ssh-keyscan github.com >> githubKey
+              ssh-keygen -lf githubKey
+              cat githubKey >> ~/.ssh/known_hosts
+
+        - run:
+            name: clone boost
+            working_directory: /tmp
+            command: |
+              BOOST_BRANCH=master
+              if [ "${CIRCLE_BRANCH}" != "master" -a "${CIRCLE_BRANCH}" != "refs/heads/master" ]
+              then
+                BOOST_BRANCH=develop
+              fi
+              git clone -b $BOOST_BRANCH --depth 1 https://github.com/boostorg/boost.git
+
+        - run:
+            name: update submodules
+            working_directory: /tmp/boost
+            command: |
+              git submodule update --init --recursive --depth 10 --jobs 4
+
+        - run:
+            name: checkout beast version
+            working_directory: /tmp/boost/libs/beast
+            command: |
+              git remote add circle ${CIRCLE_REPOSITORY_URL}
+              git fetch circle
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,58 @@
...
+            command: |
+              git remote add circle ${CIRCLE_REPOSITORY_URL}
+              git fetch circle
+              git checkout ${CIRCLE_BRANCH}
+              git submodule update --init --recursive
+
+        - run:
+            name: bootstrap
+            working_directory: /tmp/boost
+            command: ./bootstrap.sh
```
djarek: You can use this script to grab just the necessary dependencies: https://github.com/boostorg/beast/blob/develop/tools/get-boost.sh

madmongo1: I'm wary of that script because each submodule is explicitly named. Submodules seem to come and go in master boost's develop branch.

madmongo1:
OK, this PR has just become a pull request.

Circleci build and test is working. I'd like to use it as a base for further development but there is value for me in it going into the develop branch as until it does, I'll get spurious circleci errors whenever I push a branch that is based on develop.

madmongo1:
What?? Why is appveyor reporting a failure when nothing in it failed?

.circleci/config.yml
vinniefalco: No that's not correct, changes to submodules are infrequent.

madmongo1: I was caught out by a change to submodules in boost develop branch just last week!

vinniefalco: In other words, `int sample_size()` returned 1

madmongo1: Thank you for bringing the script to my attention. 

I have used pertinent details from it to improve the throughput of my circleci configuration file.

I have chosen not to use the script as-is for two reasons:

1. It would construct a git tree in the circleci host that does not exactly match the one in my workflow. This would potentially make debugging of issues more difficult for me through circle's ssh debugging interface.

2. The double-entry bookeeping required to keep the script in sync with master and develop of the git root repo is unecessary work. The discovered git repo already has this information intrinsically.

If there are any technical issues that prevent the correct functioning of a circleci build of beast in this PR please kindly bring them to my attention.

If any part of the submitted configu file can be proven subobtimal I would be happy to hear about it.

As things stand I belive this PR is perfectly correct.

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+              then
+                BOOST_BRANCH=develop
+              fi
+              git clone -b $BOOST_BRANCH --depth 1 https://github.com/boostorg/boost.git
+
+        - run:
+            name: update submodules
+            working_directory: /tmp/boost
+            command: |
+              git submodule update --init --recursive --depth 1 --jobs 4
```
djarek: A depth=1 submodule update may result in spurious failures whenever somebody commits to a dependency and the root repo isn't updated yet.

madmongo1: confirmed. I'll use 10 as per the script

PR #1829, opened 2020-01-29
fix fields erase bug

### LINKED ISSUE #1828: http::basic_fields bug
Using this sequence of actions...

beast::http::request<beast::http::string_body> req;
req.insert("abc", "1");
req.insert("abc", "2");
req.insert("abc", "3");

...a call to req.count("abc") returns 3 as expected.  If we then do the following...

auto iter = req.find("abc");
req.insert("abc", "4");
req.erase(iter);

...a call to req.count("abc") returns 0 instead of 3.  However, the header was properly updated since printing the request gives...

 HTTP/1.1
abc: 1
abc: 2
abc: 4

madmongo1:
fixes #1828

### DIFF
--- a/include/boost/beast/http/impl/fields.hpp
+++ b/include/boost/beast/http/impl/fields.hpp
@@ -617,7 +617,7 @@ erase(const_iterator pos) ->
 {
     auto next = pos;
     auto& e = *next++;
-    set_.erase(e);
+    set_.erase(set_.iterator_to(e));
     list_.erase(pos);
     delete_element(const_cast<element&>(e));
     return next;
--- a/test/beast/http/fields.cpp
+++ b/test/beast/http/fields.cpp
@@ -989,6 +989,21 @@ public:
     }
 
     void
+    testIssue1828()
+    {
+        beast::http::fields req;
+        req.insert("abc", "1");
+        req.insert("abc", "2");
+        req.insert("abc", "3");
+        BEAST_EXPECT(req.count("abc") == 3);
+        auto iter = req.find("abc");
+        BEAST_EXPECT(iter->value() == "1");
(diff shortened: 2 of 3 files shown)


include/boost/beast/http/impl/fields.hpp
```diff
@@ -617,7 +617,7 @@ erase(const_iterator pos) ->
 {
     auto next = pos;
     auto& e = *next++;
-    set_.erase(e);
+    set_.erase(set_.iterator_to(e));
```
vinniefalco: I'm a bit disappointed that `set_.erase(e);` is allowed to compile and then malfunction

vinniefalco:
"fix fields erase bug" is redundant. Prefer "fix erase field"

include/boost/beast/http/impl/fields.hpp
madmongo1: One of the dangers of composition through inheritance.

madmongo1:
all comments addressed

test/beast/http/fields.cpp
```diff
@@ -988,9 +988,25 @@ class fields_test : public beast::unit_test::suite
...
+        BEAST_EXPECT(iter->value() == "1");
+        req.insert("abc", "4");
+        req.erase(iter);
+        BEAST_EXPECT(req.count("abc") == 3);
+    }
+
     void
     run() override
     {
+        testIssue1828();
```
vinniefalco: The order of calls should match the order of function declarations/definitions. Since you put `testIssue1828()` after all the other functions (which is the right thing to do), it should be called after all the other functions.

madmongo1: done

PR #1834, opened 2020-02-04
moved-from container with unequal allocator will not be modified

madmongo1:
Currently proposed as 3 commits for discussion and analysis.

Will squash before final review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version XXX:
+
+* moved-from dynamic buffers do not clear if different allocator
+
+--------------------------------------------------------------------------------
+
 Version 284:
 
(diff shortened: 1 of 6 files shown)


vinniefalco:
not much to talk about here, it looks correct

madmongo1:
OK, will squash and force-push

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Version XXX:
```
vinniefalco: why not just put 285 here

madmongo1: in case it doesn't make it into 285

vinniefalco: XXX is what you put on a barrel full of dynamite

PR #1835, opened 2020-02-04
flat_buffer shrink_to_fit never throws

### LINKED ISSUE #1831: basic_flat_buffer::shrink_to_fit should not throw
it should eat the exception, like this:
https://github.com/vinniefalco/json/blob/2b8ac041fce779644b94c0680f0fdda589be8eac/include/boost/json/detail/impl/string_impl.ipp#L182

madmongo1:
fixes #1831

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version XXX:
+
+* flat_buffer shrink_to_fit is noexcept
+
+--------------------------------------------------------------------------------
+
 Version 284:
 
 * fix compilation macro documentation
--- a/include/boost/beast/core/impl/flat_buffer.hpp
+++ b/include/boost/beast/core/impl/flat_buffer.hpp
@@ -262,17 +262,32 @@ reserve(std::size_t n)
 template<class Allocator>
 void
 basic_flat_buffer<Allocator>::
-shrink_to_fit()
+shrink_to_fit() noexcept
 {
     auto const len = size();
+
     if(len == capacity())
         return;
+
     char* p;
     if(len > 0)
     {
         BOOST_ASSERT(begin_);
         BOOST_ASSERT(in_);
-        p = alloc(len);
+#ifndef BOOST_NO_EXCEPTIONS
+        try
+        {
+#endif
+            p = alloc(len);
+#ifndef BOOST_NO_EXCEPTIONS
+        }
+        catch(std::exception const&)
+        {
+            // swallow this exception
+            // and abort the shrink
+            return;
+        }
+#endif
         std::memcpy(p, in_, len);
     }
     else
--- a/include/boost/beast/core/flat_buffer.hpp
+++ b/include/boost/beast/core/flat_buffer.hpp
@@ -358,17 +358,17 @@ public:
     void
     reserve(std::size_t n);
 
-    /** Reallocate the buffer to fit the readable bytes exactly.
+    /** Request the removal of unused capacity.
 
-        Buffer sequences previously obtained using @ref data or
-        @ref prepare become invalid.
+        This function attempts to reduce @ref capacity()
+        to @ref size(), which may not succeed.
 
         @esafe
 
-        Strong guarantee.
+        No-throw guarantee.
     */
     void
-    shrink_to_fit();
+    shrink_to_fit() noexcept;
 
     /** Set the size of the readable and writable bytes to zero.
 


include/boost/beast/core/impl/flat_buffer.hpp
```diff
@@ -265,14 +265,29 @@ basic_flat_buffer<Allocator>::
...
         BOOST_ASSERT(in_);
-        p = alloc(len);
+#ifndef BOOST_NO_EXCEPTIONS
+        try
+        {
+#endif
+            p = alloc(len);
+#ifndef BOOST_NO_EXCEPTIONS
+        }
+        catch(std::exception&)
```
vinniefalco: shouldn't this be `const&`?

include/boost/beast/core/impl/flat_buffer.hpp
```diff
@@ -265,14 +265,29 @@ basic_flat_buffer<Allocator>::
 shrink_to_fit()
```
vinniefalco: `noexcept`?

include/boost/beast/core/impl/flat_buffer.hpp
madmongo1: I believe best practice is to catch exceptions by mutable reference. It allows the the exception to be modified and re-thrown at the catch site without slicing (since exceptions are normally caught by their base class).

include/boost/beast/core/impl/flat_buffer.hpp
madmongo1: This would change the signature of the public interface. I didn't want to do this without concensus.

vinniefalco: I think we can mark it `noexcept` safely

include/boost/beast/core/impl/flat_buffer.hpp
vinniefalco: there's no slicing, exceptions are not passed by value, and `const&` informs the reader that the exception is not modified

include/boost/beast/core/impl/flat_buffer.hpp
madmongo1: OK. I'll check the documentation in case it needs to mention it

vinniefalco:
Commit message should read `flat_buffer shrink_to_fit is noexcept`

include/boost/beast/core/impl/flat_buffer.hpp
```diff
@@ -262,17 +262,32 @@ reserve(std::size_t n)
...
     if(len > 0)
     {
         BOOST_ASSERT(begin_);
         BOOST_ASSERT(in_);
-        p = alloc(len);
+#ifndef BOOST_NO_EXCEPTIONS
+        try
+        {
+#endif
+            p = alloc(len);
```
glenfe: Isn't it an invariant of basic_flat_buffer that size() is less than allocator's max_size()? If so, you don't need to call the checked alloc() function here if you can call allocate() directly.

vinniefalco: calling `allocate()` directly is just so wordy and cumbersome that I mentally defaulted to calling the member function instead

madmongo1: `alloc()` is defined in the same translation unit as its caller. I'd be stunned if the double check isn't elided by the optimiser in all honesty.

madmongo1: resolved the comment wording. Noted the redundant test but I am reluctant to change existing logic in this commit. Is it worh opening another issue?

madmongo1: ...and learning to spell?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Version XXX:
+
+* flat_buffer shrink_to_fit is noexcept
```
vinniefalco: the commit message should match this text

include/boost/beast/core/impl/flat_buffer.hpp
glenfe: As I pointed out to Richard `allocator_traits` are not required for `allocate` and `deallocate`, so it's not really all that wordy if you just did: `p = get().allocate(len);` 

(All of the `allocate()` and `deallocate()` in this file can avoid `allocator_traits`).

vinniefalco: @madmongo1 up to you if you want to make the change that Glen suggests. I wouldn't call it a change in "logic" more like a cosmetic change. Allocators are kind of annoying - Glen, not all of us have memorized all the arcane rules for when you need traits and when you don't.

PR #1836, opened 2020-02-04
Version 285

madmongo1:


include/boost/beast/http/impl/file_body_win32.hpp
```diff
@@ -337,6 +337,46 @@ class null_lambda
     }
 };
 
+// https://github.com/boostorg/beast/issues/1815
+// This function mimics the behaviour of ASIO.
```
vinniefalco: I would write
```
// RHODGES This function mimics the behaviour of ASIO....
```
To make it clear that this is developer commentary and not documentation about the code

madmongo1: This is how I get blamed for things I didn't say :-)

vinniefalco: Alternatively you could use VFALCO

include/boost/beast/core/impl/flat_buffer.hpp
```diff
@@ -262,17 +260,32 @@ reserve(std::size_t n)
...
+#ifndef BOOST_NO_EXCEPTIONS
+        try
+        {
+#endif
+            p = alloc(len);
+#ifndef BOOST_NO_EXCEPTIONS
+        }
+        catch(std::exception const&)
+        {
+            // swallow this exception
```
vinniefalco: The optics on swallowing followed by an abortion are not great

madmongo1: :-D

PR #1838, opened 2020-02-05
Async compose 1773

madmongo1:
Track the conversion of Beast to use the async_compose mechanism in asio.

### DIFF
--- a/example/echo-op/echo_op.cpp
+++ b/example/echo-op/echo_op.cpp
@@ -97,15 +97,205 @@ async_echo (
                 >::return_type;
 //]
 
-//[example_core_echo_op_3
+//[example_core_echo_op_4
 
-template<class AsyncStream, class Handler>
-class echo_op;
+// This nested class implements the echo composed operation as a
+// stateful completion handler. We allow async_compose to
+// take care of boilerplate and we derived from asio::coroutine to
+// allow the reenter and yield keywords to work.
+
+template<class AsyncStream, class DynamicBuffer>
+class echo_op : public boost::asio::coroutine
+{
+    AsyncStream& stream_;
+    DynamicBuffer& buffer_;
+
+public:
+    echo_op(
+        AsyncStream& stream,
+        DynamicBuffer& buffer)
+    : stream_(stream)
+    , buffer_(buffer)
+    {
+    }
+
+    // If a newline is present in the buffer sequence, this function returns
+    // the number of characters from the beginning of the buffer up to the
+    // newline, including the newline character. Otherwise it returns zero.
+
+    std::size_t
+    find_newline(typename DynamicBuffer::const_buffers_type const& buffers)
+    {
+        // The `buffers_iterator` class template provides random-access
+        // iterators into a buffer sequence. Use the standard algorithm
+        // to look for the new line if it exists.
+
+        auto begin = net::buffers_iterator<
+            typename DynamicBuffer::const_buffers_type>::begin(buffers);
+        auto end =   net::buffers_iterator<
+            typename DynamicBuffer::const_buffers_type>::end(buffers);
+        auto result = std::find(begin, end, '\n');
+
+        if(result == end)
+            return 0; // not found
+
+        return result + 1 - begin;
+    }
+
+    // A type to differentiate the source of an invocation of Self
+
+    struct completion_tag {};
+
+    // It is a requirement of async operations that the completion handler must be
+    // invoked as if by @ref post. The composed operation must know whether it is
+    // a continuation or not. This function and the operator()(completion_tag, ...)
+    // allow us to detect whether we should post the completion or simply call it as a continuation.
+
+    template<class Self>
+    void
+    do_completion(Self& self, beast::error_code const& ec)
+    {
+        // Important: Allow ADL to find the correct overload.
+        // Calling boost::asio::asio_handler_is_continuation will
+        // call the default overload because Self is a type in the
+        // boost::asio::detail namespace
+        if (asio_handler_is_continuation(&self))
+        {
+            (*this)(self, completion_tag(), ec);
+        }
+        else
+        {
+            boost::asio::post(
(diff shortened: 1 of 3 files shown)


include/boost/beast/core/detect_ssl.hpp
```diff
@@ -613,7 +562,7 @@ operator()(error_code ec, std::size_t bytes_transferred, bool cont)
         // function template. This reduces compile times and the size
         // of the program executable.
 
-        if(! cont)
+        if(! asio_handler_is_continuation(&self))
```
vinniefalco: eww we don't want to use this anymore

include/boost/beast/core/detect_ssl.hpp
```diff
@@ -342,48 +343,10 @@ namespace detail {
...
 class detect_ssl_op;
 
-// This is a function object which `net::async_initiate` can use to launch
-// our composed operation. This is a relatively new feature in networking
-// which allows the asynchronous operation to be "lazily" executed (meaning
-// that it is launched later). Users don't need to worry about this, but
-// authors of composed operations need to write it this way to get the
-// very best performance, for example when using Coroutines TS (`co_await`).
-
-struct run_detect_ssl_op
```
vinniefalco: We have to teach `async_initiate`

example/echo-op/echo_op.cpp
```diff
@@ -97,132 +97,114 @@ async_echo (
...
-        beast::executor_type<AsyncStream> /*< The type of executor used by the stream to dispatch asynchronous operations >*/
-    >;
+    struct completion_tag {};
 
-    // This nested class implements the echo composed operation as a
-    // stateful completion handler. We derive from `async_base` to
-    // take care of boilerplate and we derived from asio::coroutine to
-    // allow the reenter and yield keywords to work.
+    // It is a requirement of async operations that the completion handler must be
+    // invoked as if by @ref post. The composed operation must know whether it is
```
vinniefalco: `post` is not a Beast function so making it a `@ref` won't work. Use `net::post` instead of `@ref post`.

example/echo-op/echo_op.cpp
```diff
@@ -240,10 +222,10 @@ async_echo(
...
 
                         bytes_to_read = std::min<std::size_t>(
-                              std::max<std::size_t>(512,                // under 512 is too little,
-                                  buffer_.capacity() - buffer_.size()),
-                              std::min<std::size_t>(65536,              // and over 65536 is too much.
-                                  buffer_.max_size() - buffer_.size()));
+                            std::max<std::size_t>(512,                // under 512 is too little,
+                                                  buffer_.capacity() - buffer_.size()),
+                            std::min<std::size_t>(65536,              // and over 65536 is too much.
+                                                  buffer_.max_size() - buffer_.size()));
```
vinniefalco: \sigh

example/echo-op/echo_op.cpp
```diff
@@ -297,36 +279,66 @@ async_echo(
...
+    // Perform some type checks using static assert, this helps
+    // with more friendly error messages when passing the wrong types.
+    static_assert(
+        beast::is_async_stream<AsyncStream>::value,
+        "AsyncStream type requirements not met");
+    static_assert(
+        net::is_dynamic_buffer<DynamicBuffer>::value,
+        "DynamicBuffer type requirements not met");
+
+    // Create the composed operation and launch it. @ref async_compose
```
vinniefalco: This is not a Javadoc, and even if it was, `async_compose` is not a Beast function so using `@ref` on it won't work. You need to use `boost::asio::async_compose` here, not `net::async_compose` (because it isn't in the TS yet).

example/echo-op/echo_op.cpp
```diff
@@ -297,36 +279,66 @@ async_echo(
...
+    // takes care of creating the composed operation, embedding the completion
+    // handler and binding the correct executor and allocators.
+    // We must provide the implementation of the operation (echo_op) and
+    // pass the completion token plus any io_objects involved in the composed operation.
```
vinniefalco: "I/O objects"

example/echo-op/echo_op.cpp
madmongo1: An unfortunate oversight due to ide autoformatting whil copy/pasting the code. In order to compile it could no longer be a template class.

madmongo1: An unfortunate oversight due to ide autoformatting whil copy/pasting the code. In order to compile it could no longer be a template class.

example/echo-op/echo_op.cpp
```diff
@@ -297,36 +279,66 @@ async_echo(
...
+    // to the completion token.
+
+    return boost::asio::async_compose<CompletionToken,
+        void(beast::error_code)>(
+            echo_op<
+                AsyncStream, DynamicBuffer>(
+                    stream, buffer),
+            token,
+            stream
+        );
```
vinniefalco: Example code can run wide, so this is also OK:
```
    return boost::asio::async_compose<CompletionToken, void(beast::error_code)>(
            echo_op<AsyncStream, DynamicBuffer>(stream, buffer),
            token,
            stream
        );
```

example/echo-op/echo_op.cpp
```diff
@@ -97,132 +97,114 @@ async_echo (
...
-        find_newline(typename DynamicBuffer::const_buffers_type const& buffers)
+        else
         {
-            // The `buffers_iterator` class template provides random-access
-            // iterators into a buffer sequence. Use the standard algorithm
-            // to look for the new line if it exists.
+            boost::asio::post(
+                beast::bind_front_handler(
+                    std::move(self),
+                    echo_op::completion_tag(),
```
vinniefalco: Yuck, I don't know if I want to teach `completion_tag`

madmongo1:
What are we going to do with this? It feels as if async_compose is not quite finished.

PR #1841, opened 2020-02-07
v2 dynamic_buffers

madmongo1:
NOT A PULL REQUEST

### DIFF
--- a/include/boost/beast/core/detail/config.hpp
+++ b/include/boost/beast/core/detail/config.hpp
@@ -101,4 +101,12 @@ namespace net = boost::asio;
 #define BOOST_BEAST_ASYNC_TPARAM2 BOOST_ASIO_COMPLETION_TOKEN_FOR(void(::boost::beast::error_code, ::std::size_t))
 #endif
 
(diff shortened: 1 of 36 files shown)


include/boost/beast/core/detail/config.hpp
```diff
@@ -101,4 +101,12 @@ namespace net = boost::asio;
 #define BOOST_BEAST_ASYNC_TPARAM2 BOOST_ASIO_COMPLETION_TOKEN_FOR(void(::boost::beast::error_code, ::std::size_t))
 #endif
 
+#ifndef BOOST_BEAST_V1_DYNAMIC_BUFFER_PROXY
+#define BOOST_BEAST_V1_DYNAMIC_BUFFER_PROXY(type) \
+    typename std::enable_if< \
```
vinniefalco: as long as we are fully qualifying namespaces why not `::std`?

madmongo1: actually, this define can go now. The return type of dynamic_buffer is now a documented template type.

PR #1859, opened 2020-02-25
core: fix ostream flush() usage

### LINKED ISSUE #1853: boost::beast::ostream makes the DynamicBuffer twice as big as it should
```
#include <boost/beast/core/flat_buffer.hpp>
#include <boost/beast/core/ostream.hpp>
#include <iostream>

int main()
{
    boost::beast::flat_buffer buffer;
    boost::beast::ostream(buffer) << "Hello, world!" << std::endl;
    std::cout << "string size: " << std::string{"Hello, world!\n"}.size() << std::endl;
    std::cout << "buffer size: " << buffer.size() << std::endl;
}
```
outputs

```
string size: 14
buffer size: 28
```

If you remove the `std::endl` the problem goes away.

I guess it has to do with not handling the `flush` calls correctly?

RedDwarf69:
After calling the DynamicBuffer commit function the buffer received via
prepare() can be invalidated but the ostream kept using it.

Fixes #1853.

### DIFF
--- a/include/boost/beast/core/detail/ostream.hpp
+++ b/include/boost/beast/core/detail/ostream.hpp
@@ -100,6 +100,7 @@ public:
         b_.commit(
             (this->pptr() - this->pbase()) *
             sizeof(CharT));
+        this->setp(nullptr, nullptr);
         return 0;
     }
 };
@@ -169,6 +170,7 @@ public:
         b_.commit(
             (this->pptr() - this->pbase()) *
             sizeof(CharT));
+        this->setp(nullptr, nullptr);
         return 0;
     }
 };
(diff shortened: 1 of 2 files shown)


madmongo1:
Nice, thank you.

RedDwarf69:
I took a quick look at the AppVeyor output but couldn't see the error.
Is this a problem or an "ignore it, AppVeyor has been failing for ages" situation?

madmongo1:
> Build execution time has reached the maximum allowed time for your plan (60 minutes).

Probably a question for @vinniefalco

vinniefalco:
We could remove a matrix item from Appveyor and make sure it is covered on Azure?

PR #1861, opened 2020-02-26
Deprecate mutable data type

madmongo1:
Refactor dynamic buffers to ensure conformance with DynamicBuffers_v1 interface

### DIFF
--- a/include/boost/beast/core/impl/buffers_adaptor.hpp
+++ b/include/boost/beast/core/impl/buffers_adaptor.hpp
@@ -11,6 +11,7 @@
 #define BOOST_BEAST_IMPL_BUFFERS_ADAPTOR_HPP
 
 #include <boost/beast/core/buffer_traits.hpp>
+#include <boost/beast/core/buffers_adaptor.hpp>
 #include <boost/asio/buffer.hpp>
 #include <boost/config/workaround.hpp>
 #include <boost/throw_exception.hpp>
@@ -34,71 +35,90 @@ namespace beast {
 
 template<class MutableBufferSequence>
 template<bool isMutable>
-class buffers_adaptor<MutableBufferSequence>::
-    readable_bytes
+class buffers_adaptor<MutableBufferSequence>::subrange
 {
-    buffers_adaptor const* b_;
-
 public:
-    using value_type = typename
-        std::conditional<isMutable,
-            net::mutable_buffer,
-            net::const_buffer>::type;
+    using value_type = typename std::conditional<
+        isMutable,
+        net::mutable_buffer,
+        net::const_buffer>::type;
 
-    class const_iterator;
+    struct iterator;
 
-    readable_bytes() = delete;
+    // construct from two iterators plus optionally subrange definition
+    subrange(
+        iter_type first,        // iterator to first buffer in storage
+        iter_type last,         // iterator to last buffer in storage
+        std::size_t pos = 0,    // the offset in bytes from the beginning of the storage
+        std::size_t n =         // the total length of the subrange
+        (std::numeric_limits<std::size_t>::max)());
 
 #if BOOST_WORKAROUND(BOOST_MSVC, < 1910)
-    readable_bytes(
-        readable_bytes const& other)
-        : b_(other.b_)
+    subrange(
+        subrange const& other)
+        : first_(other.first_)
+        , last_(other.last_)
+        , first_offset_(other.first_offset_)
+        , last_size_(other.last_size_)
     {
     }
 
-    readable_bytes& operator=(
-        readable_bytes const& other)
+    subrange& operator=(
+        subrange const& other)
     {
-        b_ = other.b_;
+        first_ = other.first_;
+        last_ = other.last_;
+        first_offset_ = other.first_offset_;
+        last_size_ = other.last_size_;
         return *this;
     }
 #else
-    readable_bytes(
-        readable_bytes const&) = default;
-    readable_bytes& operator=(
-        readable_bytes const&) = default;
+    subrange(
+        subrange const&) = default;
+    subrange& operator=(
+        subrange const&) = default;
 #endif
 
-    template<bool isMutable_ = isMutable, class =
-        typename std::enable_if<! isMutable_>::type>
-    readable_bytes(
-        readable_bytes<true> const& other) noexcept
-        : b_(other.b_)
+    // allow conversion from mutable to const
+    template<bool isMutable_ = isMutable, typename
+        std::enable_if<!isMutable_>::type * = nullptr>
+    subrange(subrange<true> const &other)
+        : first_(other.first_)
+        , last_(other.last_)
+        , first_offset_(other.first_offset_)
+        , last_size_(other.last_size_)
     {
     }
 
-    template<bool isMutable_ = isMutable, class =
-        typename std::enable_if<! isMutable_>::type>
-    readable_bytes& operator=(
-        readable_bytes<true> const& other) noexcept
-    {
-        b_ = other.b_;
-        return *this;
-    }
+    struct iterator;
 
-    const_iterator
+    iterator
     begin() const;
 
-    const_iterator
+    iterator
     end() const;
 
 private:
-    friend class buffers_adaptor;
 
-    readable_bytes(buffers_adaptor const& b)
-        : b_(&b)
-    {
-    }
+    friend subrange<!isMutable>;
+
+    void
+    adjust(
+        std::size_t pos,
+        std::size_t n);
+
+private:
+    // points to the first buffer in the sequence
+    iter_type first_;
+
+    // Points to one past the end of the underlying buffer sequence
+    iter_type last_;
+
+    // The initial offset into the first buffer
+    std::size_t first_offset_;
+
+    // how many bytes in the penultimate buffer are used (if any)
+    std::size_t last_size_;
 };
 
 #if BOOST_WORKAROUND(BOOST_MSVC, < 1910)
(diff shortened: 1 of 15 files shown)


include/boost/beast/core/detail/buffer_utils.hpp
```diff
@@ -0,0 +1,57 @@
...
+auto
+truncate(
+    Buffer &target,
+    std::size_t max_len)
+-> Buffer &
+{
+    target = Buffer(target.data(),
+                    (std::min)(max_len, target.size()));
+    return target;
+}
```
vinniefalco: This goes in `boost/beast/core/detail/buffer.hpp` and it should be two inline function overloads, no templates

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
-    end() const;
-
-private:
-    friend class buffers_adaptor;
-
-    mutable_buffers_type(
-            buffers_adaptor const& b)
-        : b_(&b)
+    struct iterator_category
+        : std::bidirectional_iterator_tag
```
vinniefalco: why not
```
using iterator_category = std::bidirectional_iterator_tag;
```
?

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
     }
 }
 
+template<class MutableBufferSequence>
+auto
+buffers_adaptor<MutableBufferSequence>::
+make_subrange(std::size_t pos, std::size_t n)
+-> subrange<true>
```
vinniefalco:
```
template<class MutableBufferSequence>
auto
buffers_adaptor<MutableBufferSequence>::
make_subrange(
    std::size_t pos, std::size_t n) ->
        subrange<true>
```

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
+        subrange const&) = default;
 #endif
 
-    template<bool isMutable_ = isMutable, class =
-        typename std::enable_if<! isMutable_>::type>
-    readable_bytes(
-        readable_bytes<true> const& other) noexcept
-        : b_(other.b_)
+    // allow conversion from mutable to const
+    template<bool thisIsMutable = isMutable,
```
vinniefalco: I usually just throw an underscore in the mix (or remove it), e.g. `isMutable_`

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
+        typename std::enable_if<!thisIsMutable>::type * = nullptr>
```
vinniefalco: I would put `typename` on the previous line, since it adds less than one bit of information. Not a showstopper though.

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
-        return b_ == other.b_ && it_ == other.it_;
-    }
-
-    bool
-    operator!=(const_iterator const& other) const
-    {
-        return !(*this == other);
-    }
+    iter_type
+    first() const;
```
vinniefalco: `get_buffer`, `first`, `last`, `first_offset`, and `last_size` should not be public, since they are not public interfaces.

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
 
 //------------------------------------------------------------------------------
 
 template<class MutableBufferSequence>
-class buffers_adaptor<MutableBufferSequence>::
-mutable_buffers_type
+template<bool isMutable>
+struct buffers_adaptor<MutableBufferSequence>::
+subrange<isMutable>::
+iterator
```
vinniefalco: Formatting
```
template<bool isMutable>
struct buffers_adaptor<MutableBufferSequence>::
    subrange<isMutable>::
    iterator
```

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
-    buffers_adaptor const* b_ = nullptr;
-
-public:
-    using value_type = net::mutable_buffer;
-    using pointer = value_type const*;
-    using reference = value_type;
+    };
+    using value_type = typename buffers_adaptor<MutableBufferSequence>::
+    template subrange<isMutable>::
+    value_type;
```
vinniefalco: Continuation lines have to be indented:
```
    using value_type = typename
        buffers_adaptor<MutableBufferSequence>:: template
        subrange<isMutable>::value_type;
```

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
+    using reference = typename std::add_lvalue_reference<value_type>::type;
+    using pointer = typename std::add_pointer<value_type>::type;
```
vinniefalco: There's no need for using `add_pointer` or `add_lvalue_reference` why did you do that?

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
     using difference_type = std::ptrdiff_t;
-    using iterator_category =
-        std::bidirectional_iterator_tag;
 
-    const_iterator() = default;
-    const_iterator(const_iterator const& other) = default;
-    const_iterator& operator=(const_iterator const& other) = default;
+    iterator(
+        typename buffers_adaptor<MutableBufferSequence>::
+        template subrange<isMutable> const *parent,
```
vinniefalco: Continuation lines should be indented
```
    iterator(typename
        buffers_adaptor<MutableBufferSequence>::template
            subrange<isMutable> const *parent,
        iter_type it);
```

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
-        return value_type{b.data(),
-            it_ == std::prev(b_->end_) ?
-                b_->out_end_ : b.size()} +
-                    (it_ == b_->out_ ? b_->out_pos_ : 0);
-    }
+    iterator &
+    operator++();
 
-    pointer
-    operator->() const = delete;
```
vinniefalco: Why did you remove this?

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
+    std::size_t
+    first_offset() const;
 
-    const_iterator&
-    operator++()
-    {
-        ++it_;
-        return *this;
```
vinniefalco: Why are the function definitions moved?

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+template<class MutableBufferSequence>
+template<bool isMutable>
+auto
+buffers_adaptor<MutableBufferSequence>::
+subrange<isMutable>::
+iterator::
+operator*() const
+-> value_type
+{
+    return parent_->get_buffer(it_);
```
vinniefalco: There's no need for `get_buffer` since it is only called from this place, might as well just put the code here.

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+get_buffer(iter_type it) const
+-> value_type
+{
+    auto result = *it;
+
+    if (it == first_)
+        result += first_offset_;
+
+    if (std::next(it) == last_)
+        detail::truncate(result, last_size_);
```
vinniefalco: There's no need to have a separate `truncate` function, since it is only called from here. And even if it wasn't, the operation is so trivial that it is not deserving of being elevated to its own function.

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
-        readable_bytes<true> const& other) noexcept
-        : b_(other.b_)
+    // allow conversion from mutable to const
+    template<bool thisIsMutable = isMutable,
+        typename std::enable_if<!thisIsMutable>::type * = nullptr>
+    subrange(subrange<true> const &other)
+        : first_(other.first())
+        , last_(other.last())
+        , first_offset_(other.first_offset())
+        , last_size_(other.last_size())
```
vinniefalco: These member functions `first()`, `last()`, `first_offset()`, and `last_size()` seem unnecessary, you could just access the data members. You will have to make the class template a friend first of course.

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -34,332 +35,162 @@ namespace beast {
...
-    friend class mutable_buffers_type;
 
-    const_iterator(buffers_adaptor const& b,
-            iter_type iter)
-        : it_(iter)
-        , b_(&b)
-    {
-    }
+    buffers_adaptor<MutableBufferSequence>::
+    subrange<isMutable> const *parent_;
```
vinniefalco: continuation lines should be indented
```
    buffers_adaptor<MutableBufferSequence>::
        subrange<isMutable> const *parent_;
```
I'm not a fan of "parent" but it isn't a showstopper. I would have just used `b_`.

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+
+template<class MutableBufferSequence>
+template<bool isMutable>
+auto
+buffers_adaptor<MutableBufferSequence>::
+subrange<isMutable>::
+adjust(
+    std::size_t pos,
+    std::size_t n)
+-> void
```
vinniefalco: We should never see `-> void`

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+{
+    if (n == 0)
+        last_ = first_;
+
+    // note: fallthrough
```
vinniefalco: cryptic

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+        if (adjust >= first_->size())
+        {
+            ++first_;
+            first_offset_ = 0;
+            pos -= adjust;
+        }
+        else
+        {
+            first_offset_ = adjust;
+            pos = 0;
```
vinniefalco: you could `break` and skip the loop condition...

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+    // note: fallthrough
+
+    if (first_ == last_)
+    {
+        first_offset_ = 0;
+        last_size_ = 0;
+        return;
+    }
+
+    auto last_element = [this](iter_type iter) {
```
vinniefalco: `auto is_last = ...`

include/boost/beast/core/impl/buffers_adaptor.hpp
```diff
@@ -601,6 +437,329 @@ consume(std::size_t n) noexcept
...
+    }
+}
+
+
+template<class MutableBufferSequence>
+template<bool isMutable>
+auto
+buffers_adaptor<MutableBufferSequence>::
+subrange<isMutable>::
+begin() const -> iterator
```
vinniefalco: I do like return types on their own line, whether auto or not
```
template<class MutableBufferSequence>
template<bool isMutable>
auto
buffers_adaptor<MutableBufferSequence>::
subrange<isMutable>::
begin() const ->
    iterator
```

test/beast/core/buffers_adaptor_subrange.cpp
```diff
@@ -0,0 +1,134 @@
...
+        std::size_t n = (std::numeric_limits<std::size_t>::max)())
+    -> typename buffers_adaptor<MutableBufferSequence>::const_buffers_type
+    {
+        return adaptor.make_subrange(pos, n);
+    }
+};
+
+namespace detail {
+
+class buffers_adaptor_subrange_test
```
vinniefalco: This all needs to go in buffers_adaptor.cpp not this new weird file which should not exist

test/beast/core/test_buffer.hpp
```diff
@@ -360,21 +360,21 @@ test_mutable_dynamic_buffer(
 
 /** Test an instance of a dynamic buffer or mutable dynamic buffer.
```
vinniefalco: There should be no such thing as "mutable dynamic buffer". All of Beast's dynamic buffers should support both `const` and non-`const` `data()`. You might need a separate commit for that.

test/beast/core/test_helpers.hpp
```diff
@@ -0,0 +1,218 @@
...
+        return iterator();
+    }
+private:
+    std::size_t start_;
+    std::size_t limit_;
+    std::size_t interval_;
+};
+
+
+} // test_heplers
```
vinniefalco: No I've decided I don't want to start adding this to every test, we need to think of a better way.

include/boost/beast/core/impl/buffers_adaptor.hpp
madmongo1: the iterator is now a subtype of subrange. It wasn't that I deleted this specifically, I just didn't write it. What is the purpose of this deleted function?

include/boost/beast/core/impl/buffers_adaptor.hpp
madmongo1: the enture class has been replaced.

include/boost/beast/core/impl/buffers_adaptor.hpp
madmongo1: This boils down to our fundamentally different views on nested classes. I disagree, but will change it.

include/boost/beast/core/impl/buffers_adaptor.hpp
madmongo1: strongly disagree.
My reasoning is that one logical operation deserves its own function and name as this aids higher level reasoning about intent.
Again, will change.

include/boost/beast/core/impl/buffers_adaptor.hpp
madmongo1: I think `b_` is un-necessarily cryptic - I speak as someone who used to be quite happy being limted to 6 characters for label names in 6809 assembler. Even then we'd try to spll wt mnngfl nms.

include/boost/beast/core/impl/buffers_adaptor.hpp
madmongo1: This would have no effect on the quality of gcc's code, and would for me make the loop more difficult to reason about as it's another exit point.

https://godbolt.org/z/yNsgjF

nevertheless will add

test/beast/core/test_buffer.hpp
madmongo1: I don't understand what you mean.
The code comment about was not written by me.

include/boost/beast/core/impl/buffers_adaptor.hpp
vinniefalco: to inform the user that the omission is intended (`operator->` is unimplementable)

test/beast/core/test_buffer.hpp
vinniefalco: Yep I know, lets discuss it

test/beast/core/buffers_adaptor.cpp
```diff
@@ -85,6 +113,69 @@ class buffers_adaptor_test : public unit_test::suite
...
+            using maybe_mutable =
+            typename std::conditional<
+                isMutable,
+                buffers_adaptor<std::vector<net::mutable_buffer>>&,
+                buffers_adaptor<std::vector<net::mutable_buffer>> const&>::type;
+
+            auto sub = buffers_adaptor_test_hook::make_subrange(static_cast<maybe_mutable>(adapter));
+            BEAST_EXPECTS(typeid(typename decltype(sub)::value_type) == typeid(value_type), func);
+
+            auto as_string = [](decltype(sub) const &sr) {
```
vinniefalco: What's wrong with calling `beast::buffers_to_string`? It is even more efficient than this function (since it reserves the space first)
https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__buffers_to_string.html

test/beast/core/buffers_adaptor.cpp
```diff
@@ -85,6 +113,69 @@ class buffers_adaptor_test : public unit_test::suite
...
+                std::string result;
+                auto first = net::buffer_sequence_begin(sr);
+                auto last = net::buffer_sequence_end(sr);
+                for (; first != last; ++first)
+                {
+                    auto buf = *first;
+                    auto f = reinterpret_cast<const char *>(buf.data());
+                    auto l = f + buf.size();
+                    result.insert(std::end(result), f, l);
```
vinniefalco: Why `std::end`? The type of `result` is known, this is not a generic context. Nothing is gained by calling `std::end` instead of just using `result.end`. Or even calling `result.append(f, l)`. I would like to understand why you chose one over the other, lets discuss.

madmongo1: I use std::end by force of habit, since it works in all contexts so I don't have to waste time reasoning about the underlying container. It's a zero cost abstraction, and I prefer free functions over member functions.
I could have used append, but it's specific to std::string and not generic across all containers, so to me it smells bad because it makes the code logic brittle (if I choose a different container type, I have to re-write logic, which makes no sense).

PR #1862, opened 2020-02-28
core: make detail::ostream_buffer final

RedDwarf69:
The Clang Static Analyzer was complaining about the fact the destructor
calls a virtual function. This silences it.


vinniefalco:
Commit message should read "detail::ostream_buffer is final"

Issue #1875, opened 2020-03-13
Static Analysis Issues


erikzenkerLogmein:
Boost: 1.71.0_1574063579
Beast: 266

Hi @vinniefalco thank you for the great work!

We are using the static analysis tool coverity which reports us the following issues:

- In `websocket/impl/write.hpp`
```
198            net::async_write(impl.stream(),
199                buffers_cat(impl.wr_fb.data(), cb_),
200                    beast::detail::bind_continuation(std::move(*this)));
201            bytes_transferred_ += clamp(fh_.len);
   	
CID 135675 (#3-4 of 4): Identical code for different branches (IDENTICAL_BRANCHES)
identical_branches: The same code is executed when the condition impl->check_stop_now(ec) is true or false, because the code in the if-then branch and after the if statement is identical. Should the if statement be removed?
202            if(impl.check_stop_now(ec))
203                goto upcall;
   	implicit_else: The code from the above if-then branch is identical to the code after the if statement.
204            goto upcall;
205        }
```

- in `websocket/impl/write.hpp`
```
263            impl.wr_fb.clear();
264            detail::write<flat_static_buffer_base>(
265                impl.wr_fb, fh_);
266            n = clamp(remain_, impl.wr_buf_size);
   	
CID 135672 (#11-12 of 12): Unchecked return value from library (CHECKED_RETURN)
9. check_return: Calling boost::asio::buffer_copy(boost::asio::mutable_buffers_1 const(boost::asio::buffer(impl.wr_buf.get(), n)), this->cb_) without checking return value. This library function may fail and return an error code. [Note: The source code implementation of the function has been overridden by a builtin model.]
267            net::buffer_copy(net::buffer(
268                impl.wr_buf.get(), n), cb_);
269            detail::mask_inplace(net::buffer(
```

- in `websocket/impl/read.hpp`
```
602        upcall:
603            impl.rd_block.try_unlock(this);
   	
CID 135670 (#5-8 of 8): Unchecked return value (CHECKED_RETURN)
24. check_return: Calling maybe_invoke without checking return value (as is done elsewhere 10 out of 12 times).
604            impl.op_r_close.maybe_invoke();
605            if(impl.wr_block.try_unlock(this))
606                impl.op_close.maybe_invoke()
607                    || impl.op_idle_ping.maybe_invoke()
608                    || impl.op_ping.maybe_invoke()
609                    || impl.op_wr.maybe_invoke();
610            this->complete(cont, ec, bytes_written_);
```
[...]

PR #1876, opened 2020-03-13
Ensure basic_stream::close will not throw

madmongo1:
refs: #1875

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version XXX:
+
+* Ensure basic_stream::close will not throw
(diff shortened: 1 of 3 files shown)


vinniefalco:
Suggested commit message:
"close is noexcept"

madmongo1:
> Suggested commit message:
> "close is noexcept"

This is not currently true.

I have not modified the exception specification of the public interface, only the implementation.

Do we want to also update the contract of the public interface to be incompatible with asio objects?

There is probably no return from that decision if we change our minds later.

PR #1881, opened 2020-03-18
Candidate version 288

madmongo1:


include/boost/beast/http/impl/basic_parser.hpp
```diff
@@ -56,6 +56,18 @@ put(ConstBufferSequence const& buffers,
         buf_.get(), size}, ec);
 }
 
+template<bool isRequest>
+boost::optional<std::uint64_t>
+basic_parser<isRequest>::
+impl_content_length() const
+{
+    // don't pessimise possible NRVO
+    boost::optional<std::uint64_t> result;
```
vinniefalco: What is this nonsense of NRVO? `boost::optional<std::uint64_t>` is a trivial type...

test/beast/http/parser.cpp
```diff
@@ -357,6 +357,129 @@ class parser_test
...
+            const char header[] =
+                "HTTP/1.1 200 OK\r\n"
+                "Server: nginx\r\n"
+                "Date: Wed, 11 Mar 2020 17:18:19 GMT\r\n"
+                "Content-Type: application/octet-stream\r\n"
+                "Content-Length: 0, 0, 1\r\n"
+                "Connection: keep-alive\r\n"
+                "Access-Control-Allow-Origin: *\r\n"
+                "Access-Control-Allow-Methods: GET, OPTIONS, POST\r\n"
+                "Access-Control-Allow-Headers: Authorization, Content-Type, Accept\r\n"
```
vinniefalco: Why is it necessary to have all these extra fields like `Access-Control-Allow-Origin`? Do you think it makes the test better?

test/beast/http/parser.cpp
```diff
@@ -357,6 +357,129 @@ class parser_test
...
+                "Access-Control-Max-Age: 1728000\r\n"
+                "Access-Control-Allow-Credentials: true\r\n"
+                "Content-Length: 0\r\n"
+                "Content-Type: text/plain\r\n"
+                "\r\n";
+
+            boost::beast::http::response_parser<boost::beast::http::string_body> parser;
```
vinniefalco: Why are all these names fully qualified??

test/beast/http/parser.cpp
```diff
@@ -357,6 +357,129 @@ class parser_test
...
+            boost::beast::error_code ec;
+            parser.put(boost::asio::buffer(header, sizeof(header) - 1), ec);
```
vinniefalco: Why aren't you using `string_view header` instead of `char const[]`?

include/boost/beast/http/impl/basic_parser.hpp
```diff
@@ -56,6 +56,16 @@ put(ConstBufferSequence const& buffers,
         buf_.get(), size}, ec);
 }
 
+template<bool isRequest>
+boost::optional<std::uint64_t>
+basic_parser<isRequest>::
+impl_content_length() const
```
vinniefalco: How is this different from the already existing function `basic_parser::content_length`?

include/boost/beast/http/impl/basic_parser.ipp
```diff
@@ -50,9 +50,7 @@ basic_parser<isRequest>::
 content_length() const
 {
     BOOST_ASSERT(is_header_done());
-    if(! (f_ & flagContentLength))
-        return boost::none;
-    return len0_;
+    return impl_content_length();
```
vinniefalco: Why?

include/boost/beast/http/impl/basic_parser.ipp
```diff
@@ -786,30 +784,46 @@ do_field(field f,
...
+            if (!parse_dec(tok, v))
+                return bad_content_length();
+            --tokens_unprocessed;
+            if (existing.has_value())
+            {
+                if (v != *existing)
+                    return bad_content_length();
+            }
+            else
+                existing = v;
```
vinniefalco: nit: if the `if` has braces then the `else` needs them too

test/beast/http/basic_parser.cpp
```diff
@@ -259,7 +259,10 @@ class basic_parser_test : public beast::unit_test::suite
             n = p.put(cb, ec);
             if(! ec)
                 p.put_eof(ec);
-            BEAST_EXPECTS(ec == result, ec.message());
+            if(!BEAST_EXPECTS(ec == result, ec.message()))
+            {
+                std::abort();
```
vinniefalco: `abort`?

test/beast/http/parser.cpp
```diff
@@ -366,6 +489,7 @@ class parser_test
         testGotSome();
         testIssue818();
         testIssue1187();
+        testMultipleContentLengthIssue1880();
```
vinniefalco: What's wrong with `testIssue1880()` like all the others?

test/beast/http/parser.cpp
madmongo1: This code is copy/pasted from the MCVE we were given when the bug was reported.

test/beast/http/parser.cpp
madmongo1: This code is copy/pasted from the MCVE we were given when the bug was reported.

test/beast/http/parser.cpp
madmongo1: This code is copy/pasted from the MCVE we were given when the bug was reported.

include/boost/beast/http/impl/basic_parser.hpp
madmongo1: The public function cannot be called during parsing because the assert will fail. So I have split it up into a public function which asserts the precondition and a private one that does not.

include/boost/beast/http/impl/basic_parser.ipp
madmongo1: as above

include/boost/beast/http/impl/basic_parser.ipp
madmongo1: ok

test/beast/http/parser.cpp
madmongo1: Does it really matter? I have no idea what issue 818 is about. Whoever comes after me will at least know what is being tested, even if they don't look in github (which may be out of business by then)

include/boost/beast/http/impl/basic_parser.hpp
vinniefalco: The problem is that users can subclass `basic_parser`, so you have given `parser` a capability that ordinary users do not have access to (since `impl_content_length` is private).

test/beast/http/basic_parser.cpp
vinniefalco: put the breakpoint in `test_suite::fail`

test/beast/http/parser.cpp
vinniefalco: Yes it matters because it creates noise. If you want to explain what the test does, that's fine - put it in a comment. And I think, you should probably add a link to the relevant HTTP RFC section in the implementation, as a comment.

include/boost/beast/http/impl/basic_parser.hpp
madmongo1: The capabilities and behaviour of the public API have not changed at all.
I have merely refactored the public function’s implementation to be expressed in terms of a private function which is now used in the parsing of headers (where calling the public API would assert)

PR #1883, opened 2020-03-23
Update Release Notes

madmongo1:
For comment and reference

### DIFF
--- a/doc/qbk/06_websocket/02_handshaking.qbk
+++ b/doc/qbk/06_websocket/02_handshaking.qbk
@@ -150,6 +150,27 @@ stream:
 
 [code_websocket_2_5]
 
+[heading Subprotocols]
+
+The WebSocket protocol understands the concept of subprotocols. If the client
+is requesting one of a set of subprotocols it will set the header
+[@https://tools.ietf.org/html/rfc6455#section-11.3.4 Sec-WebSocket-Protocol]
+in the initial WebSocket Upgrade HTTP request. It is up to the server to
+parse the header and select one of the protocols to accept. The server
+indicates the selected protocol by setting the
+[@https://tools.ietf.org/html/rfc6455#section-11.3.4 Sec-WebSocket-Protocol]
+header in the accept header.
+
+This is accomplished with a
+[link beast.ref.boost__beast__websocket__stream_base__decorator `decorator`].
+
+The code that follows demonstrates how a server reads an HTTP request, identifies it
+as a WebSocket Upgrade, and then checks for a preferred matching subprotocol before
+performing the WebSocket handshake:
+
+[code_websocket_2_6]
+
+
 [/-----------------------------------------------------------------------------]
 
 [endsect]
--- a/doc/qbk/release_notes.qbk
+++ b/doc/qbk/release_notes.qbk
@@ -11,6 +11,56 @@
 
 [/-----------------------------------------------------------------------------]
 
+[heading Boost 1.73]
+
+[*API Changes]
+
+* Nested `mutable_data_type` in Beast dynamic buffers is deprecated. Affected types:
+  - `buffers_adaptor`
+  - `flat_buffer`
+  - `flat_static_buffer`
+  - `multi_buffer`
+  - `static_buffer`
+
+
+[*Changes Required]
+
+* Use nested `mutable_buffers_type` instead of `mutable_data_type`,
+  or define `BOOST_BEAST_ALLOW_DEPRECATED`
+
+[*Miscellaneous]
+
+* Update root certificates in examples
+
+[*Fixes]
+
+* Fix Content-Length parsing
+* Correct `buffer_bytes` documentation
+* Fix examples to dispatch to strand
+* Ensure `basic_stream::close` will not throw
+* Fix missing include in sha1.hpp
+* Fix ostream warning
+* Field digest is endian-independent
+* Update broken links in README
+* Fix ostream flush
+* Translate some win32 errors to net error codes
+* `flat_buffer::shrink_to_fit` is `noexcept`
+* Moved-from dynamic buffers do not clear if different allocator
+* Fix erase field
+* Fix compilation macro documentation
+* Examples use strands correctly
+* Clarify end-of-file behaviour in `File::read` docs
+* `file_body` returns `short_read` on eof during read
+* Fix bug in win32 `file_body`
+* ostream_buffer satisfies preconditions of DynamicBuffer_v1::commit
+* Add accessor function to File member of `basic_file_body`
+* Fix release build of docs
+* `file_win32` supports UTF-8 paths
+* `file_stdio` supports unicode paths
+* Fix `echo-op` test
+* `file_win32` bodies respect `http::serializer::split`
+* Fix non-msvc cmake
+
 [heading Boost 1.72]
 
 [*Examples]
--- a/test/doc/websocket_2_handshaking.cpp
+++ b/test/doc/websocket_2_handshaking.cpp
@@ -132,6 +132,96 @@ snippets()
 
     //]
     }
+
+    {
+        //[code_websocket_2_6
+
+        // a function to select the most preferred protocol from a comma-separated list
+        auto select_protocol = [](string_view offered_tokens) -> std::string
+        {
+            // tokenize the Sec-Websocket-Protocol header offered by the client
+            http::token_list offered( offered_tokens );
+
+            // an array of protocols supported by this server
+            // in descending order of preference
+            static const std::array<string_view, 3>
+                supported = {
+                "v3.my.chat",
+                "v2.my.chat",
+                "v1.my.chat"
+            };
+
+            std::string result;
+
+            for (auto proto : supported)
+            {
+                auto iter = std::find(offered.begin(), offered.end(), proto);
+                if (iter != offered.end())
+                {
+                    // we found a supported protocol in the list offered by the client
+                    result.assign(proto.begin(), proto.end());
+                    break;
+                }
+            }
+
+            return result;
+        };
+
+
+        // This buffer is required for reading HTTP messages
+        flat_buffer buffer;
+
+        // Read the HTTP request ourselves
+        http::request<http::string_body> req;
+        http::read(sock, buffer, req);
+
+        // See if it's a WebSocket upgrade request
+        if(websocket::is_upgrade(req))
+        {
+            // we store the selected protocol in a std::string here because
+            // we intend to capture it in the decorator's lambda below
+            std::string protocol =
+                select_protocol(
+                    req[http::field::sec_websocket_protocol]);
+
+            if (protocol.empty())
+            {
+                // none of our supported protocols were offered
+                http::response<http::string_body> res;
+                res.result(http::status::bad_request);
+                res.body() = "No valid sub-protocol was offered."
+                              " This server implements"
+                              " v3.my.chat,"
+                              " v2.my.chat"
+                              " and v1.my.chat";
+                http::write(sock, res);
+            }
+            else
+            {
+                // Construct the stream, transferring ownership of the socket
+                stream<tcp_stream> ws(std::move(sock));
+
+                ws.set_option(
+                    stream_base::decorator(
+                        [protocol](http::response_header<> &hdr) {
+                            hdr.set(
+                                http::field::sec_websocket_protocol,
+                                protocol);
+                        }));
+
+                // Accept the upgrade request
+                ws.accept(req);
+            }
+        }
+        else
+        {
+            // Its not a WebSocket upgrade, so
+            // handle it like a normal HTTP request.
+        }
+
+        //]
+    }
+
 }
 
 struct websocket_2_test


doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,68 @@
 
 [/-----------------------------------------------------------------------------]
 
+[heading Boost 1.73]
+
+[*API Changes]
+
+* Nested const and mutable buffer types for all
+  Beast dynamic buffers are refactored. Affected types:
+  - buffers_adapter
```
vinniefalco: need backticks for identifiers, e.g. `buffers_adaptor` (note spelling on this one)

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,68 @@
...
+[*Changes Required]
+
+* Use nested mutable_buffers_type instead of mutable_data_type,
+  or define BOOST_BEAST_ALLOW_DEPRECATED
+
+
+[*Miscellaneous]
+
+* Refactor websocket read
+* Refactor multi_buffer
```
vinniefalco: Users do not see any effect from a refactoring, thus we should exclude it from the release notes.

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,56 @@
 
 [/-----------------------------------------------------------------------------]
 
+[heading Boost 1.73]
+
+[*API Changes]
+
+* Nested mutable_data_type in Beast dynamic buffers is deprecated. Affected types:
```
vinniefalco: `mutable_data_type`

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,56 @@
...
+  or define `BOOST_BEAST_ALLOW_DEPRECATED`
+
+[*Miscellaneous]
+
+* Update root certificates in examples
+
+[*Fixes]
+
+* Fix Content-Length parsing
+* Correct buffer_bytes documentation
```
vinniefalco: `buffer_bytes`

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,56 @@
...
+* Clarify end-of-file behaviour in File::read docs
+* file_body returns short_read on eof during read
+* Fix bug in win32 file_body
+* ostream_buffer satisfies preconditions of DynamicBuffer_v1::commit
+* Add accessor function to File member of basic_file_body
+* Fix release build of docs
+* file_win32 supports UTF-8 paths
+* file_stdio supports unicode paths
+* Fix echo-op test
+* file_win32 bodies respect http::serializer::split
```
vinniefalco: backticks

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,74 @@ snippets()
...
+
+        // Read the HTTP request ourselves
+        http::request<http::string_body> req;
+        http::read(sock, buffer, req);
+
+        // See if it's a WebSocket upgrade request
+        if(websocket::is_upgrade(req))
+        {
+            // tokenize the Sec-Websocket-Protocol header offered by the client
+            http::token_list offered { req[http::field::sec_websocket_protocol] };
```
vinniefalco: why braces

doc/qbk/06_websocket/02_handshaking.qbk
```diff
@@ -150,6 +150,19 @@ stream:
 
 [code_websocket_2_5]
 
+[heading Sub Protocols]
+
+The Websocket protocol understands the concept of sub-protocols. If the client
+is requesting one of a set of sub-protocols it will set the header `Sec-Websocket-Protocol`
```
vinniefalco: It is spelled **Sec-WebSocket-Protocol** note capitalization. And it does not use a monospaced font, since it is not a C++ identifier.

doc/qbk/06_websocket/02_handshaking.qbk
```diff
@@ -150,6 +150,19 @@ stream:
 
 [code_websocket_2_5]
 
+[heading Sub Protocols]
```
vinniefalco: "Subprotocols" is one word:
https://tools.ietf.org/html/rfc6455#section-1.9

doc/qbk/06_websocket/02_handshaking.qbk
```diff
@@ -150,6 +150,19 @@ stream:
...
+
+The Websocket protocol understands the concept of sub-protocols. If the client
+is requesting one of a set of sub-protocols it will set the header `Sec-Websocket-Protocol`
+in the initial HTTP request. It is up to the server to parse the header and select one of the
```
vinniefalco: in the initial WebSocket Upgrade HTTP request

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,74 @@ snippets()
...
+                            }));
+
+                    // Accept the upgrade request
+                    ws.accept(req);
+                }
+            }
+
+            // none of our supported protocols were offered
+
+            http::response<http::string_body> resp;
```
vinniefalco: `res`

doc/qbk/06_websocket/02_handshaking.qbk
```diff
@@ -150,6 +150,19 @@ stream:
...
+protocols to accept. The server indicates the selected protocol by setting the `Sec-Websocket-Protocol`
+header in the accept header.
+
+This is accomplished by means of a [link beast.ref.boost__beast__websocket__stream_base__decorator `decorator`]:
```
vinniefalco: You need to explain what the code does. For example:

"The code that follows demonstrates how a server reads an HTTP request, identifies it as a WebSocket Upgrade, and then checks for a matching subprotocol before performing the WebSocket handshake."

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,74 @@ snippets()
...
+                            [active_protocol](http::response_header<> &hdr) {
+                                hdr.set(http::field::sec_websocket_protocol, active_protocol);
+                            }));
+
+                    // Accept the upgrade request
+                    ws.accept(req);
+                }
+            }
+
+            // none of our supported protocols were offered
```
vinniefalco: There's a bug here, you call `ws.accept` and then you try to send an HTTP response

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,74 @@ snippets()
...
+            // an array of protocols supported by this server
+            // in descending order of preference
+            static const std::array<string_view, 3>
+                supported = {
+                    "v3.my.chat",
+                    "v2.my.chat",
+                    "v1.my.chat"
+                };
+
+            for (auto proto : supported)
```
vinniefalco: I would factor out the subprotocol checks into its own function with this signature:
```
string_view
select_subprotocol( string_view subprotocols );
```

test/doc/websocket_2_handshaking.cpp
madmongo1: yup, saw it. fixed

test/doc/websocket_2_handshaking.cpp
madmongo1: wow!

doc/qbk/06_websocket/02_handshaking.qbk
```diff
@@ -150,6 +150,25 @@ stream:
 
 [code_websocket_2_5]
 
+[heading Subprotocols]
+
+The Websocket protocol understands the concept of sub-protocols. If the client
```
vinniefalco: "sub-protocols" is not hyphenated, check the RFC

vinniefalco: WebSocket is always camel cased

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,94 @@ snippets()
...
+            // an array of protocols supported by this server
+            // in descending order of preference
+            static const std::array<string_view, 3>
+                supported = {
+                "v3.my.chat",
+                "v2.my.chat",
+                "v1.my.chat"
+            };
+
+            std::string result;
```
vinniefalco: Why `std::string` instead of `string_view`?

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,94 @@ snippets()
...
+        flat_buffer buffer;
+
+        // Read the HTTP request ourselves
+        http::request<http::string_body> req;
+        http::read(sock, buffer, req);
+
+        // See if it's a WebSocket upgrade request
+        if(websocket::is_upgrade(req))
+        {
+            std::string selected_protocol =
```
vinniefalco: why not just
```
string_view subprotocol = select_subprotocol( req[http::field::sec_websocket_protocol] );
```

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,94 @@ snippets()
...
+                "v3.my.chat",
+                "v2.my.chat",
+                "v1.my.chat"
+            };
+
+            std::string result;
+
+            for (auto proto : supported)
+            {
+                auto iter = std::find(offered.begin(), offered.end(), proto);
```
vinniefalco: If subprotocol matching is case-insensitive you could use this function:
https://github.com/boostorg/beast/blob/develop/include/boost/beast/http/rfc7230.hpp#L280

You have to find out if case matters though

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,94 @@ snippets()
...
+                http::write(sock, res);
+            }
+            else
+            {
+                // Construct the stream, transferring ownership of the socket
+                stream<tcp_stream> ws(std::move(sock));
+
+                ws.set_option(
+                    stream_base::decorator(
+                        [selected_protocol](http::response_header<> &hdr) {
```
vinniefalco: This is rather annoying...A copy of `selected_protocol` exists for the lifetime of the `stream` (or until the decorator is changed). This isn't your problem but it does point out a design flaw.

test/doc/websocket_2_handshaking.cpp
vinniefalco: Is this because it has to be bound into the decorator? You really need to add comments explaining this. If there was ever a need for comment, this is an obvious case of it.

madmongo1: because the passing copies of string_views around seems like a really bad idea

test/doc/websocket_2_handshaking.cpp
madmongo1: see above

test/doc/websocket_2_handshaking.cpp
madmongo1: it's absolutely down to the implementor of the subprotocol

test/doc/websocket_2_handshaking.cpp
madmongo1: probably doesn't need to be a string in the sync version. But I'd rather have a string here otherwise I'm going to get a bunch of support emails from people using string_view objects in async initiators.

test/doc/websocket_2_handshaking.cpp
vinniefalco: Why not
```
std::string subprotocol = select_subprotocol( req[http::field::sec_websocket_protocol] );
```
then?

doc/qbk/06_websocket/02_handshaking.qbk
```diff
@@ -150,6 +150,25 @@ stream:
 
 [code_websocket_2_5]
 
+[heading Subprotocols]
+
+The WebSocket protocol understands the concept of subprotocols. If the client
+is requesting one of a set of subprotocols it will set the header
+[@https://tools.ietf.org/html/rfc6455#section-11.3.4 Sec-WebSocket-Protocol]
+in the initial WebSocket Upgrade HTTP request. It is up to the server to parse the header and select one of the
```
vinniefalco: overlong line, but not a showstopper

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,97 @@ snippets()
 
     //]
     }
+
+    {
+        //[code_websocket_2_6
+
+        // a function to select the most preferred protocol from a comma-separated list
+        auto best_protocol = [](string_view offered_tokens) -> std::string
```
vinniefalco: Prefer `select_subprotocol` over `best_protocol`

test/doc/websocket_2_handshaking.cpp
```diff
@@ -132,6 +132,97 @@ snippets()
...
+        // Read the HTTP request ourselves
+        http::request<http::string_body> req;
+        http::read(sock, buffer, req);
+
+        // See if it's a WebSocket upgrade request
+        if(websocket::is_upgrade(req))
+        {
+            // we store the selected protocol in a std::string here
+            // because we intend to capture it in the decorator's lambda
+            // below
```
vinniefalco: widow

test/doc/websocket_2_handshaking.cpp
madmongo1: my thinking was that the later line:
```
if (protocol.empty())
```

would not read as cleanly as:

```
if (selected_protocol.empty())
```

test/doc/websocket_2_handshaking.cpp
madmongo1: Perhaps I misunderstood. I put this comment in because I thought you were asking for exposition of why the protocol was being captured in a string rather than a string_view

vinniefalco: What I mean is that the word "below" is all by iteself

madmongo1: Oh I see. I was trying to avoid crossing the undefined column boundary.

PR #1905, opened 2020-04-18
Add [issue] tags in release_notes.qbk

madmongo1:


vinniefalco:
Where's the version tag? Also consider this commit message instead:

> Add issue links in release notes

madmongo1:
You mean a git tag? I don’t think we’ve covered this. 
Would you like me to add a v1.73 tag and change the commit message?
After this, I take it one of us merges to master and informs the boost channel?

vinniefalco:
> Would you like me to add a v1.73 tag and change the commit message?

I"m talking about a commit that changes the Beast version number, which should be 290 now.

vinniefalco:
The 2 commits which affect CI need to be squashed into 1

PR #1919, opened 2020-04-27
Fix async_detect_ssl for net::use_awaitable 

### LINKED ISSUE #1918: Can't use boost::beast::async_detect_ssl with coroutines TS non experimental ASIO
Beast built from f5064e0c460664fa34fdad44e2379a2309d39685 (Beast version 290).

With experimental Boost.asio coroutines TS support I used the function like this:
```
  auto token = co_await this_coro::token();
  boost::system::error_code ec;

  // This buffer is required to persist across reads
  boost::beast::flat_buffer buffer;

  using boost::asio::experimental::redirect_error;

  boost::logic::tribool result =
    co_await
      boost::beast::async_detect_ssl(
        socket,
        buffer,
        redirect_error(token, ec));
```
Now that Boost.asio (from Boost version 1.71 on) moved experimental implementation to finalized version I expected the function to be called like this:
```
bool result = // perhaps still tribool?
    co_await
      boost::beast::async_detect_ssl(
        socket,
        buffer,
        boost::asio::use_awaitable);
```
but this fails to compile.

Am I doing something wrong/what would be a temporary workaround in case this requires substantial code changes?

Thanks.

madmongo1:
fixes #1918

### DIFF
--- a/test/beast/core/detect_ssl.cpp
+++ b/test/beast/core/detect_ssl.cpp
@@ -16,6 +16,11 @@
 #include <boost/beast/core/flat_buffer.hpp>
 #include <boost/beast/core/string.hpp>
 #include <boost/core/exchange.hpp>
+#if BOOST_ASIO_HAS_CO_AWAIT
+#include <boost/asio/awaitable.hpp>
+#include <boost/asio/co_spawn.hpp>
+#include <boost/asio/use_awaitable.hpp>
+#endif
 namespace boost {
 namespace beast {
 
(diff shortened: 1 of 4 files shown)


test/beast/core/detect_ssl.cpp
```diff
@@ -166,9 +171,110 @@ class detect_ssl_test : public unit_test::suite
...
+                [](std::exception_ptr ep, bool is_ssl)
+                {
+                    BEAST_EXPECT(!ep);
+                    BEAST_EXPECT(is_ssl);
+                });
+
+            test::run(ioc);
+        }
+
+        // true
```
vinniefalco, line 203: Do we need all these extra tests? It should be enough just to check that it compiles and runs the algorithm.

test/beast/core/detect_ssl.cpp
```diff
@@ -166,9 +171,110 @@ class detect_ssl_test : public unit_test::suite
         }
     }
 
+#if BOOST_ASIO_HAS_CO_AWAIT
+    void testAsyncReadCoro()
```
vinniefalco, line 175: I think this test should just instantiate the initiating function with a `net::use_awaitable` (for example, by taking the address of it)

test/beast/core/detect_ssl.cpp
madmongo1: pdimov  17:33
the test tests exactly what it needs - the actual "customer" scenario
17:34
no sane user just takes the address of the function

test/beast/core/detect_ssl.cpp
madmongo1: pdimov  17:33
the test tests exactly what it needs - the actual "customer" scenario
17:34
no sane user just takes the address of the function

PR #1922, opened 2020-04-28
Version 291

madmongo1:


include/boost/beast/core/detail/is_invocable.hpp
```diff
@@ -12,6 +12,8 @@
 
 #include <type_traits>
 #include <utility>
+#include <boost/asio/compose.hpp>
```
vinniefalco, line 15: Use `<boost/asio/async_result.hpp>` instead?

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+Version 291:
+
+* Test websocket with use_awaitable (#1920)
```
vinniefalco, line 3: No need to include the issue numbers (especially since they aren't linked).

include/boost/beast/core/detail/is_invocable.hpp
```diff
@@ -51,6 +53,26 @@ struct is_invocable<C, R(A...)>
 };
 /** @} */
 
+template<class CompletionToken, class Signature, class = void>
+struct is_completion_token_for : std::false_type
+{
+};
+
+struct dummy_initiation
```
vinniefalco, line 61: I avoid using the word "dummy" in my code

vinniefalco:
Can we please capitalize the first letter of the commit message (assuming it does not name a type)?

CHANGELOG.md
madmongo1: I do this so that the release notes will be easier to compile. If it's ok with you I'd prefer to leave them in at least until release time. What do you think?

include/boost/beast/core/detail/is_invocable.hpp
madmongo1: alternative?

include/boost/beast/core/detail/is_invocable.hpp
madmongo1: good catch. I started with async_compose and then realised I could get away with less.

CHANGELOG.md
vinniefalco: The release notes should be compiled from the commit log not the CHANGELOG

include/boost/beast/core/detail/is_invocable.hpp
vinniefalco: `any_initiation`

PR #1929, opened 2020-04-30
Issue 1913 - Fix standalone compilation error with std::string_view

madmongo1:
* Fix standalone compilation error with std::string_view 
* Test standalone compilation with std::string_view

### DIFF
--- a/include/boost/beast/http/impl/field.ipp
+++ b/include/boost/beast/http/impl/field.ipp
@@ -11,10 +11,14 @@
 #define BOOST_BEAST_HTTP_IMPL_FIELD_IPP
 
 #include <boost/beast/http/field.hpp>
+#include <boost/assert.hpp>
 #include <algorithm>
 #include <array>
 #include <cstring>
-#include <boost/assert.hpp>
+#if BOOST_BEAST_SEPARATE_COMPILATION
+#include <iostream>
+#endif
+
 
 namespace boost {
 namespace beast {
@@ -563,6 +567,12 @@ field
 string_to_field(string_view s)
 {
     return detail::get_field_table().string_to_field(s);
+}
+
+std::ostream&
+operator<<(std::ostream& os, field f)
+{
+    return os << to_string(f);
 }
 
 } // http
--- a/test/beast/http/field_compiles.cpp
+++ b/test/beast/http/field_compiles.cpp
@@ -1,0 +1,11 @@
+//
+// Copyright (c) 2016-2019 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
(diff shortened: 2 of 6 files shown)


test/beast/http/field_compiles.cpp
```diff
@@ -0,0 +1,4 @@
+//
+// check that http/field.hpp compiles unaided
+//
+#include <boost/beast/http/field.hpp>
```
vinniefalco, line 4: `field.cpp` should accomplish this, no?

include/boost/beast/http/impl/field.ipp
```diff
@@ -15,6 +15,10 @@
 #include <array>
 #include <cstring>
 #include <boost/assert.hpp>
+#if BOOST_BEAST_SEPARATE_COMPILATION
+#include <iostream>
+#endif
```
vinniefalco, line 20: can we please reorder these to put all std includes last?

test/beast/http/field_compiles.cpp
vinniefalco: See:
https://github.com/boostorg/beast/blob/83fc0018a89863c6c2aa3cf932313664f8651e9b/test/beast/http/field.cpp#L11

madmongo1: no, because it pulls in the test stuff, which pulls in iostream.

I obviously tried this first.

include/boost/beast/http/impl/field.ipp
madmongo1: This would be touching more of the file than is relevant to the fix, but sure.

test/beast/http/field_compiles.cpp
vinniefalco, line 4: The file is missing the copyright notice and license. You can look at field.cpp and copy that

include/boost/beast/http/impl/field.ipp
```diff
@@ -11,10 +11,14 @@
 #define BOOST_BEAST_HTTP_IMPL_FIELD_IPP
 
 #include <boost/beast/http/field.hpp>
+#include <boost/assert.hpp>
 #include <algorithm>
 #include <array>
 #include <cstring>
-#include <boost/assert.hpp>
+#if BOOST_BEAST_SEPARATE_COMPILATION
+#include <iostream>
```
pdimov, line 19: `<ostream>`, not `<iostream>`.

Not sure about the ifdef; the header-only config will fail in the same way without the include.

vinniefalco: The intent is that the user must include `<ostream>` if they want to use `operator<<` on `http::field`. For header-only, the user is the user of Beast. For the compiled lib, the user is the TU used to build the lib.

pdimov: As I said, I don't see why this code won't fail in the same way then, when the user doesn't use op<<.

vinniefalco: Yes I see what you mean. `<ostream>` should be included unconditionally.

PR #1945, opened 2020-05-07
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74. When including cstdio the FILE type is in the std:: namespace.

eldiener:


eldiener:
I submitted a PR to zlib for adding Embarcadero C++ clang-based support ( https://github.com/madler/zlib/pull/490 ). The open PRs for zlib are currently at 142 going back to 2013. This is why I think that sometimes Boost needs to fix things for themselves rather than wait for some 3rd party library to do so.

PR #1957, opened 2020-05-19
Fix http read bytes_transferred

### LINKED ISSUE #1942: Fix http async_read_some
refs #1677

http::async_read_some / read_some should report the number of bytes consumed by the http parser.

At present it does not, as async_read_some is implemented in terms of async_read_until, which reports the number of bytes consumed from the stream - which may well be more than are required to complete the parse.

madmongo1:
fixes #1942

### DIFF
--- a/include/boost/beast/http/impl/read.hpp
+++ b/include/boost/beast/http/impl/read.hpp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2016-2019 Vinnie Falco (vinnie dot falco at gmail dot com)
+// Copyright (c) 2020 Richard Hodges (hodges.r@gmail.com)
 //
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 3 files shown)


include/boost/beast/http/impl/read.hpp
```diff
@@ -250,6 +144,225 @@ struct run_read_msg_op
...
+};
+
+template<class Stream, class DynamicBuffer, bool isRequest, class Condition>
+class read_op
+    : asio::coroutine
+{
+    Stream& s_;
+    DynamicBuffer& b_;
+    basic_parser<isRequest>& p_;
+    std::size_t bytes_transferred_;
```
vinniefalco, line 248: Why didn't you just initialize this in-class?

madmongo1: Old habits die hard...

Issue #1944, opened 2020-05-06
file_stdio should add missing std namespace to FILE type


spdw:
In Boost Beast `FILE` is used without `std` namespace, but the C++ header is included. Thus is does not compile for some operating systems.

```
boost/beast/core/file_stdio.hpp:29:5: error: 'FILE' does not name a type
      FILE* f_ = nullptr;
      ^
```

A change from
```
#include <cstdio>
```
to
```
#include <stdio.h>
```
would compile, but the C include is deprecated in C++. Therefore it is recommended to replace `FILE` with `std::FILE`. There seems to be 7 occurrences of `FILE`: 4 of them in `file_stdio.hpp` and 3 in the corresponding `file_stdio.ipp`.

madmongo1:
@spdw Interesting. Do you know how well supported `std::FILE` is? We're still testing against clang 3.

Would you be able to recommend a compiler/stdlib combination where this fails so I can add it to our test suite?

vinniefalco:
A fix for this is already in the works, see:
https://github.com/boostorg/beast/pull/1906
and
https://github.com/vinniefalco/beast/commits/embarcadero

spdw:
> A fix for this is already in the works, see:
> #1906
> and
> https://github.com/vinniefalco/beast/commits/embarcadero

To me it looks like the header file is adapted only, the ipp file on that branch still lacks the `std::` prefix for `FILE`.

vinniefalco:
Attn @madmongo1

PR #1962, opened 2020-05-26
Refix std::FILE in ipp

madmongo1:
refs #1944


madmongo1:
@vinniefalco Ideally I'd like you to rewrite history on develop so we don't have to issue a new version

vinniefalco:
Impossible, changing history on a main branch would create havoc. It isn't a big deal to add the new commit.

PR #1974, opened 2020-06-05
iless and iequal take part in Heterogenious Lookup

### LINKED ISSUE #1949: boost::beast::iless should define is_transparent
In order to use the `iless` predicate with `std::map` (for example), it would be very helpful to have a definition of `is_transparent` included, so that heterogenous key comparisons work.

See also https://en.cppreference.com/w/cpp/container/map/find

specifically the signatures (3) and (4).

madmongo1:
Fixes #1949

### DIFF
--- a/include/boost/beast/core/string.hpp
+++ b/include/boost/beast/core/string.hpp
@@ -33,6 +33,10 @@ iequals(
 /** A case-insensitive less predicate for strings.
 
     The case-comparison operation is defined only for low-ASCII characters.
+
+    As of C++14, containers using this class as the `Compare` type will take part
+    in heterogeneous lookup if the search term is implicitly convertible to
+    @ref string_view.
 */
 struct iless
 {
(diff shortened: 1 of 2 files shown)


include/boost/beast/core/string.hpp
```diff
@@ -33,6 +33,9 @@ iequals(
 /** A case-insensitive less predicate for strings.
 
     The case-comparison operation is defined only for low-ASCII characters.
+
+    As of c++14 containers using this class as the Compare type will take part
+    in heterogenious lookup if the search term is implicitly convertible to string_view.
```
vinniefalco, line 38: Spelling: heterogeneous
Capitalization: C++14
Punctuation: comma after C++14
"Compare" should be in back-quotes and string_view should be a `@ref`.

madmongo1: For the error in spelling I am contrite. Unforgivable.

vinniefalco: Probably shouldn't capitalize heterogeneous

vinniefalco: I would say "key" instead of "search term".

madmongo1: I see you are a proponent of the Oxford comma. The subject of bitter debates between literary experts in the UK :-)

vinniefalco: I could go either way on the comma

PR #1984, opened 2020-06-12
Version 297

madmongo1:


vinniefalco:
Spelling in commit message "requied "

include/boost/beast/core/string.hpp
```diff
@@ -56,6 +66,8 @@ struct iequal
     {
         return iequals(lhs, rhs);
     }
+
+    struct is_transparent {};
```
vinniefalco, line 70: I would use `using is_transparent = void;` if possible, since it prevents doxygen from emitting a separate class type.

madmongo1: I'll check the standard and make it so.

madmongo1:
> Spelling in commit message "requied "

Solved and also corrected in CHANGELOG

PR #1997, opened 2020-06-25
Support BOOST_ASIO_NO_TS_EXECUTORS

madmongo1:
Placeholder PR for testing fixes to the the NO_TS_EXECUTORS build mode.


include/boost/beast/core/async_base.hpp
```diff
@@ -186,7 +186,22 @@ class async_base
...
+    /** The type of executor associated with this object.
+
+    If a class derived from @ref async_base is a completion
+    handler, then the associated executor of the derived class will
+    be this type.
+*/
+    using executor_type = typename
+        net::associated_executor<
+            Handler,
+            typename detail::select_work_guard_t<Executor1>::executor_type
```
vinniefalco, line 201: This is kind of annoying, a `detail` type will show up in the docs.

include/boost/beast/core/async_base.hpp
```diff
@@ -229,7 +244,7 @@ class async_base
         Handler_&& handler,
         Executor1 const& ex1)
         : h_(std::forward<Handler_>(handler))
-        , wg1_(ex1)
+        , wg1_(detail::make_work_guard(ex1))
```
vinniefalco, line 247: Should this be called `make_tracked`?

include/boost/beast/core/async_base.hpp
madmongo1: we could replace it with a __deduced__ in the docs?

vinniefalco: You might be able to say this instead:
```
#ifdef BOOST_BEAST_DOCS
    using executor_type = typename
        net::associated_executor<
            Handler,
            __implementation_defined__>::type
#else
  ...
#endif
```

Or just

```
using executor_type = __see_below__;
```

include/boost/beast/core/async_base.hpp
madmongo1: I don't know whether the net.ts or the execution nightmare is more likely to win out in future. I called it something I felt we would both woud understand.

test/beast/core/async_base.cpp
```diff
@@ -32,6 +33,57 @@ namespace beast {
...
+struct ex1_type
+{
+
+    net::execution_context &
+    query(net::execution::context_t c) const noexcept
+    { return *reinterpret_cast<net::execution_context *>(0); }
+
+    net::execution::blocking_t
+    query(net::execution::blocking_t) const noexcept
+    { return net::execution::blocking; };
```
vinniefalco, line 46: Please, separate lines:
```
    net::execution::blocking_t
    query(net::execution::blocking_t) const noexcept
    {
        return net::execution::blocking;
    };
```

test/beast/core/async_base.cpp
```diff
@@ -32,6 +33,57 @@ namespace beast {
...
+    {}
+
+    bool
+    operator==(ex1_type const &) const noexcept
+    { return true; }
+    bool
+    operator!=(ex1_type const &) const noexcept
+    { return false; }
+};
+static_assert(net::execution::can_execute<ex1_type, net::execution::invocable_archetype>::value, "");
```
vinniefalco, line 80: `BOOST_STATIC_ASSERT`?

include/boost/beast/core/async_base.hpp
vinniefalco: yeah deduced might work, check the main.qbk for the exact spelling

include/boost/beast/core/async_base.hpp
vinniefalco: fair, and its in detail so users won't/shouldn't see it.

example/http/client/coro/http_client_coro.cpp
```diff
@@ -142,3 +144,8 @@ int main(int argc, char** argv)
 
     return EXIT_SUCCESS;
 }
+#else
+int main()
+{
+}
+#endif
```
vinniefalco, line 151: missing newline

test/doc/core_3_timeouts.cpp
```diff
@@ -283,7 +283,7 @@ void do_async_echo (basic_stream<Protocol, Executor>& stream)
 
 //]
 
-#if BOOST_BEAST_ENABLE_STACKFUL
+#if BOOST_BEAST_ENABLE_STACKFUL_TESTS
```
vinniefalco, line 286: If you leave this in, it will mess up the docs. Suggested:
```
#if defined(BOOST_BEAST_DOCS) || defined(BOOST_BEAST_ENABLE_STACKFUL_TESTS)
```

(The macro might be `BOOST_BEAST_DOXYGEN`, I can't recall)

madmongo1: I should be able to remove it completely now

PR #1999, opened 2020-06-26
Changes required to compile with new Asio executors model

madmongo1:
#refs 1994

PR #2005, opened 2020-07-03
Version 298

madmongo1:
closes #1999

### DIFF
--- a/doc/qbk/release_notes.qbk
+++ b/doc/qbk/release_notes.qbk
@@ -11,6 +11,92 @@
 
 [/-----------------------------------------------------------------------------]
 
+[heading Boost 1.74]
+
+[*API Changes]
+
+* The API to Asio has undergone changes. Please refer to the Asio release notes for details.
+
+* Beast has been updated to track and respect developer choices in the use of Asio. In particular:
+ * Define `BOOST_ASIO_NO_DEPRECATED` to disallow deprecated invocation and allocation hooks.
+ * Define `BOOST_ASIO_NO_TS_EXECUTORS` to ensure that executors conform to the
+   [@http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0443r11.html Standard Executors] proposal.
+ * Define `BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT` to select
+   [@https://cplusplus.github.io/networking-ts/draft.pdf Networking TS] style executors by default.
+   If this macro is not defined, Asio default executors will be the
+   [@http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0443r11.html Standard Executors] implementation.
+
+* [issue 1897] Parser `body_limit` is optional (API Change)
+  ['Actions Required]
+  * The signature of `basic_parser<>::body_limit(n)` has changed. It now accepts an
+    optional `std::uint64_t`. The caller may indicate that no body limit is required
+    by calling `body_limit(boost::none)``. The default limits remain in place in order
+    to maintain 'safe by default' behaviour.
+
+* [issue 1934] Remove deprecated interfaces (API Change)
+  ['Actions Required]
+  * The macro `BOOST_BEAST_NO_DEPRECATED` will no longer be noticed by Beast. The only way to
+    enable deprecated functionality is now the macro `BOOST_BEAST_ALLOW_DEPRECATED` which is
+    undefined by default. That is, all deprecated behaviour is disabled by default.
+  * The following deprecated functions have been removed:
+    - `websocket::async_accept_ex`
+    - `websocket::async_handshake_ex`
+    - `websocket::accept_ex`
+    - `websocket::handshake_ex`
+    Programs still using these names should be refactored to use the `decorator` feature and
+    the remaining handshake and accept functions.
+  * `websocket::role_type` has been removed. Users should use `beast::role_type` instead.
+  * `handler_ptr` has been removed. Users should use `net::bind_handler` and/or
+    `bind_front_handler` instead.
+  * Code that depends on `mutable_data_type` should be refactored to use
+    `mutable_buffers_type`. Classes affected are:
+    - `buffers_adaptor`
+    - `flat_buffer`
+    - `flat_static_buffer`
+    - `multi_buffer`
+    - `static_buffer`
+  * The `reset` function has been removed from `flat_static_buffer`. Use the
+    `clear` function instead.
+  * The `core/type_traits.hpp` public header has been removed and along with it
+    the type trait `is_completion_handler`. Beast uses the CompletionHandler correctness
+    checks provided by Asio. In a c++20 environment, these convert to concept checks.
+  * The error code enum `invalid_code_lenths` (sic) was a synonym of `invalid_code_lengths`.
+    Affected programs should be modified to use `invalid_code_lengths`.
+  * The file `core/buffers_adapter.hpp` has been removed along with the deprecated
+    alias typename `buffers_adapter`. Affected programs should use
+  `  core/buffers_adapator.hpp` and the type `buffers_adaptor`.
+
+* [issue 1956] Deprecate `string_param` (API Change)
+  ['Actions Required]
+  `string_param`, which was previously the argument type when setting field values
+  has been replaced by `string_view`. Because of this, it is no longer possible to
+  set message field values directly as integrals.
+  Users are requied to convert numeric arguments to a string type prior to calling
+  `fields::set` et. al.
+  Beast provides the non-allocating `to_static_string()` function for this purpose.
+  To set Content-Length field manually, call `message::content_length`.
+
+[*Fixes]
+
+* [issue 1913] Fix standalone compilation error with `std::string_view`
+* [issue 1925] [issue 1916] Fix compile errors on Visual Studio with /std:c++latest
+* [issue 1924] Fix c++20 deprecation warning in `span_body`
+* [issue 1920] Fix use `buffered_read_stream` with `use_awaitable`
+* [issue 1918] Fix `async_detect_ssl` with `use_awaitable`
+* [issue 1944] Fix `FILE` namespace qualification
+* [issue 1942] Fix http read `bytes_transferred`
+* [issue 1943] Fix `basic_stream` `expires_after`
+* [issue 1980] Fix `max` compile error
+* [issue 1949] `iless` and `iequal` take part in Heterogeneous Lookup
+
+[*Miscellaneous]
+
+* [issue 1907]  OpenSSL 1.0.2 or later is required when using SSL/TLS streams.
+  This is a requirement inherited from Boost.Asio.
+
+* Additional tests have been added to ensure correct integration with C++20
+  coroutines when avaialable.
+
 [heading Boost 1.73]
 
 [*API Changes]
--- a/example/http/client/crawl/http_crawl.cpp
+++ b/example/http/client/crawl/http_crawl.cpp
@@ -355,11 +355,12 @@ int main(int argc, char* argv[])
     // The io_context is required for all I/O
     net::io_context ioc;
 
-    // The work keeps io_context::run from returning
-    auto work = net::any_io_executor(
+    // Building a tracked executor ensures that the underlying context's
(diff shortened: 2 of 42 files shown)


doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,43 @@
...
 [/-----------------------------------------------------------------------------]
 
+[heading Boost 1.74]
+
+[*API Changes]
+
+* The API to Asio has changed significantly. Please refer to the Asio release notes for details.
+
+* Beast has been updated to track and respect developer choices in the use of Asio. In particular:
+ * To respect the definition of `BOOST_ASIO_NO_DEPRECATED` in
```
vinniefalco, line 21: Non-sterile language

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,43 @@
...
+   relation to the invocation and allocation hooks.
+ * To support the proposed C++23 standard executors implementation as
+   an additive feature.
+ * To respect the user's definition of `BOOST_ASIO_NO_TS_EXECUTORS`.
+
+[*Fixes]
+
+* [issue 1913] Fix standalone compilation error with std::string_view
```
vinniefalco, line 29: `std::string_view` in back-quotes

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,43 @@
...
+* [issue 1920] Fix use buffered_read_stream with use_awaitable
+* [issue 1918] Fix async_detect_ssl with use_awaitable
+* [issue 1944] Fix FILE namespace qualification
+* [issue 1942] Fix http read bytes_transferred
+* [issue 1943] Fix basic_stream expires_after
+* [issue 1897] Parser body_limit is optional (API Change)
+* [issue 1934] Remove deprecated interfaces (API Change)
+* [issue 1956] Deprecate string_param (API Change)
+* [issue 1980] Fix `max` compile error
+* [issue 1949] iless and iequal take part in Heterogeneous Lookup
```
vinniefalco, line 41: All identifiers should use back-quotes

include/boost/beast/core/detail/bind_default_executor.hpp
```diff
@@ -67,38 +67,49 @@ class bind_default_executor_wrapper
             h_, this->get());
     }
 
+    // The invocation hook is no longer defined because it customises behaviour
+    // without forwarding to a user's hook.
+
+#if 0
```
vinniefalco, line 73: Should we just delete the code and comment?

test/beast/core/_detail_get_io_context.cpp
```diff
@@ -32,13 +32,13 @@ class get_io_context_test : public beast::unit_test::suite
         BEAST_EXPECT(get_io_context(ioc) == &ioc);
         BEAST_EXPECT(get_io_context(ioc.get_executor()) == &ioc);
         BEAST_EXPECT(get_io_context(net::make_strand(ioc)) == &ioc);
-        BEAST_EXPECT(get_io_context(net::executor(ioc.get_executor())) == &ioc);
+        BEAST_EXPECT(get_io_context(net::any_io_executor(ioc.get_executor())) == &ioc);
 #if 0
```
vinniefalco, line 36: is it time to delete this if0 stuff?

test/beast/core/basic_stream.cpp
```diff
@@ -43,12 +43,12 @@ namespace beast {
 #if 0
 template class basic_stream<
     net::ip::tcp,
-    net::executor,
+    net::any_io_executor,
```
vinniefalco, line 46: it looks like we're fixing code that isn't compiled?

test/doc/core_1_refresher.cpp
```diff
@@ -288,7 +288,7 @@ struct associated_allocator<handler, Allocator>
 template<class Executor>
 struct associated_executor<handler, Executor>
 {
-    using type = boost::asio::executor;
+    using type = boost::asio::any_io_executor;
```
vinniefalco, line 291: `net::`

vinniefalco: Especially important as this appears in the docs

include/boost/beast/core/detail/bind_default_executor.hpp
madmongo1: sure, why not?

test/beast/core/basic_stream.cpp
madmongo1: I suspect Chris did a global search/replace

test/beast/core/_detail_get_io_context.cpp
madmongo1: I'm in favour of removing cruft - less to confuse me later.

test/beast/core/bind_handler.cpp
```diff
@@ -106,7 +118,67 @@ class bind_handler_test : public unit_test::suite
             return ex_ != other.ex_;
         }
 
-        net::io_context&
+        net::execution_context& query(net::execution::context_t c) const noexcept
+        {
+            return net::query(ex_, c);
+        }
+
+        net::execution::blocking_t query(net::execution::blocking_t) const noexcept
```
vinniefalco, line 135: The return type on its own line would make this easier to read

example/http/client/crawl/http_crawl.cpp
```diff
@@ -356,7 +356,10 @@ int main(int argc, char* argv[])
     net::io_context ioc;
 
     // The work keeps io_context::run from returning
```
vinniefalco, line 358: Comment needs adjusting

include/boost/beast/core/async_base.hpp
```diff
@@ -293,8 +304,7 @@ class async_base
     executor_type
     get_executor() const noexcept
     {
-        return net::get_associated_executor(
-            h_, wg1_.get_executor());
+        return net::get_associated_executor(h_, wg1_.get_executor());
```
vinniefalco, line 307: This was previously on two lines, you made it one

include/boost/beast/core/detail/work_guard.hpp
```diff
@@ -0,0 +1,93 @@
...
+make_work_guard(Executor const& exec) noexcept
+{
+    return select_work_guard_t<Executor>(exec);
+}
+
+}
+}
+}
+
+#endif // BOOST_BEAST_CORE_DETAIL_WORK_GUARD_HPP
```
vinniefalco, line 93: I don't bother with the trailing comment. We already know that the last `#endif` in any header file matches the include guard, the comment adds no information. And since I am lazy, I will likely forget to edit it if the name of the file ever changes. But if you think it adds value, then I don't mind.

test/beast/core/test_handler.hpp
```diff
@@ -90,6 +97,15 @@ class simple_executor
     }
 };
 
+#if defined(BOOST_ASIO_NO_TS_EXECUTORS)
+static_assert(net::execution::can_execute<simple_executor, net::execution::invocable_archetype>::value, "");
```
vinniefalco, line 101: `BOOST_STATIC_ASSERT`?

doc/qbk/release_notes.qbk
```diff
@@ -11,6 +11,43 @@
...
+
+[*Miscellaneous]
+
+* [issue 1907]  OpenSSL 1.0.2 or later is required when using SSL/TLS streams.
+  This is a requirement inherited from Boost.Asio.
+
+* Additional tests have been added to ensure correct integration with C++20
+  coroutines when avaialable.
```
Eelis, line 49: "available"

test/doc/core_snippets.cpp
```diff
@@ -38,7 +38,9 @@ namespace ssl = boost::asio::ssl;
 using tcp = net::ip::tcp;
 
 net::io_context ioc;
-auto work = net::make_work_guard(ioc);
+auto work = net::any_io_executor(
```
vinniefalco, line 41: This is weird, we need to put a pin in this and determine if it is what we really want to teach, as it shows up in the docs.

test/doc/snippets.ipp
```diff
@@ -17,7 +17,10 @@ using tcp     = net::ip::tcp;
 
 error_code ec;
 net::io_context ioc;
-auto work = net::make_work_guard(ioc);
+auto work = net::any_io_executor(
+    net::prefer(
+        ioc.get_executor(),
+        net::execution::outstanding_work.tracked));
```
vinniefalco, line 23: Use this instead
```
net::any_io_executor work( net::prefer(
        ioc.get_executor(),
        net::execution::outstanding_work.tracked) );
```

example/http/client/crawl/http_crawl.cpp
```diff
@@ -356,7 +356,10 @@ int main(int argc, char* argv[])
...
-    auto work = net::make_work_guard(ioc);
+    auto work = net::any_io_executor(
+        net::prefer(
```
chriskohlhoff, line 360: I have a preference for using `require` instead of `prefer` for getting the `outstanding_work.tracked` property in this particular kind of context. My work-in-progress rule of thumb is:

* Use `require` if all of the following are true for the use case:
  - The executor is a known, concrete type (an `io_context` executor type here) rather than a template parameter
  - The concrete executor is known to support the `outstanding_work.tracked` property
  - The correctness of the code in question (non-generic code; the whole program in this case) needs work to be tracked

* Use `prefer` if the following are all true:
  - The code is generic and the executor type is a template parameter, dependent type, or `any_io_executor`
  - The executor is not known or not guaranteed to support the `outstanding_work.tracked` property
  - The correctness of the code (e.g. a generic, templated algorithm or class, or a runtime-polymorphic algorithm) does not in and of itself depend on work being tracked; i.e. think of it as generic code in a self-contained, testable unit

In the above line, we require work to be tracked against the `io_context` for the program to work, so we use `require`.

vinniefalco: The old `work_guard` was much easier to use and understand, why are we getting rid of it?

chriskohlhoff: It's not officially being got rid of from Asio (yet), but it doesn't really fit the model in the same way any more. This is because `executor_work_guard` is an RAII adapter over a resource acquisition function `Executor.on_work_started()` and a resource release function `Executor.on_work_started()`. With the proposed standard executors, the executor *is* the RAII type.

I took a closer look at this example, and actually I think it might be able to do its work tracking in a different, hopefully clearer way. I've pushed two branches with some experimental refactoring for this:

https://github.com/chriskohlhoff/Beast/commit/3f7ef481251e35f9c33c592958d13df2cbce0062
https://github.com/chriskohlhoff/Beast/commit/b58fd9698b8faf8dbdea767b5d2b9c82e7a33987

There are no doubt more variations on these themes.

I particularly like the first because it reflects the fact that each worker thread represents the promise of some future delivery of aggregation work, and that promise corresponds to an obligation to represent that future delivery as outstanding work. I think this could be a better way of thinking about it in this example even when using `make_work_guard`, and seems less arbitrary than capturing outstanding work simply to stop an `io_context` from running out of things to do.

vinniefalco: I had a similar notion but I wasn't able to put it in such elegant terms, thanks!

test/doc/core_1_refresher.cpp
madmongo1: net:: won't compile here

doc/qbk/release_notes.qbk
madmongo1: documentation has been rewritten

example/http/client/crawl/http_crawl.cpp
madmongo1: @chriskohlhoff To be honest, having now had some real-world use, I am of the view that asio's work_guard object should remain. It's purpose was (and still is) explicit. Whether it is implemented in terms of properties is of no consequence. The fact that it has a singular purpose makes it easy to identify in code and easy to teach.

The reaction we've observed from colleagues when presenting property-based executor manipulation in mainline code is along the lines of "WTF! Why?"

You and I of course know why technically, but in my view it adds very little in terms of expressing intent.

PR #2008, opened 2020-07-07
Refactor crawl

madmongo1:
* Applies Chris' recommendation re work guard.
* One strand per worker rather than two.
* Fixes lockup due to scheduler race.

### DIFF
--- a/example/http/client/crawl/http_crawl.cpp
+++ b/example/http/client/crawl/http_crawl.cpp
@@ -150,6 +150,7 @@ class worker : public std::enable_shared_from_this<worker>
     };
 
     crawl_report& report_;
+    net::strand<net::io_context::executor_type> exec_;
     tcp::resolver resolver_;
     beast::tcp_stream stream_;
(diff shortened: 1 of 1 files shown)


madmongo1:
Note, I intend to squash the commits after review

example/http/client/crawl/http_crawl.cpp
```diff
@@ -150,6 +150,7 @@ class worker : public std::enable_shared_from_this<worker>
     };
 
     crawl_report& report_;
+    net::strand<net::io_context::executor_type> exec_;
```
vinniefalco, line 151: The canonical name for variables that are executors is `ex_`

vinniefalco:
If you intend to squash the commits, consider labelling them thusly in the future:

```
Fix usage message
[FOLD] Refactor executor use
[FOLD] Remove redundant ioc reference in crawl_report
[FOLD] Fix lockup in http-crawl
```

Or some variation

PR #2010, opened 2020-07-07
Version 299

madmongo1:


example/http/client/crawl/http_crawl.cpp
```diff
@@ -164,8 +163,9 @@ class worker : public std::enable_shared_from_this<worker>
         crawl_report& report,
         net::io_context& ioc)
```
vinniefalco, line 164: I think this parameter should be `net::io_context::executor_type` but we don't have to hold up the release for it.

madmongo1: I did actually consider this. We probably need to agree on a policy. I'm personally ok with encapsulating intent in executors anywhere other than `main()`

vinniefalco: I think the policy is that we should avoid passing execution contexts.

vinniefalco:
Accepted, merge when ready

PR #2012, opened 2020-07-10
Bug in ASIO (includes Dockerfile for windows)

sdarwin:
Dockerfile to run tests on Microsoft Windows platform.

You can observe an error during the tests:

```
.\boost/asio/detail/win_iocp_overlapped_ptr.hpp(145): error C2039: 'context': is not a member of 'boost::asio::io_context::basic_executor_type<std::allocator<void>,0>'
...
...failed compile-c-c++ bin.v2\libs\beast\example\advanced\server\msvc-14.2\release\asio.mode-nots\cxxstd-2a-iso\threading-multi\advanced_server.obj..
```

With 8 cpus, this occurs at around 200 targets completed.
With 4 cpus, this occurs at around 6000 targets completed.
With 2 cpus, reaches 9900 targets, and then a different error.

It might be a race condition. To deal with this, the first part of tests.bat detects number-of-processors and then caps it. Can be adjusted.

### DIFF
--- a/.dockers/windows-vs-32/Dockerfile
+++ b/.dockers/windows-vs-32/Dockerfile
@@ -1,0 +1,61 @@
+# escape=`
+
+# Use the latest Windows Server Core image with .NET Framework 4.8.
+FROM mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2019
+
+# Restore the default Windows shell for correct batch processing.
+SHELL ["cmd", "/S", "/C"]
+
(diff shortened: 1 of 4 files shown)


madmongo1:
Thank you. Will investigate post haste.

madmongo1:
It's not a race condition. There seems to be a bug in ASIO.

I have informed Chris.

This is extremely helpful. Thank you.

madmongo1:
Merged to develop.

PR #2017, opened 2020-07-14
simplify openssl on windows

sdarwin:
For your consideration.  
I don't believe this creates any new requirement on OpenSSL which wasn't there before. The goal is only to make including the libraries easier. Let me know what you think.


vinniefalco:
Don't need punctuation in the first line of commit messages

PR #2025, opened 2020-07-20
Websocket SSL teardown also tears down underlying TCP

### LINKED ISSUE #2023: Missing TCP connection teardown in Websocket when wrapped in net::ssl::stream
### Version of Beast

Boost 1.73.0
BOOST_BEAST_VERSION 290

When using WWS client + server I noticed that only the net::ssl::stream is shutdown properly by `beast` but the TCP socket is left behind. When using the same scenario without SSL, `beast` takes care of shutting down the TCP socket properly.

I'm not sure if this is an issue or it is intended to work like that, so I've decided to post here. :)

madmongo1:
fixes #2023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+* Websocket SSL `teardown` also tears down underlying TCP.
+
+API Changes:
+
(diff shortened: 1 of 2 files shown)


madmongo1:
Merged with #2037

PR #2064, opened 2020-08-25
Bugfix: on_read() need to return on error

### LINKED ISSUE #2063: buffer_ get send back even if the an error occured
https://github.com/boostorg/beast/blob/201105e66ab738bd8e027cff8c11ddd7cd5c7587/example/advanced/server-flex/advanced_server_flex.cpp#L324

buffer_ get send back even if the an error occured. 

Instead of
```
        if(ec)
            fail(ec, "read");
```

it should be 
```
        if(ec)
            return fail(ec, "read");
```

schorsch1976:
fixes #2063

### DIFF
--- a/example/advanced/server-flex/advanced_server_flex.cpp
+++ b/example/advanced/server-flex/advanced_server_flex.cpp
@@ -321,7 +321,7 @@ class websocket_session
             return;
 
         if(ec)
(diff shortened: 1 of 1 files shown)


madmongo1:
Wonderful. Thank you.

PR #2124, opened 2020-11-20
Avoid the need for floating point arithmetic

nbougalis:
The intent of the code maybe to other, non-integer growth factors (e.g. 1.5 or 2.5) but, as is, the `float` value is both unnecessary and introduces the need for a `static_cast`. Note that this could be further simplified, since `x * 2 - x` will _always_ yield `x`.

### DIFF
--- a/include/boost/beast/core/impl/multi_buffer.hpp
+++ b/include/boost/beast/core/impl/multi_buffer.hpp
@@ -912,13 +912,12 @@ prepare(size_type n) ->
         destroy(reuse);
         if(n > 0)
         {
-            auto const growth_factor = 2.0f;
+            std::size_t const growth_factor = 2;
             auto const size =
                 (std::min<std::size_t>)(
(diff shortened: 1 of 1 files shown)


madmongo1:
Thanks for submitting this PR! 

One minor pedantic point is that this will fail if the previous size is greater than half the extent of a size_t. Of course it would actually have failed in the original code too.

It might be an idea to at least check for this scenario and clamp the result of the increment.

nbougalis:
> One minor pedantic point is that this will fail if the previous size is greater than half the extent of a size_t. Of course it would actually have failed in the original code too.
> 
> It might be an idea to at least check for this scenario and clamp the result of the increment.

Thanks for the comment. I'm happy to add simple overflow detection. I think anything more than overkill, and I suspect that it's highly unlikely for this case to trigger anyways. Let me know if you want to keep it in.

PR #2148, opened 2021-01-12
Any buffers

madmongo1:


CMakeLists.txt
```diff
@@ -144,10 +144,21 @@ if (MSVC)
     include_directories (${BOOST_ROOT})
     link_libraries( bcrypt.lib )
 else()
-    set(BOOST_INCLUDEDIR ${BOOST_ROOT})
-    set(BOOST_LIBRARYDIR ${BOOST_ROOT}/stage/lib)
-    find_package(Boost COMPONENTS coroutine filesystem system REQUIRED)
-    link_libraries(Boost::coroutine Boost::filesystem Boost::system)
+    option(BEAST_ALTERNATE_PATH "" ON)
```
vinniefalco, line 147: What is this funny business?

vinniefalco:
Missing linefeed at the end of the file, and some formatting that I would change but it is not critical.

However, I don't see the part where it calculates space for the array of `value_type` ?

CMakeLists.txt
madmongo1: Helps my workflow without damaging yours

madmongo1:
Yes you’re right. Begin and end aren't enough

include/boost/beast/core/detail/any_buffers.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+    any_buffers(any_buffers const& r);
+
+    any_buffers(any_buffers && r) noexcept;
+
+    any_buffers&
+    operator=(any_buffers const& r);
+
+    any_buffers&
+    operator=(any_buffers && r);
```
vinniefalco, line 101: I don't think assignment should be supported

madmongo1: now you mention it, that probably makes sense

include/boost/beast/core/detail/impl/any_buffers.hpp
```diff
@@ -0,0 +1,58 @@
...
+auto
+any_buffers<IsConst>::construct(Buffers const& b, Alloc const& a)
+-> impl_base*
+{
+    using impl_type = impl<Buffers, Alloc>;
+    using allocator_type = typename std::allocator_traits<Alloc>::template rebind_alloc<impl_type>;
+    auto alloc = allocator_type(a);
+    using atraits = std::allocator_traits<allocator_type>;
+
+    auto p = atraits::allocate(a, sizeof(impl_type));
```
vinniefalco, line 42: `sizeof(impl_type)` needs to be something like `sizeof(impl_type) + sizeof(value_type) * std::distance(net::buffer_sequence_begin(b), net::buffer_sequence_end(b))`

madmongo1: yes, that's going in next.
Since this object never moves or copies I can actually make it variable size at construction by having the last member as a 1-array of value_type.

include/boost/beast/core/detail/impl/any_buffers.hpp
```diff
@@ -39,7 +38,13 @@ any_buffers<IsConst>::construct(Buffers const& b, Alloc const& a)
...
     using atraits = std::allocator_traits<allocator_type>;
 
-    auto p = atraits::allocate(a, sizeof(impl_type));
+    std::size_t extra = (std::min)(
+        static_cast<std::size_t>(
+            std::distance(net::buffer_sequence_begin(b),
+                          net::buffer_sequence_end(b))),
+            std::size_t(1)) - 1;
+
+    auto p = atraits::allocate(a, sizeof(impl_type) + sizeof(value_type) * extra);
```
vinniefalco, line 47: This formula calculates bytes, but the allocator type is measured in units of `sizeof(impl_type)` which means you are over-allocating by quite a lot.

include/boost/beast/core/detail/any_buffers.hpp
```diff
@@ -66,14 +66,52 @@ class any_buffers
...
         Buffers b_;
         Alloc   a_;
 
+        // Dependency: ORDER
+        // buffers_ represents a variable-size array. It must be the last data member
+        // in this class.
+        typename std::aligned_storage<
+            sizeof(value_type),
+            alignof(value_type)>::type
+                buffer_store_[1];
```
vinniefalco, line 100: There's no need for this, you have to use `this+1` to get the storage past the end of the struct, and then work off that.

include/boost/beast/core/detail/impl/any_buffers.hpp
madmongo1: oh, that's annoying.

include/boost/beast/core/detail/any_buffers.hpp
madmongo1: wouldn't that potentially have alignment issues?

include/boost/beast/core/detail/impl/any_buffers.hpp
madmongo1:
```
    std::size_t extra = (std::min)(
        static_cast<std::size_t>(
            std::distance(net::buffer_sequence_begin(b),
                          net::buffer_sequence_end(b))),
        std::size_t(1)) - 1;
    
    const auto bytes = sizeof(impl_type) + sizeof(value_type) * extra;
    const auto units = (bytes + sizeof(impl_type) - 1) / sizeof(impl_type);
    auto p = atraits::allocate(a, units);
```

include/boost/beast/core/detail/any_buffers.hpp
vinniefalco: you align the struct itself to the `value_type`. Boost.JSON has an example of it:
https://github.com/boostorg/json/blob/develop/include/boost/json/impl/array.hpp#L23

include/boost/beast/core/detail/any_buffers.hpp
```diff
@@ -0,0 +1,197 @@
...
+    {
+        impl(
+            Buffers const& b,
+            Alloc const& a)
+        : b_(b)
+        , a_(a)
+        {
+            value_type* dest = reinterpret_cast<value_type*>(&buffer_store_);
+            this->buffers = dest;
+            for(auto& buf : b_)
```
vinniefalco, line 71: This doesn't work, you have to use `net::buffer_sequence_begin` and end.

include/boost/beast/core/detail/any_buffers.hpp
```diff
@@ -0,0 +1,197 @@
...
+                buffer_store_[1];
+
+    private:
+        void
+        destroy() noexcept override;
+
+        void unwind() noexcept
+        {
+            value_type* p = this->buffers + this->size;
+            while (this->size)
```
vinniefalco, line 109: infinite loop here

include/boost/beast/core/detail/impl/any_buffers.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    std::size_t extra = (std::min)(
+        static_cast<std::size_t>(
+            std::distance(net::buffer_sequence_begin(b),
+                          net::buffer_sequence_end(b))),
+        std::size_t(1)) - 1;
+
+    const auto bytes = sizeof(impl_type) + sizeof(value_type) * extra;
+    const auto units = (bytes + sizeof(impl_type) - 1) / sizeof(impl_type);
+    auto p = atraits::allocate(a, units);
```
vinniefalco, line 51: Still wasting memory here, you want to use an allocator with the smallest type that has the alignment common to both `impl_type` and `value_type`.

include/boost/beast/core/detail/impl/any_buffers.ipp
```diff
@@ -0,0 +1,79 @@
...
+        destroy();
+    return count;
+}
+
+// --------------------------------------------------------------
+
+
+template<bool IsConst>
+any_buffers<IsConst>::any_buffers(any_buffers const& r)
+: impl_(copy(r.impl_))
```
vinniefalco, line 39: if `copy` is only called from here, I would just inline the function into the constructor

include/boost/beast/core/detail/impl/any_buffers.hpp
madmongo1: struggling to see how this would work as if we used a different allocator for the allocation and the construction, the address of the object would strictly need to be stored, as it may not be at the address returned by the allocation.

madmongo1: ok, I think I have a solution

vinniefalco: @glenfe ?

glenfe: Given one `T` and `n` lots of  `U`, you want to allocate storage for `m` lots of `max(alignof(T), alignof(U))`. It should be obvious how to compute `m`. 

The first `sizeof(T)` bytes are for the one `T`, and the offset of the first `U` can always be computed at compile time.

glenfe: For an example, see `<boost/smart_ptr/allocate_shared_array.hpp>` which allocates 1 control block of type `T` followed by `n` lots of a user type `U`   (which makes `allocate_shared<U[]>(n)` possible).

include/boost/beast/core/detail/any_buffers.hpp
```diff
@@ -24,6 +24,34 @@ namespace boost {
 namespace beast {
 namespace detail {
 
+template<class ValueType>
```
vinniefalco, line 27: I would just use `bool IsConst` instead of `ValueType`, because `ValueType` is unconstrained while `bool` accurately reflects the actual cardinality of the value.

include/boost/beast/core/detail/any_dynamic_buffer_v0_ref.hpp
```diff
@@ -0,0 +1,243 @@
...
+#include <boost/beast/core/detail/config.hpp>
+#include <boost/beast/core/detail/any_buffers.hpp>
+#include <type_traits>
+
+namespace boost {
+namespace beast {
+namespace detail {
+
+/// @brief Wraps a reference to any
+class any_dynamic_buffer_v0_ref
```
vinniefalco, line 23: How did this slip in? We want `any_buffers` as its own commit

vinniefalco:
Didn't we measure this, and it turned out to be a loss size-wise?

madmongo1:
I think we felt that the gains did not justify the effort.

vinniefalco:
This was a mess

PR #2167, opened 2021-02-18
Replace sed-based customization with XSLT import:

evanlenz:
fix boostorg/docca#40

This must be updated in parallel with the corresponding change in boostorg/docca. Otherwise, the doc build will fail.


madmongo1 (commented):
How do I get this to run through CI without a failure?
Is the docca develop branch updated?

evanlenz:
It indeed depends on the corresponding pull request in docca. So we would
need to review and approve that one first. This is a one-time
compatibility-breaking change; I have also updated Boost.JSON’s doc files
in the same way.

madmongo1:
OK, let me know when you want to update docca and I'll get this merged

PR #2191, opened 2021-03-15
Fix invalid partial_deflate_block error code.

dmorilha-twilio:
fixes the assertion which wrongly raised a`partial_deflate_block` error code when the last received deflated chunk does not fit into a single `buffers_front(cb_)` and a remainder is still pending to be inflated. Example:

```
 -> partial deflate <-
  sz.total_out: 54
 21-03-12 14:07:14.560 TRACE: on_read(const beast::error_code &, const std::size_t)
 21-03-12 14:07:14.560 DEBUG: WssTransport::onMessage (7680 bytes)
 {"body" ... {"id":"b4b2bb07-586a-48e1-8064-a9c4b7423fe9","sid":"MT760583713619794ebf5adc89e42722e4"},{"id":"c6232956-79da-4726-835e-dfb37bbd35ed","sid":                            +> "MTea63d62fc35de55234fce1b8fd19a4f8"}]}
 21-03-12 14:07:14.560 ERROR: JSON parsing failed with error "* Line 1, Column 7681
   Missing ',' or '}' in object declaration
 ```

Wireshark had no problems to inflate it, the deflated payload was 69 bytes and after inflating, it became 7734 = 7680 + 54.

### CODE REVIEW

**Comment by madmongo1** [comment]
Thank you for the commit. 
I've cancelled the test run as we're getting some false positives due to a change in the ubuntu::xenial image/packages.
I'll rebase and reshedule once we have a stable CI testbed.

**Comment by vinniefalco** [comment]
Any chance we could add a test which demonstrates the incorrect behavior (i.e. before this patch) ?

**Comment by dmorilha-twilio** [comment]
> Any chance we could add a test which demonstrates the incorrect behavior (i.e. before this patch) ?
is it possible to control the size of the output buffer?

**Comment by dmorilha-twilio** [comment]
> Any chance we could add a test which demonstrates the incorrect behavior (i.e. before this patch) ?

Added the test. Without the fix, the forth fragment generates the `partial_deflate_block` error code.

File: test/beast/websocket/read3.cpp [hunk 1]
```diff
@@ -16,6 +16,7 @@
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/strand.hpp>
 #include <boost/asio/write.hpp>
+#include <boost/algorithm/hex.hpp>
```
vinniefalco, line 19: This introduces a dependency on Boost.Algorithm. We could probably get rid of that by reformatting the string constant:

```
char const lines[] ={
    0xc1, 0x2d, 0xaa, 0x56, 0xca, ... };
```

vinniefalco:
Thanks for the patch, this is great work. I have to admit, that the permessage-deflate code is quite complex and difficult to analyze, and these changes are non-trivial. As we are just about to release a new version of Boost, I think it would be best if we merge this after the release, so it will have 4 months worth of testing.

test/beast/websocket/read3.cpp
dmorilha-twilio: done

dmorilha-twilio:
> Thanks for the patch, this is great work. I have to admit, that the permessage-deflate code is quite complex and difficult to analyze, and these changes are non-trivial. As we are just about to release a new version of Boost, I think it would be best if we merge this after the release, so it will have 4 months worth of testing.

It is really a corner case, and I must say I really like how you leveraged other parts of boost to build a powerful API. I worked with other streaming libraries and beast is the most natural for a C++ developer. I am personally happy to have had time to and being able to contribute back to the project!

The release schedule makes sense to me. We will probably retrofit it into 73 and test, do you foresee any problems porting it back?

Thank you

test/beast/websocket/read3.cpp
```diff
@@ -625,170 +624,369 @@ class read3_test : public websocket_test_suite
...
+            sbuf(
+                "\xc1\x45"
+                "\xec\xda\x31\x0d\x00\x00\x00\xc3\x20\xff\xae"
+                "\x67\xa1\xf7\x82\x0e\x68\x0b\x91\x1a\x53\x63"
+                "\x6a\x4c\x8d\xa9\x31\x35\xa6\xc6\xd4\x98\x1a"
+                "\x53\x63\x6a\x4c\x8d\xa9\x31\x35\xa6\xc6\xd4"
+                "\x98\x1a\x53\x63\x6a\x4c\x8d\xa9\x31\x35\xa6"
+                "\xc6\xd4\x98\x1a\x53\x63\x6a\x4c\x8d\xa9\xf1"
+                "\xa7\x1a\x0f"
+            ),
```
vinniefalco, line 982: beautiful

test/beast/websocket/read3.cpp
```diff
@@ -625,170 +624,369 @@ class read3_test : public websocket_test_suite
...
         };
 
-        for (const auto & line : lines) {
-            std::vector<uint8_t> out;
-            boost::algorithm::unhex(line.begin(), line.end(), std::back_inserter(out));
-            net::write(wss.next_layer(), net::const_buffer(out.data(), out.size()));
+        for (const auto & packet : packets) {
```
vinniefalco, line 985: I would have used `const packet` instead of a reference but I suppose it doesn't matter

include/boost/beast/websocket/impl/read.hpp
```diff
@@ -563,7 +563,7 @@ class stream<NextLayer, deflateSupported>::read_some_op
                     else if(impl.rd_fh.fin)
                     {
                         // append the empty block codes
-                        std::uint8_t constexpr
+                        static std::uint8_t constexpr
```
vinniefalco, line 566: I was wondering if `static` is needed. It _shouldn't_ be. But who knows. I am not very proficient with constexpr.

vinniefalco:
> do you foresee any problems porting it back?

I don't think that code has changed since then, so it should be fairly straightforward.

One thing on our agenda is to split zlib to its own Boost library and work on improvements to the tests and API.

include/boost/beast/websocket/impl/read.hpp
dmorilha-twilio: Same here, the only problem I was able to find later was since this is created inside the `else if` scope, the stack will unwind the the `sz.next_in` could potentially point to overwritten bytes, our release version with optimizations turned on was crashing with `zlib::invalid_block_length`.

madmongo1: Whether constexpr or const results in the same code: 
https://godbolt.org/z/1j9z18

It definitely has to be static otherwise it's dangling pointer.
In this case, gcc actually optimises out the initialisation of empty_block.
https://godbolt.org/z/9xd3jv

dmorilha-twilio:
thank you!

PR #2193, opened 2021-03-15
Version 311 - MSVC Fix and Extend MSVC tests

madmongo1:


sdarwin:
@madmongo1 ,  I just noticed something else:

Immediately after each line of actual testing in the drone script, add this line, so the script will exit with an error.     That is, if the script runs 3 tests, this should  appear 3 times:
```
if %errorlevel% neq 0 exit /b %errorlevel%
```

madmongo1:
> @madmongo1 , I just noticed something else:
> 
> Immediately after each line of actual testing in the drone script, add this line, so the script will exit with an error. That is, if the script runs 3 tests, this should appear 3 times:
> 
> ```
> if %errorlevel% neq 0 exit /b %errorlevel%
> ```

done on next push. thanks

include/boost/beast/core/impl/multi_buffer.hpp
```diff
@@ -916,7 +916,7 @@ prepare(size_type n) ->
             std::size_t altn = in_size_ * growth_factor;
 	    // Overflow detection:
             if(in_size_ > altn)
-                altn = std::numeric_limits<std::size_t>::max();
+                altn = (std::numeric_limits<std::size_t>::max)();
```
vinniefalco, line 919: Maybe do a search for `"max("` (and `"min("`) in all files?

vinniefalco:
This looks good, as long as there are no other occurrences

include/boost/beast/core/impl/multi_buffer.hpp
madmongo1: You must be looking at an old commit?
I have 
```
            if(in_size_ > altn)
                altn = (std::numeric_limits<std::size_t>::max)();
```

madmongo1: Ah I see. it was just a comment.

PR #2197, opened 2021-03-17
Rebase to master for 1.76.0 release

madmongo1:


madmongo1:
I have requested review because the travis CI is reporting what appear to me to be spurious errors. I believe they should not be allowed to hold up this rebase.

vinniefalco:
We still use Travis? If this is already in develop then it should be merged to master.

madmongo1:
> We still use Travis?

I'm afraid so. 

I've started work on moving all that over to github actions.

PR #2201, opened 2021-03-18
Fix missing check for error code after header is parsed

haubenmi:
The missing error check leads to completely ignoring the body limit,
as the body limit is compared to the "Content-Length" header inside
the "finish_header" method.

This is my first pull request to beast, or boost in general. Please let me 
know if it works for you, or if you prefer me opening an issue instead.
We found this issue as a regression when we wanted to upgrade from 
Boost 1.72.0 to 1.75.0

### DIFF
--- a/include/boost/beast/http/impl/basic_parser.ipp
+++ b/include/boost/beast/http/impl/basic_parser.ipp
@@ -164,6 +164,8 @@ loop:
             goto done;
         }
         finish_header(ec, is_request{});
+        if(ec)
+            goto done;
         break;
 
     case state::body0:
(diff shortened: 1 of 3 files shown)


haubenmi:
I feared as much :) I will look into how beast does its testing, and hopefully update the request over the weekend 👍

vinniefalco:
Add a function called `testIssue2201` here
https://github.com/boostorg/beast/blob/develop/test/beast/http/basic_parser.cpp#L1549

haubenmi:
I added a test case as you asked. I had to tweak the `test_parser.hpp` a little bit, as the bug only occurs because the real implementation resets the error code in the `on_body_init_impl`.
I tried to keep the diff minimal, but I think conceptually the `ec = {};` should be added to all `on_*` hooks in `test_parser.hpp`.

madmongo1:
Many thanks!

vinniefalco:
@madmongo1 Before you merge this, please edit the commits from the contributor to have the proper email address.

PR #2209, opened 2021-03-29
http: rfc7230: Support slash character '/'

anoo1:
Some http headers like Accept[1] have a format that includes a slash
character, example: MIME_type/MIME_subtype;q=factor.

Add support to treat this slash as a regular character.

1: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept


include/boost/beast/http/detail/rfc7230.ipp
```diff
@@ -80,15 +80,14 @@ char
 is_token_char(char c)
 {
     /*
-        tchar = "!" | "#" | "$" | "%" | "&" |
-                "'" | "*" | "+" | "-" | "." |
-                "^" | "_" | "`" | "|" | "~" |
+        tchar = "!" | "#" | "$" | "%" | "&" | "'" | "*" | "+" |
```
vinniefalco, line 83: This function is used by the HTTP parser. Changing it would change what is allowed in HTTP headers.

vinniefalco:
Is this change intending only to affect the algorithms in rfc7230.hpp?

anoo1:
> Is this change intending only to affect the algorithms in rfc7230.hpp?

Yeah, I want to parse the http Accept header, and the most straight forward iterator is to use ext_list, but none of the http iterators (ext_list, param_list, etc), handle the '/', instead they stop processing the string when they encounter one.

I noticed there's a "is_text()" function that's not currently being used. Would that be something that could be used during the string processing instead of relying just in "is_token_char()"?

anoo1:
Maybe the question is, could this be supported:
```
boost::beast::http::ext_list{"*/*,text/plain;q=1,text/html;q=0.8"}
```
If yes, what would be the recommended changes?
If not, which parser would you suggest, since the advantage of an ext_list is that it automatically separates the types (*/*,) from the factor (;q=1) so the Accept types can easily be compared to see which format the caller expects, without doing additional parsing like using boost::contains.

vinniefalco:
> could this be supported:

I thought I already implemented that, but if I didn't - yes, it would be nice to have. However, you cannot change `detail::is_token_char` because the HTTP parser uses that, and the RFC does not allow a slash for the places where `is_token_char` is called from the parser. You would need to make your own function, e.g. `is_ext_token_char` and use that instead.

anoo1:
> You would need to make your own function, e.g. `is_ext_token_char` and use that instead.

Make sense. Thanks, closing the PR.

PR #2217, opened 2021-04-19
Fix async_op_base typo

gummif:


madmongo1:
Thanks @gummif, much appreciated.

PR #2232, opened 2021-05-13
[Fix] Fix bug.

innerNULL:
Fix compile error.


madmongo1:
HI @innerNULL, out of interest, with which configuration waas the example not compiling?

innerNULL:
> HI @innerNULL, out of interest, with which configuration waas the example not compiling?

@madmongo1 sorry, must be my fault, the original version is OK for compiling

Issue #2231, opened 2021-05-12
Issues using Beast with Raspberry Pi 4


AnthonySabbagh:
I am working on a project that is meant to be run on a Raspberry Pi 4. I can compile and run the project on my Ubuntu system. However, after adding the websocket part of my project using Beast, I can no longer compile on the Raspberry Pi, or run a cross-compiled binary that I compile on my Ubuntu machine. I will describe the issue that occurs when I try to compile on the Raspberry Pi 4 below.

As for cross-compiling, I have set up a cross compiler mostly following this guide. I am able to cross-compile the project, but after adding the websocket feature, I get this error when running the executable: ./handover_websocket: /usr/lib/arm-linux-gnueabihf/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by ./handover_websocket)

The Pi only has  GLIBCXX_3.4.25 available. Does Beast require 3.4.26, or have I set up something wrong in my cross-compiler?

###Version of Beast

306

###Steps necessary to reproduce the problem

I am trying to create a callback as follows:
on_connect_callback_ = boost::beast::bind_front_handler(
            &WebSocket::on_connect,
            shared_from_base<WebSocket>());

The method shared_from_base is defined in the parent class of WebSocket as follows:
std::shared_ptr<Derived> shared_from_base()
{
    return std::static_pointer_cast<Derived>(shared_from_this());
}

The compiling error is the following:
In file included from include/boost/beast/core/bind_handler.hpp:14,
                 from include/boost/beast/core/async_base.hpp:14,
                 from include/boost/beast/core.hpp:15,
                 from src/WebSocketSession.h:12,
                 from src/WebSocket.h:19,
                 from src/WebSocket.cpp:13:
include/boost/beast/core/detail/bind_handler.hpp: In instantiation of ‘boost::beast::detail::bind_front_wrapper<Handler, Args>::bind_front_wrapper(Handler_&&, Args_&& ...) [with Handler_ = boost::beast::detail::bind_front_wrapper<void (WebSocket::*)(), std::shared_ptr<WebSocket> >&; Args_ = {}; Handler = void (WebSocket::*)(); Args = {std::shared_ptr<WebSocket>}]’:
[...]

PR #2234, opened 2021-05-13
Remove use of POSIX-only constant

### LINKED ISSUE #2233: ref #2231 - use of posix SSIZE_MAX
see #2231

POSIX defined SSIZE_MAX but there is no guarantee in C++ that this symbol will be defined (even on a posix build environment) by including `<limits>`.

This is causing a compile error for one of our users.

Poposed fix is to use `std::make_unsigned<std::size_t>` and the corresponding `std::numeric_limits<T>::max()` in place of `ssize_t` and `SSIZE_MAX`.

madmongo1:
fixes #2233
refs #2231

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 # ignore cmake builds
 cmake-build-*/
-
+.dockers
+.dockerignore
 # ignore b2 builds
 bin.v2/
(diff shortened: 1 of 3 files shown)


madmongo1:
merged in PR #2255

PR #2243, opened 2021-05-25
Small update to drone script

sdarwin:
A fix for Ubuntu 20.04 valgrind.  
The drone scripts for Beast still need to be streamlined (similar to a recent update done on boost.json), and review the job definitions and compilers.


madmongo1 (approved):
Oh this is great, thanks.

PR #2253, opened 2021-06-03
Tracing charts

sdarwin:
Testing out a new profiler. Don't merge.

Some time ago, @madmongo1 suggested implementing the charts explained here:
https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-profiler-for-Clang/

This has been configured with a jenkins job.    Include -ftime-trace flag:

```
using clang :   : clang++ : <stdlib>"libc++" <cxxflags>"-ftime-trace -Wno-c99-extensions" ;
```

And then compile:

```
./b2 --user-config=./user-config.jam \
  toolset=clang \
  asio.mode=dflt,nodep,nots,ts,nodep-nots,nodep-ts \
  variant=release \
  cxxstd=14 \
  -j`grep processor /proc/cpuinfo | wc -l` \
  -q \
  libs/$REPONAME/example libs/$REPONAME/test
```

which generates tracing data in .json format for many source files.

trace2html from https://chromium.googlesource.com/catapult is run on each .json file to generate a corresponding .html file.
All the resulting pages are uploaded to Amazon S3.
puppeteer (a nodejs module) takes a snapshot of one of the charts.
Jenkins then posts the snapshot image and hyperlink to the charts in the pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,3 +275,4 @@ stakeholders about its interfaces. Even if you just have questions,
 asking them in the code review or in issues provides valuable information
 that can be used to improve the library - do not hesitate, no question
 is insignificant or unimportant!
+
--- a/README.md
+++ b/README.md
@@ -276,3 +276,4 @@ asking them in the code review or in issues provides valuable information
 that can be used to improve the library - do not hesitate, no question
 is insignificant or unimportant!
 
+
--- a/README.md
+++ b/README.md
@@ -277,3 +277,4 @@ that can be used to improve the library - do not hesitate, no question
 is insignificant or unimportant!
 
 
+


sdarwin:
- Both the .json and .html versions of the tracing files will be uploaded to S3.  If you'd like to view one or the other, switch the URL in the browser window (from .html to .json)

- A lifecycle rule on AWS S3 will delete files after 6 months. 

- Around 1000 files are created, at a total size of approximately 9GB, which is fairly large.  Enough to crash Jenkins, until I increased heap memory settings.



Testing complete for now. Closing pull request.

PR #2254, opened 2021-06-04
Add a Boost-friendly subproject case to CMakeLists

pdimov:


madmongo1 (commented):
I'm surprised `cmake_minimum_required (VERSION 3.5.1)` is allowed to come after an `if()`!

Thank you for the PR

pdimov:
> I'm surprised `cmake_minimum_required (VERSION 3.5.1)` is allowed to come after an `if()`!

It's not good practice because the policies can affect the behavior of `if`, but in this case it should be fine.

Issue #2266, opened 2021-06-25
Example server certificate has incorrectly formatted subject


utsavm9:
There is an example of an OpenSSL command here:
https://github.com/boostorg/beast/blob/710cc53331f197f6f17e8c38454c09df68e43c03/example/common/server_certificate.hpp#L32-L35

I think the subject field is not set correctly set by that:

- On a recent OpenSSL (LibreSSL 2.8.3 on macOS), using that command outputs an error `Subject Attribute /C has no known NID, skipped` and generates a certificate with an empty subject.
- Checking the certificate in the `std::string cert` on that code file with: `openssl x509 -in cert.pem -text -noout`, I see `Subject: C=US, ST=CA, L=Los AngelesO=BeastCN=www.example.com`. Here, the common name seems to have become a part of the location.

## Fix
I think the command and the example certificate in the string needs to be generated using:
```bash
openssl req \
    -newkey rsa:2048 -nodes -keyout key.pem \
    -x509 -days 10000 -out cert_correct.pem \
        -subj "/C=US/ST=CA/L=Los Angeles/O=Beast/CN=www.example.com"
```
as then the subject field is set as `Subject: C=US, ST=CA, L=Los Angeles, O=Beast, CN=www.example.com`
See `-subj arg` in https://www.openssl.org/docs/man1.1.1/man1/req.html.

## Why
For local testing, I just replaced `www.example.com` with `localhost` in the command to test if my server based on https://github.com/boostorg/beast/tree/develop/example/http/server/async-ssl is working or not. But since the common name was not set right, `curl --cacert <file>` for local integration testing would not accept the HTTPS connection to a locally running server. Most resources online on this error point to using `curl --insecure`/`curl -k` to fix this, which is quite misleading.

utsavm9:
Do you think those certificates should be corrected? I can make a PR for that, using latest OpenSSL on Ubuntu.

madmongo1:
Please go ahead and create a PR. Thank you.

utsavm9:
Let me know if #2279 resolves this.

PR #2279, opened 2021-07-06
Correct common name of example server certificate

utsavm9:
Resolves https://github.com/boostorg/beast/issues/2266.

Now, testing the async-ssl server with `curl https://localhost:8000/ --cacert cert.pem` will not give a relatively cryptic error message like:
```
curl: (60) SSL: unable to obtain common name from peer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html
```
but would hint the developer to create their own certificate by showing a message like:
```
curl: (60) SSL: certificate subject name 'www.example.com' does not match target host name 'localhost'
```

## Testing
Compiled and ran the async-ssl server example code in a seperate CMake project. Could not figure out how to build the project in the `beast` repo as is, so I have also not run any tests locally.


utsavm9:
I am guessing the coverage went down because some of the errors do not occur after the change, but I am not sure. Let me know what I can do to increase the coverage, if that is needed in this PR to merge.

nancho84:
[]()

madmongo1:
I'm happy to overlook the coverage "drop". I'll get it merged today

vinniefalco:
@alandefreitas please verify the commit is correct and then merge

PR #2280, opened 2021-07-06
Fix file open with append/append_existing flag on windows

vm2mv:


vinniefalco:
Nice :) This needs a test though... @madmongo1 ?

madmongo1:
yep, working on it

PR #2295, opened 2021-08-05
Add missing header from stream_state header

laudrup:
The stream_service class in stream_state makes use of
boost::make_shared so ensure the header is included.

This is required to include <beast/test/stream.hpp> from thirdparty
code that hasn't included the make_shared header already.

### DIFF
--- a/include/boost/beast/_experimental/test/detail/stream_state.hpp
+++ b/include/boost/beast/_experimental/test/detail/stream_state.hpp
@@ -13,6 +13,7 @@
 
 #include <boost/asio/any_io_executor.hpp>
 #include <boost/beast/core/detail/service_base.hpp>
+#include <boost/smart_ptr/make_shared.hpp>
 #include <boost/smart_ptr/weak_ptr.hpp>
 
 #include <condition_variable>


madmongo1:
Thanks for this. It's actually uncovered a few missing includes in that file. However, I don't think there is a dependency on make_shared in `include/boost/beast/_experimental/test/detail/stream_state.hpp`.

Were you getting a compile error somewhere else?

laudrup:
> Thanks for this. It's actually uncovered a few missing includes in that file. However, I don't think there is a dependency on make_shared in `include/boost/beast/_experimental/test/detail/stream_state.hpp`.
> 
> Were you getting a compile error somewhere else?

You are very welcome.

It does seem like at least `boost/smart_ptr/shared_ptr.hpp` ought to be included and possibly more. I just wasn't sure.

It's actually the implementation file that uses it (`.ipp`), but it looks like headers should be included in the corresponding header file and no the implementation file.

I just found it slightly more useful to create a pull request than opening an issue. This at least fixes the issue for me :-)

madmongo1:
https://github.com/madmongo1/beast/commit/8a2bffb9fc82fc6483f2789724c7ca4da09c64aa

madmongo1:
https://github.com/madmongo1/beast/commit/5c9e8ad118460378d436191fa19fdff4371b5173

PR #2301, opened 2021-08-24
Update GitHub Actions CI file

sdarwin:
The Ubuntu 16.04 environment is scheduled to be removed from GitHub Actions in September 2021. Migrating those jobs to Docker containers or Ubuntu 18.04.

One small error appears on gcc-4.8.   In order for the tests to display "passing", I have set "supported: false" on gcc-4.8.    If you could review that error, and possibly fix it, then the job could be re-enabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,19 +16,22 @@ jobs:
 # minimum requirement is gcc 4.8
           - toolset: gcc-4.8
             cxxstd: "11"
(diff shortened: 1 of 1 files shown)


madmongo1:
It's failing to build on drone. What should be done about that? @sdarwin

sdarwin:
> It's failing to build on drone. 

Not a problem, that's only a coincidence because the main branches have newer/different drone files than this PR.

PR #2315, opened 2021-09-20
fixed an error in websocket_client_async_ssl

ShuangLiu1992:
in this example host_ string should be updated after SSL_set_tlsext_host_name just like the synced version, otherwise this would cause a handshake error


madmongo1 (approved):
Yes, this looks correct.

Issue #2313, opened 2021-09-16
Beast doesn't reuse the sliding window when activating permessage_deflate as described in the RFC7692 Specs / Streaming compression


chreniuc:
We activated `permessage_deflate` and we've noticed that the beast library doesn't behave as the RFC7692 specs say [here](https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.2), which means that it doesn't reuse the sliding window of the [LZ77](https://en.wikipedia.org/wiki/LZ77_and_LZ78#LZ77) algorithm.

**Summary of the specs:**

If a client wants to send two compressed messages with the same contents(`Hello`), and the client did not receive `client_no_context_takeover` from the server. The same thing applies to the server, when it doesn't receive `server_no_context_takeover`. If the server sends that message twice, the first one will have the following payload:

```
0xf2 0x48 0xcd 0xc9 0xc9 0x07 0x00
```

And the second message will have the following payload:

```
0xf2 0x00 0x11 0x00 0x00
```

So 2 bytes are saved.

**How beast does it:**

I've started from the async [client](https://github.com/boostorg/beast/blob/b15a5ff0e47e72ba3d4711d2514bc65749d036ae/example/websocket/client/async/websocket_client_async.cpp) and [sever](https://github.com/boostorg/beast/blob/b15a5ff0e47e72ba3d4711d2514bc65749d036ae/example/websocket/server/async/websocket_server_async.cpp) examples, with a few minor additions, the main one was activating `permessage_deflate`, both in client and server. I've created [this mini](https://github.com/chreniuc/beast_permessage_deflate_sliding_window) repository that contains the code to reproduce this problem(?).

The most important addition to the examples is [this](https://github.com/chreniuc/beast_permessage_deflate_sliding_window/commit/6045604b4c787bca976a25475e426b9088954147#diff-37416bd90f6b7e851b4e0c946f5f40c2b35a74e75f14268a0184e2122bfdfd84L125), which activates `permessage_deflate`:

```c++
boost::beast::websocket::permessage_deflate opt;
opt.client_enable = true; // for clients
opt.server_enable = true; // for servers
opt.client_no_context_takeover = false;
opt.server_no_context_takeover = false;
ws_.set_option(opt);
```

[...]

PR #2318, opened 2021-09-23
Use zlib::Flush::sync instead of zlib::Flush::full, for stream compre…

chreniuc:
…ssion

https://github.com/boostorg/beast/issues/2313

### DIFF
--- a/include/boost/beast/websocket/detail/impl_base.hpp
+++ b/include/boost/beast/websocket/detail/impl_base.hpp
@@ -137,7 +137,7 @@ struct impl_base<true>
                     return false;
                 if(zs.avail_out >= 6)
                 {
-                    zo.write(zs, zlib::Flush::full, ec);
+                    zo.write(zs, zlib::Flush::sync, ec);
                     BOOST_ASSERT(! ec);
                     // remove flush marker
                     zs.total_out -= 4;


chreniuc:
I've noticed that this failed, due to lower code coverage. But its a little bit confusing, it shows that the code coverage dropped in [this file](https://app.codecov.io/gh/boostorg/beast/compare/2318/changes)(`include/boost/beast/websocket/impl/read.hpp`). Which was not modified by me, and also that file doesn't use the `deflate()` method that was changed.

madmongo1:
This has passed all tests. Thank you for your valuable contribution. 

I would welcome a suggestion on a test I could implement to prove that this is a fix/improvement.

R

chreniuc:
The method that I've changed is [this](https://github.com/boostorg/beast/blob/9f2b0ce1db9868b9ec54ebed18bebced53e76782/include/boost/beast/websocket/detail/impl_base.hpp#L83), I tried to find it's related test file, but I only found [this](https://github.com/boostorg/beast/blob/develop/test/beast/websocket/_detail_impl_base.cpp), which is empty.
 
 I saw that `deflate()` is used in [this file](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/impl/write.hpp#L462), when `write_some_op::operator()` is called  or when [write_some](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/impl/write.hpp#L608) is called, so I think the test might be done [here](https://github.com/boostorg/beast/blob/develop/test/beast/websocket/write.cpp) somehow.
 
The test that I was thinking about if we were to only test the `deflate()` method:
 - start with `Hello` string
 - call `deflate()`, the resulted string will be 7 bytes or will have this content: `0xf2 0x48 0xcd 0xc9 0xc9 0x07 0x00`
 - call `deflate()` with the same string, and the result will be 5 bytes or will have this content: `0xf2 0x00 0x11 0x00 0x00`
 
 Which means that compressing the same string, results has two different outputs, the second time the size is smaller.
 
A better test would be to test this in the [write_some](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/impl/write.hpp#L608) method, because there we can also check that the `client_no_context_takeover` or `server_no_context_takeover` do actually work. So there would be two tests:

**Test 1: Test where the sliding window is not used between writes**

 - start with `Hello` string, role(server or client), `server_no_context_takeover` or `client_no_context_takeover` set to `true`(based on the role)
 - call `write_some` with that string, the resulted string will be 7 bytes or will have this content: `0xf2 0x48 0xcd 0xc9 0xc9 0x07 0x00`
 - call `write_some`  with the same string, and the result will be 7 bytes or will have this content: `0xf2 0x48 0xcd 0xc9 0xc9 0x07 0x00`

This is because `server_no_context_takeover` or `client_no_context_takeover` were set to true, so when [this](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/impl/write.hpp#L685) is called, it resets the compression stream, [here](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/detail/impl_base.hpp#L155).

**Test 2: Test where the sliding window is used between writes**

 - start with `Hello` string, role(server or client), `server_no_context_takeover` or `client_no_context_takeover` set to `false`(based on the role)
 - call `write_some` with that string, the resulted string will be 7 bytes or will have this content: `0xf2 0x48 0xcd 0xc9 0xc9 0x07 0x00`
 - call `write_some`  with the same string, and the result will be 5 bytes or will have this content: `0xf2 0x00 0x11 0x00 0x00`

This is because `server_no_context_takeover` or `client_no_context_takeover` were set to false, so when [this](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/impl/write.hpp#L685) is called, it won't reset the compression stream, [here](https://github.com/boostorg/beast/blob/b7344b0d501f23f763a76488826dde8c31f34b5d/include/boost/beast/websocket/detail/impl_base.hpp#L155).

Hope this helps a little.

vinniefalco:
The coverage report is a tiny bit flaky, don't worry too much about it

PR #2352, opened 2021-12-02
codecov.yml file - ignore dirs

sdarwin:


madmongo1:
Do you want this merged into develop?

sdarwin:
Yes sure, if it doesn't interfere with the boost release.

Issue #2364, opened 2021-12-18
Response not correctly set on handshake failure?


matthijs:
As already said in thet title, response doesn't seem to be correctly set when an handshake failure occurs.

### Version of Beast
```c++
#define BOOST_BEAST_VERSION 318
```

### Steps necessary to reproduce the problem
https://github.com/matthijs/websocket-handshake

I've made small change to the websocket_client_sync_ssl.cpp from the examples. The diff:
```diff
--- websocket_client_sync_ssl.cpp       2021-12-18 15:43:59.986833842 +0100
+++ ../websocket-handshake/websocket_client_sync_ssl.cpp        2021-12-18 15:48:46.847431047 +0100
@@ -13,7 +13,7 @@
 //
 //------------------------------------------------------------------------------

-#include "example/common/root_certificates.hpp"
+#include "root_certificates.hpp"

 #include <boost/beast/core.hpp>
 #include <boost/beast/ssl.hpp>
@@ -96,7 +96,14 @@
             }));

         // Perform the websocket handshake
-        ws.handshake(host, "/");
+        boost::beast::websocket::response_type response;
+        boost::system::error_code ec;
+        ws.handshake(response, host, "/401", ec);
+        std::cout << ec.message() << std::endl;
+        std::cout << response << std::endl;
+
+        if (ec)
+            return -1;

         // Send the message
         ws.write(net::buffer(std::string(text)));
```

When compiled call it with the following arguments:
```bash
% ./main httpstat.us 443 blabla
The WebSocket handshake was declined by the remote peer
HTTP/1.1 200 OK


```
I expect the above to return the 401 response from httpstat.us but it still shows the 200 response which I believe is incorrect. I am not able to inspect the response at this point.

### All relevant compiler information
Compiler: clang++-12
Boost: 1.77.0
OpenSSL: 1.1.1d

vinniefalco:
Looks like `response` is default constructed and then not set.

madmongo1:
I can't get your repo to build at the moment. Any special command line parameters I should set for cmake?

matthijs:
I use the clang-12 compiler.
```bash
export CC=clang-12
export CXX=clang++-12

git clone https://github.com/matthijs/websocket-handshake.git
mkdir websocket-handshake/build
cd websocket-handshake/build
cmake -DCMAKE_BUILD_TYPE=Debug ..
make -j$(nproc)
```
Thats all I did (oh and I use the libc++ library instead of libstdc++).

[...]

PR #2365, opened 2021-12-20
Add test for websocket handshake failure HTTP response

madmongo1:
refs #2364


madmongo1:
@vinniefalco I'd value it if you could comment on the additional wording in the documentation of websocket::stream::handshake

PR #2384, opened 2022-01-26
Improve MIME type matching

ho-229:
Use boost::unordered_map::find to match the MIME type


sehe:
Can you describe in what way the proposed implementation is improved, and why it matters.

As it stands I see negatives as well as positives. E.g. if you care about efficiency, then perhaps look at the implementations with benchmarks given here: https://stackoverflow.com/a/48235134/85371

As you can probably see, the current implementation is taking the "do no unnecessary work" and "avoid clever code" route. I'm curious what reasons you see why we should consider the switch.

ho-229:
Well, thank you for your review.

PR #2389, opened 2022-02-09
s390x architecture drone jobs

sdarwin:


sdarwin:
There might be issues with the specific commit this branch diverged from.    I will rebase and push to the pull request again.

vinniefalco:
@alandefreitas if this doesn't fail we should merge it

alandefreitas:
It does fail in the new architecture.

alandefreitas:
> the tests try to link to coroutine, which doesn't work

(Peter)

klemens-morgenstern:
Boost context supports [s390](https://github.com/boostorg/context/blob/develop/src/asm/make_s390x_sysv_elf_gas.S) - will investigate.

klemens-morgenstern:
@sdarwin it looks like boost.context needs you to explicitly add `architecture=s390x` for it to pick up the right assembly file. I could possibly hack something together, but I reckon it's better if you add that parameter the proper way.

https://github.com/boostorg/context/blob/5852b493554ba8be5459147abf226f57d6bfa74c/build/Jamfile.v2#L473

klemens-morgenstern:
Oh, that's an issue in the build script of `context` - I'll create a PR.

sdarwin:
@klemens-morgenstern nice work. Switching from clang-12 to clang-14 solved it. However the update to boostorg/context is still required, and they haven't merged it, so this pull request will still show errors.

Issue #2394, opened 2022-03-09
websocket::stream can't be reused once timed-out


DavidKeller:
###  Problem

When calling `async_handshake()` on a `websocket::stream` that previously timed-out, the `stream::impl_type::idle_counter` attribute is still equal to `1`, hence the `BOOST_ASSERT(idle_counter == 0);` from `stream::impl_type::update_timer()` aborts the whole program.

```
/home/david/.conan/data/boost/1.78.0/_/_/package/524ea35a8120baabdde02483add58d81bf541327/include/boost/beast/websocket/impl/stream_impl.hpp:417: void boost::beast::websocket::stream< <template-parameter-1-1>, <anonymous> >::impl_type::update_timer(const Executor&) [with Executor = boost::asio::any_io_executor; NextLayer = boost::asio::ssl::stream<boost::asio::basic_stream_socket<boost::asio::ip::tcp> >; bool deflateSupported = true]: Assertion `idle_counter == 0' failed.
Aborted (core dumped)
```

### Proposal

Call `stream::impl_type::reset_idle()` from both `stream::impl_type::open()` and `stream::impl_type::reset()`.

### Version of Beast

**322** (Boost 1.78)

### Steps necessary to reproduce the problem

- Compile the code with `BOOST_ASSERT` enabled (e.g. Debug)
- Connect a websocket and wait for timeout
- Try to re-connect the very same websocket

madmongo1:
This behaviour is expected.
websocket streams are not reuseable after a failure. The stream must be destroyed and recreated.

PR #2396, opened 2022-03-09
Issue 2394

DavidKeller:
Provide test case for issue #2394


madmongo1:
Confirmed that this is expected behaviour, see #2394

PR #2406, opened 2022-04-16
Use static_string from Boost.StaticString

glenfe:


vinniefalco:
if this passes tests on CI then it looks good to go.

PR #2411, opened 2022-04-21
Use span from Boost.Core

glenfe:


vinniefalco:
please squash them down to 1 commit

vinniefalco:
231 fewer lines, nice :)

sehe:
Is drone test failing on OSX (due to missing openssl) a known issue?

vinniefalco:
@sdarwin OSX?

sdarwin:
It should be fixed now.  The brew package changed. 
Added symlink `ln -s /usr/local/opt/openssl@3 /usr/local/opt/openssl`  
The following text is only for reference, since the symlink solves the problem.

```
openssl@3 is keg-only, which means it was not symlinked into /usr/local,
because macOS provides LibreSSL.

If you need to have openssl@3 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/openssl@3/bin:$PATH"' >> ~/.zshrc
For compilers to find openssl@3 you may need to set:
  export LDFLAGS="-L/usr/local/opt/openssl@3/lib"
  export CPPFLAGS="-I/usr/local/opt/openssl@3/include"
For pkg-config to find openssl@3 you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/openssl@3/lib/pkgconfig"
```

glenfe:
The only X now is Code coverage: 94.69% (changed -0.01%) .

vinniefalco:
I wouldn't worry about that

PR #2414, opened 2022-04-25
return statement to exit reading loop

oerol:
I used this example to start building my WebSocket application, however, I noticed that upon removing the echoing in the `on_read` method, the server repeatedly throws errors without exiting the method. 

You can recreate this problem by replacing `async_write` with `do_read` and cause any error (i.e. client reloads the page).

Adding a return statement here would make debugging for future changes much easier and it wouldn't harm the existing echoing function in any way.

### DIFF
--- a/example/websocket/server/async/websocket_server_async.cpp
+++ b/example/websocket/server/async/websocket_server_async.cpp
@@ -126,7 +126,7 @@ public:
             return;
 
         if(ec)
-            fail(ec, "read");
+            return fail(ec, "read");
 
         // Echo the message
         ws_.text(ws_.got_text());


vinniefalco:
This has to be squashed down to 1 commit

vinniefalco:
I suppose returning on an error is more correct

oerol:
Is there anything I can do to turn this change into one single commit?

sehe:
In general `git rebase`. @vinniefalco If you want it merged after a squash, I'm happy to do so from the command line? Merge takes the `--squash` option that does it.

vinniefalco:
I prefer a linear history, except for the ultra-rare case where master and develop truly diverge (for example to commit a release patch to master after develop has new commits).

sehe:
Merged manually 90e37ae4

Issue #2417, opened 2022-04-27
use boost::core::string_view


vinniefalco:
We should use `boost::core::string_view` instead of `boost::utility::string_view`, because it has better interoperability with `std::string_view`. However, before we make this change public we need to write a short matrix of tests that mix the three principle string views and the standard string `std::string`, `std::string_view`, `boost::utility::string_view`, `boost::core::string_view` in order to understand what compatibility problems users may experience (if any).

sehe:
I've created a whole slew of test-cases to excavate differences between various implementations. These are not up to standard for inclusion, but perhaps someone sees some useful bits?

Goggles advised: https://github.com/sehe/beast/blob/core_string_view/test/beast/core/string.cpp

Note it uses c++17 for brevity - not in the last place to be able to compare with c++17 std::string_view, but liberally using contexpr and traits that are specific to c++17.

sehe:
@vinniefalco Just realized another loose end: despite `std::string_view` compatibility, `core::string_view` still doesn't support some templated free functions, like

 - std::quoted (too bad I guess)
 - asio::buffer
 - asio::const_buffer [demo](https://compiler-explorer.com/z/TxPx4qKv8)

The latter are a shame. They circle back to your discussion about custom buffer types in boost-http somewhat.

sehe:
Status: 

 - IMO This should remain draft until https://github.com/boostorg/core/issues/110 lands, to avoid inconveniencing Beast users with breaking change in their own code using `beast::string_view`

 - @vinniefalco / @pdimov Do you see any mechanism to fix the compatiblity with std::string_view for free function templates (std::quoted, asio::buffer and possible others in thirdparty code)? Perhaps there's already a niebloid like `beast::buffer_bytes` that I'm not aware of for this?

pdimov:
What are we supposed to do about `std::quoted` or `asio::const_buffer`?

pdimov:
It has been merged to master. Next release, well that's another story.

We can't change `std::quoted`, so I'm not sure what we can possibly do about it.

vinniefalco:
What the heck is `std::quoted`

pdimov:
It's a manipulator that quotes strings on op<< and unquotes them on op>>. https://godbolt.org/z/3srYsno9K

sehe:
[...]

PR #2418, opened 2022-05-02
Issue #2417 use boost::core::string_view

sehe:
This improves inter-conversion between string_view implementations. Some observable differences for users:
 - core::string_view no longer supports the .to_string() or .clear() extensions from Utility
 - code that relied on .max_size() returning .size(), needs to be fixed to .size() instead
 - remove_suffix() and remove_prefix() were more lenient than the standard specs; be sure you don't rely on it clamping the argument to valid range
 - BOOST_NO_CXX11_EXPLICIT_CONVERSION_OPERATORS no longer suppresses conversions to std::string
 - core::string_view adds .contains() and various bugs fixed
    
This also revealed several bugs against Boost Utility, and two feature requests for Boost Core

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 331: DRAFT
+
+* Using core::string_view instead of utility::string_view
+
+--------------------------------------------------------------------------------
+
 Version 330:
 
 * Update release notes for Boost 1.79.
--- a/doc/qbk/release_notes.qbk
+++ b/doc/qbk/release_notes.qbk
@@ -11,6 +11,17 @@
 
 [/-----------------------------------------------------------------------------]
 
+[heading Boost 1.80]
+
+[*Miscellaneous]
+
+* [issue 2417] use boost::core::string_view. This improves inter-conversion between string_view implementations. Some observable differences for users:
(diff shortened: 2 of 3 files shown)


sehe:
@alandefreitas Correct, which is why I went ahead and created the PR and suggested release notes to detail any _external_ observabilities that users might encounter in their own code using Beast.

sehe:
@vinniefalco marked as ready for review since the hash_value support is in Core. If you think we should go the deprecation warning/conditional compilation route instead, let me know.

vinniefalco:
I would just change the commit message to
```
Use core string_view:

fix #2417

This improves inter-conversion...
```

sehe:
Also, turns out I was just copy pasting the bullet list formatting from a place in the release-notes where it was /also/ not rendering correctly. :derp: Hopefully now it comes out readable.

vinniefalco:
Yep that's why we have doc previews :) (Thanks Sam!)
https://2418.beastdocs.prtest.cppalliance.org/libs/beast/doc/html/beast/release_notes.html

sehe:
Finally fixed the release notes. Took the liberty to fix some of the prior broken lists as well. It took me setting up local automatic doc generation, but now we have something :)

vinniefalco:
Please merge this if it is ready

Issue #2431, opened 2022-05-22
Memory leak in http -> fast -> fields_alloc


deW1:
Allocation of buffer never gets released because ref count is decremented after the last reference check.

Ex: https://godbolt.org/z/baGd7hvoE change line 56 `--refs`.

The issue is at https://github.com/boostorg/beast/blob/develop/example/http/server/fast/fields_alloc.hpp#L59

PR #2432, opened 2022-05-22
Fix memory leak in fields_alloc

deW1:
This fixes Issue #2431


example/http/server/fast/fields_alloc.hpp
```diff
@@ -56,7 +56,7 @@ struct static_pool
     void
     destroy()
     {
-        if(refs_--)
```
vinniefalco, line 59: short and sweet

PR #2433, opened 2022-05-22
typo

landron:
The man's name is Alexandrescu


sehe:
Good catch. Thanks @landron

PR #2435, opened 2022-05-25
Update zlib to 1.2.12

JackyYin:
Greeting,

What this PR does is simply update `zlib` library from `1.2.11` to `1.2.12`, this is because there is a CVE issue affect `zlib` library before `1.2.12`.

CVE link: https://access.redhat.com/security/cve/CVE-2018-25032 

Currently `zlib` library is only used in `test/beast` and `test/bench`, I don't think this CVE will affect the beast application itself, but I think it's still a good idea to update this and avoid any related problems.


vinniefalco:
Yeah, not a bad idea I suppose although I don't like "chores" lol, not a good look :)

The bigger question is why in the hell is a 20+ year old library still discovering CVEs???

JackyYin:
Hi @vinniefalco ,
Thanks for your feedback!

> Yeah, not a bad idea I suppose although I don't like "chores" lol, not a good look :)

Actually it's a git commit convention I have been used for a long time, please see this:
https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716.

But I think maybe it's more appropriate to use `fix` because those commits include production code changes, right?

> The bigger question is why in the hell is a 20+ year old library still discovering CVEs???

Actually, in our company, we use `blackduck` to statically scan source code for possible vulnerability, It will indicate CVE in every third party library and it's regularly update.

I think there are a bunch of threat experts looking for these security problems everyday.

alandefreitas:
#2439

alandefreitas:
> But I think maybe it's more appropriate to use fix because those commits include production code changes, right?

Yes. This would be a `fix` in this terminology. Not that the project uses semantic commit messages anyway.

We should have "fix #2439" in the message body though.

JackyYin:
Hi @alandefreitas ,
Thanks for your feedback.

I will squash the two commits and add "fix #2439" to the commit message body.

alandefreitas:
> Hi @alandefreitas , Thanks for your feedback.
> 
> I will squash the two commits and add "fix #2439" to the commit message body.

No problem. 

I also had a look at the code and everything seems safe, especially since only the tests are affected.

It would be good if we also approved running the GHA workflow. Drone is already running.

JackyYin:
> It would be good if we also approved running the GHA workflow. Drone is already running.

Sure, actually I'm not quite sure about the difference between the tests being ran by Drone CI and GHA.
Seems like the Drone CI includes `codecov` test, but they both ran test cases in `test/beast`.

vinniefalco:
@alandefreitas you need to port the fix to Beast

alandefreitas:
Merged at ca824f607ccab6682b955eb91ed49753e6433069

sehe:
@alandefreitas I noticed that the Bjam/CMake changes included a number of files that were previously excluded because they were emitting a slew of C99 violation warnings (implicit function declarations)

```
test/extern/zlib-1.2.12/gzread.c|35 col 15| warning: implicit declaration of function ‘read’; did you mean ‘fread’? [-Wimplicit-function-declaration]
test/extern/zlib-1.2.12/gzread.c|649 col 11| warning: implicit declaration of function ‘close’; did you mean ‘pclose’? [-Wimplicit-function-declaration]
test/extern/zlib-1.2.12/gzwrite.c|89 col 20| warning: implicit declaration of function ‘write’; did you mean ‘fwrite’? [-Wimplicit-function-declaration]
test/extern/zlib-1.2.12/gzlib.c|14 col 17| warning: implicit declaration of function ‘lseek’; did you mean ‘fseek’? [-Wimplicit-function-declaration]
test/extern/zlib-1.2.12/gzwrite.c|673 col 9| warning: implicit declaration of function ‘close’; did you mean ‘pclose’? [-Wimplicit-function-declaration]
```

Opened a PR to address this by excluding the unused sources once again: https://github.com/boostorg/beast/pull/2452

PR #2446, opened 2022-06-06
Fix #2445 Remove mention of ptree

### LINKED ISSUE #2445: custom body example should use streaming Boost.JSON, not property tree
image

sehe:
Actually, one mention still remains, which I cannot remove [unless
DigiCert changes their G4 Trusted Root CA certificate](https://github.com/boostorg/beast/blob/develop/example/common/root_certificates.hpp#L1134:~:text=pTREE)...


vinniefalco:
This is good I suppose

PR #2451, opened 2022-06-15
Remove BOOST_BEAST_USE_STD_STRING_VIEW

### LINKED ISSUE #2363: Use boost::core::string_view
We need to deprecate `BOOST_BEAST_USE_STD_STRING_VIEW` and change the type to use `boost::core::string_view`

sehe:
Fixes #2363

### DIFF
--- a/include/boost/beast/core/string_type.hpp
+++ b/include/boost/beast/core/string_type.hpp
@@ -12,32 +12,22 @@
 
 #include <boost/beast/core/detail/config.hpp>
 
-#if defined(BOOST_BEAST_USE_STD_STRING_VIEW)
-#include <string_view>
-#else
 #include <boost/core/detail/string_view.hpp>
+#if defined(BOOST_BEAST_USE_STD_STRING_VIEW)
+#include <boost/config/pragma_message.hpp>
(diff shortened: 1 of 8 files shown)


vinniefalco:
We can't just remove it, we have to deprecate it first for at least one release

sehe:
Does that include a preprocessor warning (well, error, for portability) when it's used? Or just a deprecation notice in the release notes

vinniefalco:
it should be an error or warning unless they define `BOOST_BEAST_ALLOW_DEPRECATED` or something like that.

include/boost/beast/core/string_type.hpp
```diff
@@ -12,32 +12,18 @@
 
 #include <boost/beast/core/detail/config.hpp>
 
-#if defined(BOOST_BEAST_USE_STD_STRING_VIEW)
-#include <string_view>
-#else
 #include <boost/core/detail/string_view.hpp>
```
klemens-morgenstern, line 15:
```suggestion
#include <boost/core/detail/string_view.hpp>

#if defined(BOOST_BEAST_USE_STD_STRING_VIEW)
BOOST_HEADER_DEPRECATED("BOOST_BEAST_USE_STD_STRING_VIEW is deprecated, use BOOST_NO_CXX17_HDR_STRING_VIEW instead")
#endif
```

vinniefalco:
This needs to be squashed, with authorship preserved, and the commit message edited to reflect the semantics of the change, i.e. "std::string_view is not an option"

PR #2452, opened 2022-06-15
Exclude unused zlib sources from build

sehe:
Recent update zlib-1.2.11->1.2.12 included included a number of source
files in the Bjam/CMake changes that were previously excluded.

This leads to a bunch of C99 violation warnings (implicit function
declarations).

This commit excludes the culprits again, but leaves them actively
commented-out so the same mistake is less likely to happen by accident
in the future.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -23,27 +23,27 @@ endif()
 set (ZLIB_SOURCES
     ${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/crc32.h
     ${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/deflate.h
-    ${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/gzguts.h
+   #${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/gzguts.h
     ${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/inffast.h
     ${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/inffixed.h
     ${CMAKE_CURRENT_LIST_DIR}/extern/zlib-1.2.12/inflate.h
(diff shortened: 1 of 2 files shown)


alandefreitas (approved):

sehe:
@sdarwin Do you understand why code-coverage keeps complaining that coverage is reduced on branches that clearly don't impact that file? Is it to do with changed build matrix configs perchance? Can we "reset" it or what can we do to stop the noise?

vinniefalco:
I would have just left the commit message empty except for the first line and the "close 2452." Being lazy is often more correcterer.

sdarwin:
> @sdarwin Do you understand why code-coverage keeps complaining that coverage is reduced on branches that clearly don't impact that file? Is it to do with changed build matrix configs perchance? Can we "reset" it or what can we do to stop the noise?

@sehe from https://docs.codecov.com/docs/commit-status this looks like a plausible fix to the problem, add the following section to .codecov.yml at the end of the file.

```
coverage:
  status:
    project:
      default:
        threshold: 1%
```

vinniefalco:
Umm... that just covers up the problem :) WHY is it report a discrepancy?

sdarwin:
The coverage difference is extremely small such as .03%. This change could
be caused by adding or removing any  lines of code, affecting the total
count,  which is then used in the percentage calculation.

vinniefalco:
> This change could be caused by adding or removing any  lines of code

Yes of course, but it is often the case that a pull request does not change any lines of covered code, such as in this pull request. And yet the coverage changes.

sdarwin:
In a percentage calculation there is a numerator and a denominator.   If
"any" character changes whatsoever, the total sum of all code changes.
Regardless of "covered code".   Then, the covered is compared against the
total sum.      (unless this isn't right.   :-). Just the theory.

sdarwin:
Or, .03% is an artifact of some internal calculation that codecov is making
and it is an error.

alandefreitas:
This fixes #2462, right? We should link that.

alandefreitas:
Merged at 4f3276fd704009c0417fa1bca282f7f82acd3d1e

PR #2455, opened 2022-06-16
Remove unused method

### LINKED ISSUE #2410: basic_multi_buffer::destroy(const_iter it)
Hi,
In order to investigate the performance of beast::multi_buffer,  I read most the code of class basic_multi_buffer today, and found a small potential issue about the destroy() method of this class. In implementation of the overload destroy(const_iter it), at line around 1208 of file beast/core/impl/multi_buffer.hpp, does the list_.erase(it) returns the next iterator after the erased item, but not the reference to the erased item (i.e. the element)? If so, this method may only recursively erase the items in the list, leaving the 'element' memory not freed, which will results in memory leak. Is this behavior intended or by mistake? Fortunately, I found that this overload is never been invoked by other methods of this class.

Here is a copy of the implementation code:
```cpp
template<class Allocator>
void
basic_multi_buffer<Allocator>::
destroy(const_iter it)
{
    auto& e = list_.erase(it);
    destroy(e);
}
```

alandefreitas:
fix #2410

### DIFF
--- a/include/boost/beast/core/impl/multi_buffer.hpp
+++ b/include/boost/beast/core/impl/multi_buffer.hpp
@@ -1203,15 +1203,6 @@ destroy(list_type& list) noexcept
 template<class Allocator>
 void
 basic_multi_buffer<Allocator>::
-destroy(const_iter it)
-{
-    auto& e = list_.erase(it);
-    destroy(e);
-}
-
-template<class Allocator>
-void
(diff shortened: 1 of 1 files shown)


vinniefalco:
Good

alandefreitas:
Merged at 46729654c1978aee4c1c1d0ac0a0061b8589b831

PR #2458, opened 2022-06-18
Fix broken reference links

### LINKED ISSUE #2333: Doxygen is being passed too many folders
The _experimental/unit_test directory was never supposed to be passed to Doxygen:
https://github.com/boostorg/beast/blob/boost-1.76.0/doc/source.dox#L6

alandefreitas:
fix #2333

This PR fixes broken links in the reference removes links that shouldn't be in the reference


vinniefalco:
Good

alandefreitas:
Merged at eaa51201bdc1b801202d2d83434718c30e2d4948

PR #2464, opened 2022-06-18
Document fields::equal_range

### LINKED ISSUE #1251: Documentation Around Guaranteed Order for `http::fields::equal_range`
The `equal_range` member function internally guarantees that the range's left-to-right traversal matches the field ordering found in the message.

This is an undocumented feature which can lead to users assuming the range is unordered which would create needless routines based around being unordered.

alandefreitas:
fix #1251

### DIFF
--- a/include/boost/beast/http/fields.hpp
+++ b/include/boost/beast/http/fields.hpp
@@ -615,6 +615,14 @@ public:
 
     /** Returns a range of iterators to the fields with the specified name.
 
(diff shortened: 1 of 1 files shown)


vinniefalco:
Good

alandefreitas:
Already merged at 3272dc0aeb400e17cd39f628791c817e4b368ec5

Issue #227, opened 2017-01-18
intelligent compression


vinniefalco:
When the message or frame is small or empty we might be smarter about the compression settings (or even if we compress at all).

vinniefalco:
We aren't doing this anymore

vinniefalco:
The config for the stream could have a size threshold parameter

klemens-morgenstern:
For reference

https://www.igvita.com/2013/11/27/configuring-and-optimizing-websocket-compression/

klemens-morgenstern:
it seems this spec is already implemented: https://datatracker.ietf.org/doc/html/draft-ietf-hybi-permessage-compression

is this issue still open or do we need to refine the logic?
@vinniefalco

vinniefalco:
A simple algorithm is fine, a configurable size threshold below which, we don't compress.

ashtum:
Addressed in https://github.com/boostorg/beast/pull/2480/commits/10ad742752601ac18af0f712f255423bb6ee52ca.

PR #2480, opened 2022-07-05
per-message compression options

### LINKED ISSUE #226: allow per-message compression option
When permessage-deflate is negotiated on a stream, It should be possible to control whether compression is enabled on a per-message basis.

alandefreitas:
fix #226, #227

### DIFF
--- a/include/boost/beast/websocket/impl/write.hpp
+++ b/include/boost/beast/websocket/impl/write.hpp
@@ -86,7 +86,7 @@ public:
         // Set up the outgoing frame header
         if(! impl.wr_cont)
         {
-            impl.begin_msg();
+            impl.begin_msg(beast::buffer_bytes(bs));
             fh_.rsv1 = impl.wr_compress;
         }
         else
@@ -114,7 +114,7 @@ public:
             else
             {
                 BOOST_ASSERT(impl.wr_buf_size != 0);
-                remain_ = buffer_bytes(cb_);
+                remain_ = beast::buffer_bytes(cb_);
                 if(remain_ > impl.wr_buf_size)
                     how_ = do_nomask_frag;
                 else
@@ -130,7 +130,7 @@ public:
             else
             {
                 BOOST_ASSERT(impl.wr_buf_size != 0);
-                remain_ = buffer_bytes(cb_);
+                remain_ = beast::buffer_bytes(cb_);
                 if(remain_ > impl.wr_buf_size)
                     how_ = do_mask_frag;
                 else
@@ -207,7 +207,7 @@ operator()(
         {
             // send a single frame
             fh_.fin = fin_;
-            fh_.len = buffer_bytes(cb_);
+            fh_.len = beast::buffer_bytes(cb_);
             impl.wr_fb.clear();
             detail::write<flat_static_buffer_base>(
                 impl.wr_fb, fh_);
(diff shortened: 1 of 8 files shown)


include/boost/beast/websocket/option.hpp
```diff
@@ -55,6 +55,9 @@ struct permessage_deflate
 
     /// Deflate memory level, 1..9
     int memLevel = 4;
+
+    /// The minimum size a message should have to be compressed
+    std::size_t msg_size_threshold = 0;
```
alandefreitas, line 60: I couldn't find a reference for a reasonable default yet...

include/boost/beast/websocket/impl/write.hpp
```diff
@@ -86,7 +86,7 @@ class stream<NextLayer, deflateSupported>::write_some_op
         // Set up the outgoing frame header
         if(! impl.wr_cont)
         {
-            impl.begin_msg();
+            impl.begin_msg(buffer_bytes(bs));
```
vinniefalco, line 89: You need to inhibit ADL here, because `bs` is user-defined

include/boost/beast/websocket/impl/write.hpp
```diff
@@ -303,7 +303,7 @@ operator()(
         if(how_ == do_mask_nofrag)
         {
             // send a single frame using multiple writes
-            remain_ = beast::buffer_bytes(cb_);
+            remain_ = buffer_bytes(cb_);
```
vinniefalco, line 306: You need to inhibit ADL here

include/boost/beast/websocket/impl/write.hpp
```diff
@@ -622,7 +622,7 @@ write_some(bool fin,
     detail::frame_header fh;
     if(! impl.wr_cont)
     {
-        impl.begin_msg();
+        impl.begin_msg(buffer_bytes(buffers));
```
vinniefalco, line 625: you might need to inhibit ADL here

include/boost/beast/websocket/stream_fwd.hpp
```diff
@@ -12,18 +12,18 @@
...
 
+#ifndef BOOST_BEAST_DOXYGEN
+
 //[code_websocket_1h
 
 namespace boost {
 namespace beast {
 namespace websocket {
 
-#ifndef BOOST_BEAST_DOXYGEN
```
vinniefalco, line 21: Why did you remove the ifdef? This will probably confuse doxygen

alandefreitas: I didn't remove it. I just moved it outside the quickbook comment to fix this small bug in the docs that were rendering `BOOST_BEAST_DOXYGEN` sometimes.

image

To be honest, I just checked https://www.boost.org/doc/libs/1_79_0/libs/beast/doc/html/beast/using_websocket.html and it seems like this only happens locally. I have no idea why.

klemens-morgenstern:
Looks to be failing due to #2481.

klemens-morgenstern:
Drone passes now, the 4.8 issues seems fixed. @vinniefalco where are we at with this?

vinniefalco:
Should be ok

klemens-morgenstern:
Commited directly as squash

PR #2493, opened 2022-08-04
Switch to mac-11 in GHA

sdarwin:


klemens-morgenstern:
This is bad - we're getting a SEGFAULT from this change?  Only TSAN, but that should warrant an investigation.

@vinniefalco @madmongo1

klemens-morgenstern:
It looks like #2521 might have fixed this.

PR #2504, opened 2022-08-17
Remove unnecessary constant_time_size

ecatmur:
http::basic_fields::set_t::size() is unused
core::multi_buffer::list_type::size() is only compared to 1 (in shrink_to_fit()); compare iterators instead.

### DIFF
--- a/include/boost/beast/core/impl/multi_buffer.hpp
+++ b/include/boost/beast/core/impl/multi_buffer.hpp
@@ -779,7 +779,7 @@ shrink_to_fit()
         };
 
     // partial last buffer
-    if(list_.size() > 1 && out_ != list_.end())
+    if(out_ != list_.begin() && out_ != list_.end())
     {
         BOOST_ASSERT(out_ ==
             list_.iterator_to(list_.back()));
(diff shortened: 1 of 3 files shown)


vinniefalco:
Are there any benchmarks?

madmongo1:
@ecatmur Thank you for taking the time to post the PR.
What do we gain in practice from this change?
I am thinking in terms of memory use, search times, insertion times and so on.

ecatmur:
> Are there any benchmarks?

The only benchmark that is affected by this change is bench/buffers; I ran the benchmark (Ubuntu gcc 9, x86-64, release, threading-multi) but didn't observe any consistent differences in performance.

> @ecatmur Thank you for taking the time to post the PR. What do we gain in practice from this change? I am thinking in terms of memory use, search times, insertion times and so on.

http::fields reduces from 88 to 80 bytes (x86-64).
multi_buffer reduces from 72 to 64 bytes.

In terms of code size, multi_buffer::prepare() reduces from 1492 to 1378 bytes. There is no code change to fields test; gcc is smart enough to observe that the size field is never read, so entirely optimizes out stores to it.

As such, this change is unlikely to improve insertion times, at least for release builds with decent compilers.  It's more of a cleanup exercise.

vinniefalco:
hey, bytes are bytes :) they start adding up

madmongo1:
I agree. I think we should accept this PR.

klemens-morgenstern:
@madmongo1 @vinniefalco What's preventing this from getting merged?

vinniefalco:
Merge it I guess. Note that in the new message model, the container is not node based anyway. It is just a string with the serialized message (plus a small table of offsets to allow random access).

madmongo1:
I though I had merged it

PR #2511, opened 2022-08-30
Added json_body example.

klemens-morgenstern:


example/http/client/Jamfile
```diff
@@ -16,3 +16,5 @@ build-project sync ;
 build-project async-ssl ;
 build-project coro-ssl ;
 build-project sync-ssl ;
+
+build-project body ;
```
vinniefalco, line 20: json-body ?

klemens-morgenstern: I named it body in case we get more body examples.

klemens-morgenstern:
@vinniefalco waiting for your approval.

vinniefalco:
Commit messages should not describe a past event but should instead a sentence using a form of the verb to-be. For example: "streaming JSON body is an example"

vinniefalco:
Seems ok

Issue #2325, opened 2021-10-09
Regarding cancellation


Chlorie:
The latest version of the asio library (standalone 1.19 / boost 1.77) now supports per-operation cancellation. Is it possible to integrate this mechanic into beast so that operation cancellation could be more flexible? Thanks!

madmongo1:
It is on my roadmap.

madmongo1:
My feeling at the moment is that `terminal` cancellation is trivial (and
might even just work off the bat).
It probably won't be too much work to make HTTP operations support
`partial` cancellation.
With a bit of work, websocket streams could conceivably support `total`
cancellation.

vinniefalco:
websocket streams already support "total cancelation" just close the socket!

madmongo1:
That would be "terminal" cancellation.

"total" means that the cancellation happened without causing observable
side effects. I believe we can do this in the websocket stream because we
buffer internally.

vinniefalco:
> Is it possible to integrate this mechanic into beast so that operation cancellation could be more flexible? 

We could integrate it, but how do you think it will be better than what is already possible right now by either calling `cancel` on the underlying I/O object (`socket` or `ssl_stream`) or by simply closing the socket?

madmongo1:
The default cancellation mode requested is "all" which implies "total |
partial | terminal", so the implementation is free to implement the one it
is best able to do and ignore the others if necessary.

So if we were to request a total cancellation of a read and the websocket
is not in a position to buffer the input gathered so far, it is free to
ignore the cancellation request.

If it receives a "terminal" request, it should honour this. The simplest
way to do so is to simply close the underlying stream (assuming there is a
generic way to do that - streams are not always sockets). Another way to do
this is to simply ignore the cancellation in which case it will get passed
on to the next layer automatically. This will in most cases give us what we
want (any outstanding operations will be interrupted and complete as
operation_aborted).

So I think the default case is probably automatic, but I also think we can
do better - assuming there is a use case. The presence of a use case for
this is by no means proven at this point.

vinniefalco:
> if we were to request a total cancellation of a read and the websocket is not in a position to buffer the input gathered so far, it is free to ignore the cancellation request.


What good is canceling if the behavior is unpredictable, i.e. sometimes it works and sometimes it doesn't?

madmongo1:
I didn’t mean to imply that cancellation can be arbitrarily downgraded.
The caller is free to request “any kind of cancellation” of course. This is the default.
[...]

PR #2512, opened 2022-08-31
Added cancellation_slot support.

klemens-morgenstern:
Validating & implementing support for asio::cancellation_slot.

Implements #2325

### DIFF
--- a/include/boost/beast/_experimental/test/impl/stream.hpp
+++ b/include/boost/beast/_experimental/test/impl/stream.hpp
@@ -77,6 +77,16 @@ class basic_stream<Executor>::read_op : public detail::stream_read_op_base
           return net::get_associated_allocator(h_);
         }
 
+        using cancellation_slot_type =
+            net::associated_cancellation_slot_t<Handler>;
+
+        cancellation_slot_type
+        get_cancellation_slot() const noexcept
+        {
+            return net::get_associated_cancellation_slot(h_,
+                net::cancellation_slot());
+        }
+
         void
         operator()(error_code ec)
         {
--- a/include/boost/beast/http/impl/write.hpp
+++ b/include/boost/beast/http/impl/write.hpp
@@ -178,8 +178,18 @@ class write_op
     Stream& s_;
     serializer<isRequest, Body, Fields>& sr_;
     std::size_t bytes_transferred_ = 0;
+    net::cancellation_state st_{this->
+        beast::async_base<Handler, beast::executor_type<Stream>>
+            ::get_cancellation_slot()};
 
 public:
+    using cancellation_slot_type = net::cancellation_slot;
+    cancellation_slot_type get_cancellation_slot() const noexcept
+    {
+        return st_.slot();
+    }
+
+
     template<class Handler_>
     write_op(
         Handler_&& h,
@@ -227,6 +237,8 @@ public:
                         s_, sr_, std::move(*this));
                 }
                 bytes_transferred_ += bytes_transferred;
+                if (!ec && st_.cancelled() != net::cancellation_type::none)
+                    ec = net::error::operation_aborted;
                 if(ec)
                     goto upcall;
                 if(Predicate{}(sr_))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 342:
+
+* Support per operation cancellation
+
+--------------------------------------------------------------------------------
+
 Version 341:
 
 *  Expect header field with the "100-continue" is handled in upgrade.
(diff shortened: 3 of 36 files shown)


klemens-morgenstern:
The basic_stream functions are all inheriting async_base and are used in the main op. This will work fine with cancellation. 

It was mainly the wrappers (bind_front) and the write_op composed op that needed some help.

klemens-morgenstern:
This will need more validation work and there are a few gaps, mainly when the websockets does a ` post`. that's technically a race where we might miss a cancellation. I'll need some help on it's composition to figure that out though.

include/boost/beast/http/impl/write.hpp
```diff
@@ -227,6 +237,8 @@ class write_op
                         s_, sr_, std::move(*this));
                 }
                 bytes_transferred_ += bytes_transferred;
+                if (!ec && st_.cancelled() != net::cancellation_type::none)
+                    ec = net::error::operation_aborted;
```
madmongo1, line 241: I suspect this should not cancel when cancellation type is total.
does the op need to apply a filter to the cancellation state to prevent total cancellation (since it can't be supported)?
partial would be ok
terminal also ok

klemens-morgenstern: It's `terminal` by filter in the state already.

https://github.com/chriskohlhoff/asio/blob/fc901d79c11c72a0ba81764820d3c6a788c0f075/asio/include/asio/cancellation_state.hpp#L88

vinniefalco:
1. does each handler need a cancellation slot? why not put it on the stream (or is that nonsensical)?

2. will this be squashed

klemens-morgenstern:
1. This is per-op cancellation, not stream cancellation
2. Yes.

klemens-morgenstern:
@madmongo1 can you review the current version?

vinniefalco:
individual ops cannot be reliably canceled though, and there's no way to know what state the stream is in afterwards

klemens-morgenstern:
Cancelling http::async_read is currently unreliable, because the cancellation state of a pr is protected - probably just a mistake in asio, will make a pr.

vinniefalco:
Are you going to squash this?

include/boost/beast/_experimental/test/impl/stream.hpp
```diff
@@ -77,6 +77,16 @@ class basic_stream<Executor>::read_op : public detail::stream_read_op_base
           return net::get_associated_allocator(h_);
         }
 
+        using cancellation_slot_type =
```
vinniefalco, line 80: do we need to support cancelation for the test stream?

include/boost/beast/_experimental/test/impl/stream.hpp
```diff
@@ -77,6 +77,16 @@ class basic_stream<Executor>::read_op : public detail::stream_read_op_base
...
+            net::associated_cancellation_slot_t<Handler>;
+
+        cancellation_slot_type
+        get_cancellation_slot() const noexcept
```
vinniefalco, line 84: What if people are currently using `websocket::stream` with a NextLayer that doesn't support cancelation slots? Will they get a compile error about a missing nested type and `get_cancellation_slot` function?

include/boost/beast/_experimental/test/impl/stream.hpp
klemens-morgenstern: I am using tcp streams on purpose so we don't.

include/boost/beast/_experimental/test/impl/stream.hpp
klemens-morgenstern: No, it will just get ignored. It's a property on the handler, not the stream

include/boost/beast/_experimental/test/impl/stream.hpp
vinniefalco: but it looks like we are adding cancelation to the test stream here?

klemens-morgenstern:
@sehe @madmongo1 I think it's ready, so I could use another review, since this is stuff can be critical.

klemens-morgenstern:
Note to self: document which op supports what kind of cancellation. Most likely only `terminal` everywhere. Reference:

https://github.com/chriskohlhoff/asio/blob/master/asio/include/asio/ssl/stream.hpp#L567

madmongo1:
Is it possible to squash all the commits that relate to cancelation?

klemens-morgenstern:
Note to self: Verify partial cancellation doesn't break the `websocket::async_read_some` and `websocket::async_writes_some`

klemens-morgenstern:
Sockets `.async_read_some` and `async_write_some` can be cancelled `partial`ly.

https://www.boost.org/doc/libs/1_80_0/doc/html/boost_asio/reference/basic_stream_socket/async_read_some.html

klemens-morgenstern:
wrong cancellation type issue & docs are done. Should be ready to merge.

PR #2526, opened 2022-10-05
Sendfile

klemens-morgenstern:


klemens-morgenstern:
I will possibly be working on an asio impl of sendfile/TransmitFile. asio is where this belongs anyhow, so closing the PR for now.

PR #2532, opened 2022-10-09
Added logic to explicitly close TCP socket for sync HTTPS client

pavel-odintsov:
Hello!

I noticed that example client does not include logic which explicitly closes TCP socket, I think it's good ideas to add it here.

In addition to leaked socket I hit quite nasty issue when my CDN (Fastly) does not implement proper TLS termination and shutdown() hands indefinitely. So I've added comment about this case too.

### DIFF
--- a/example/http/client/sync-ssl/http_client_sync_ssl.cpp
+++ b/example/http/client/sync-ssl/http_client_sync_ssl.cpp
@@ -107,6 +107,9 @@ int main(int argc, char** argv)
 
         // Gracefully close the stream
         beast::error_code ec;
+
+        // For some web servers it may hang as they do not implement TLS session termination properly
+        // In that case you may need to skip it and close TCP socket
         stream.shutdown(ec);
         if(ec == net::error::eof)
         {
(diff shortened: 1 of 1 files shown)


example/http/client/sync-ssl/http_client_sync_ssl.cpp
```diff
@@ -118,6 +121,13 @@ int main(int argc, char** argv)
             throw beast::system_error{ec};
 
         // If we get here then the connection is closed gracefully
+
+        // As final step we need to close underlying TCP connection
+        beast::get_lowest_layer(stream).socket().shutdown( boost::asio::socket_base::shutdown_both, ec);
```
vinniefalco, line 126: well this might need to be added to ALL the SSL clients :) but read this:

http://www.serverframework.com/asynchronousevents/2011/01/time-wait-and-its-design-implications-for-protocols-and-scalable-servers.html

pavel-odintsov:
I had conversation with my friend from OpenSSL and I think this issue will need more investigation on my side. I use old OpenSSL (1.0) and it may be root cause of such behaviour. I'll repeat my tests with fresh OpenSSL and will report back.

pavel-odintsov:
After upgrade to OpenSSL 1.1.1q shutdown works fine and does not block execution. I can confirm that old OpenSSL (1.0.2d) was root cause.

Socket close is not needed too, shutdown moves socket to TIME_WAIT status which is perfectly fine.

PR #2533, opened 2022-10-10
added api version section to the docs

klemens-morgenstern:


include/boost/beast/version.hpp
```diff
@@ -22,6 +22,7 @@
 */
 #define BOOST_BEAST_VERSION 336
 
+/// A stringified BOOST_BEAST_VERSION
```
vinniefalco, line 25: Do the docs pick this up?

PR #2534, opened 2022-10-10
Issue/264

klemens-morgenstern:


include/boost/beast/_experimental/test/handler.hpp
```diff
@@ -35,47 +35,55 @@ class handler
...
         pass_ = true;
     }
 
     void
     operator()()
     {
-        BEAST_EXPECT(! pass_); // can't call twice
-        BEAST_EXPECT(! ec_);
+        ::boost::beast::unit_test::suite::this_suite()->expect(!pass_, loc_.file_name(), loc_.line());
+        if (ec_)
```
vinniefalco, line 82: `if( ec_.failed() )`

include/boost/beast/websocket/impl/accept.hpp
```diff
@@ -192,6 +194,14 @@ class stream<NextLayer, deflateSupported>::response_op
         , res_(beast::allocate_stable<response_type>(*this,
             sp->build_response(req, decorator, result_)))
     {
+        auto itr = req.find(http::field::expect);
+        if (itr != req.end() && itr->value() == "100-continue") // do
+        {
+            res_100_.version(res_.version());
+            res_100_.result(http::status::continue_);
+            res_100_.prepare_payload();
```
vinniefalco, line 202: This is ok I guess, kind of.. but.. it will be missing the Server: field no? And whatever other decorators the caller wants. Also, is the "100-continue" value case insensitive?

vinniefalco: I don't want to invest too much into this because I think that the number of times this code will actually execute in the wild is exactly zero.

include/boost/beast/_experimental/test/handler.hpp
klemens-morgenstern: `ec_ ` is the comparison, so if it's set the error-less completion is wrong. but it should be 

`ec_ && ec_->failed()` since ec_ could be set to success (in theory)

PR #2538, opened 2022-10-11
Cpp20

klemens-morgenstern:


vinniefalco:
No updates to CHANGELOG.md per commit?

klemens-morgenstern:
No API changes, so no changelog. I interpreted that to be the API version. I guess the default_token stuff could be interpreted as such, but surely examples don't?

example/advanced/server-flex-awaitable/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+#
+# Copyright (c) 2016-2017 Vinnie Falco (vinnie dot falco at gmail dot com)
```
vinniefalco, line 2: Your name goes here not mine

klemens-morgenstern:
Blocked by #2512, since the flex example is using cancellation_signals.

example/http/client/awaitable/http_client_awaitable.cpp
```diff
@@ -0,0 +1,158 @@
...
+namespace beast = boost::beast;         // from <boost/beast.hpp>
+namespace http = beast::http;           // from <boost/beast/http.hpp>
+namespace net = boost::asio;            // from <boost/asio.hpp>
+using tcp = boost::asio::ip::tcp;       // from <boost/asio/ip/tcp.hpp>
+
+//------------------------------------------------------------------------------
+
+// Report a failure
+void
+fail(beast::error_code ec, char const* what)
```
yuvalif, line 42: this function is not used in the code. error is printed in the exception handler

example/http/client/awaitable/http_client_awaitable.cpp
```diff
@@ -0,0 +1,158 @@
...
+net::awaitable<void>
+do_session(
+    std::string host,
+    std::string port,
+    std::string target,
+    int version)
+{
+    // These objects perform our I/O
+
+    auto resolver = net::use_awaitable.as_default_on(tcp::resolver(co_await net::this_coro::executor));
```
yuvalif, line 57: what is the benefit of creating the resolver in this way compared to:
```cpp
tcp::resolver resolver(ioc);
```
and then invoking the coroutine with `net::use_awaitable` as an argument?
```cpp
co_await resolver.async_resolve(host, port, net::use_awaitable);
```

example/http/client/awaitable/http_client_awaitable.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
+#include <boost/beast/version.hpp>
+#include <boost/asio/awaitable.hpp>
+#include <boost/asio/co_spawn.hpp>
+#include <boost/asio/detached.hpp>
+#include <boost/asio/use_awaitable.hpp>
+
+#if defined(BOOST_ASIO_HAS_CO_AWAIT)
```
yuvalif, line 26: why not adding `<cxxflags>"-std=c++20"` to the Jamfile ?

example/http/client/awaitable/http_client_awaitable.cpp
klemens-morgenstern: Brevity.

example/http/client/awaitable/http_client_awaitable.cpp
klemens-morgenstern: Good catch, will remove.

example/http/client/awaitable/http_client_awaitable.cpp
klemens-morgenstern: Because it should also build with C++23.

example/http/client/awaitable/http_client_awaitable.cpp
vinniefalco: No, not just brevity - passing `io_context` around is a code smell. You should be thinking and writing in terms of executors not contexts. That line would be better if it read:
```
tcp::resolver resolver( ioc.get_executor() );
```

Even better would be to write `make_executor` and call that from everywhere, then you can switch to a strand easily by changing one function.

example/advanced/server-flex-awaitable/advanced_server_flex_awaitable.cpp
```diff
@@ -0,0 +1,1188 @@
...
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+//------------------------------------------------------------------------------
+//
+// Example: Advanced server, flex (plain + SSL)
```
madhur4127, line 12: Can we please add a comment on what this advanced server does? what's _advanced_ about it? what features does this example show?

sorry for being lazy but its ~1100 lines and its a big time investment to read and then figure out if I need it.

PR #2542, opened 2022-10-17
libs/variant2 is available for windows ci runs.

klemens-morgenstern:


tools/get-boost.sh
```diff
@@ -83,6 +83,7 @@ git submodule update --init --depth 20 --jobs 4 \
     libs/type_index \
     libs/typeof \
     libs/unordered \
+    libs/variant2 \
```
vinniefalco, line 86: Did Beast use variant2 before?

klemens-morgenstern: No, it looks to be a transitive dependency from boost/system from by json example. I am currently searching for where it's included from.

klemens-morgenstern: Added to json as a dep 3 days ago, so that would be it.

vinniefalco: Ok

PR #2550, opened 2022-10-23
extern template declarations for default usages.

klemens-morgenstern:
Checking if this make the compile times better.


klemens-morgenstern:
if done this needs explicit .ipp files.

PR #2553, opened 2022-10-31
boost 1.81 release notes.

klemens-morgenstern:
Should I add another api version bump commit so it's on top of the commit history?


vinniefalco:
Yes increment the version number, and this looks very exciting !

klemens-morgenstern:
I have another asio based fix coming in (Chris changes something) - will do that after the merging this pr

PR #2558, opened 2022-11-01
Add request.prepare_payload() to http_client_sync example

meastp:
Even though the string_body is empty it should call prepare_payload to avoid an easy mistake for anyone reusing the example,

This is identical in the other client examples as well. If acceptet, this PR could be amended, or I can make similar PRs.

### DIFF
--- a/example/http/client/sync/http_client_sync.cpp
+++ b/example/http/client/sync/http_client_sync.cpp
@@ -66,6 +66,7 @@ int main(int argc, char** argv)
         http::request<http::string_body> req{http::verb::get, target, version};
         req.set(http::field::host, host);
         req.set(http::field::user_agent, BOOST_BEAST_VERSION_STRING);
+        req.prepare_payload();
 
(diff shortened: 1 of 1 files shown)


klemens-morgenstern:
Should't the `get` us an `empty-body` instead?

meastp:
@klemens-morgenstern Well, it could (and perhaps should, since there is no body content), but then the documentation should provide an example with `string_body` as well, that does call `prepare_payload`, imho. :) It is unavoidable that users often start by copying the examples in the documentation, and providing an example with a body and prepare_payload will avoid a small roadblock.

vinniefalco:
I agree that we should do something, I'm not exactly sure what.

klemens-morgenstern:
Please check #2598, which should be a better basis for copy-pasta. Although maybe we need to instead do a cook-book.

ashtum:
Closing this pull request because calling `prepare_payload()` is not required in those examples, and adding it might cause confusion for readers.

PR #2560, opened 2022-11-07
basic_fields handles initializer-lists.

klemens-morgenstern:
This is very convenient for my requests library that (for now) uses the fields structure directly.


vinniefalco:
I don't think we should keep adding stuff like this, and also we are in the release cycle so it is too late to add a new API

klemens-morgenstern:
It is meant for 1.82.

vinniefalco:
You should write a free function `init_fields` that lets you construct with an initializer-list, keep it in your library, and not add it to beast

PR #2577, opened 2022-12-07
Switched self-posting to the io-object's executor.

klemens-morgenstern:
This is a conceptual change in the recent asio version that is reflected in any_completion_handler.

1.82 change, as it's behaviour changing.

### DIFF
--- a/include/boost/beast/core/async_base.hpp
+++ b/include/boost/beast/core/async_base.hpp
@@ -389,11 +389,13 @@ public:
         if(! is_continuation)
         {
             auto const ex = get_executor();
-            net::post(net::bind_executor(
-                ex,
-                beast::bind_front_handler(
-                    std::move(h_),
-                    std::forward<Args>(args)...)));
+            net::post(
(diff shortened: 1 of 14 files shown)


include/boost/beast/core/async_base.hpp
```diff
@@ -389,11 +389,13 @@ class async_base
         if(! is_continuation)
         {
             auto const ex = get_executor();
-            net::post(net::bind_executor(
-                ex,
-                beast::bind_front_handler(
-                    std::move(h_),
-                    std::forward<Args>(args)...)));
+            net::post(
+                wg1_.get_executor(),
```
vinniefalco, line 393: this sounds dangerously close to "wg21"

vinniefalco:
I have no idea if this is good or bad, because I have lost a firm understanding of executors after all the changes in Asio (plus an unhealthy dose of poor documentation). So I will have to take your word. It would be nice if there was something I could read which provides a firm understanding of the fundamentals.

PR #2592, opened 2022-12-17
File body improve

klemens-morgenstern:
- commit 1 is a config option, that's surely useful
- commit 2 bugfix
- commit 3 a very useful feature.

### DIFF
--- a/doc/qbk/03_core/8_conf_macros.qbk
+++ b/doc/qbk/03_core/8_conf_macros.qbk
@@ -35,6 +35,13 @@ failure to do so may result in violations of ODR (One Definition Rule).
         Enables the use of deprecated APIs within Beast.
     ]
 ]
+[
+    [
+        BOOST_BEAST_FILE_BUFFER_SIZE
+    ][
+        Sets the small buffer size for the file_body. Defaults to 4096.
+    ]
+]
 ]
 
 [endsect]
--- a/include/boost/beast/core/detail/config.hpp
+++ b/include/boost/beast/core/detail/config.hpp
@@ -113,4 +113,8 @@ namespace net = boost::asio;
 
 #endif
 
+#ifndef BOOST_BEAST_FILE_BUFFER_SIZE
+#define BOOST_BEAST_FILE_BUFFER_SIZE 4096
+#endif
+
 #endif
--- a/include/boost/beast/http/basic_file_body.hpp
+++ b/include/boost/beast/http/basic_file_body.hpp
@@ -160,6 +160,19 @@ public:
     */
     void
     reset(File&& file, error_code& ec);
+
+    /** Set the cursor position of the file.
+
+        This function can be used to move the cursor of the file ahead
+        so that only a part gets read. This file will also adjust the
+        value_type, in case the file is already part of a body.
+
+        @param offset The offset in bytes from the beginning of the file
+
+        @param ec Set to the error, if any occurred
+    */
+
+    void seek(std::uint64_t offset, error_code& ec);
 };
 
 template<class File>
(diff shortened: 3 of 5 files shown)


include/boost/beast/http/basic_file_body.hpp
```diff
@@ -210,8 +223,29 @@ reset(File&& file, error_code& ec)
...
+        file_size_ -= file_.pos(ec);
+}
+
+template<class File>
+void
+basic_file_body<File>::
+value_type::
+seek(std::uint64_t offset, error_code& ec)
+{
+    file_.seek(offset, ec);
```
malixi94, line 251: In my testing of Beast with my own implementation of file serving and seeking from the system, reading the last part of a file will not only set the eof bit (which seeking automatically clears) but also the fail bit. So before seeking here I think you might want to test for failure / error and if so, clear the fail bit condition.

Though this is only if you want to make the file class here be more stateful with general seeking.  But if you were to only allow seeking the file once (say through an overload of the reset method and getting rid of the seek method) then you don't need to test for this condition. If the user wants to seek the file again then they need to reset it or discard the current object and instantiate a new file class object.

include/boost/beast/http/basic_file_body.hpp
```diff
@@ -133,6 +133,19 @@ class basic_file_body<File>::value_type
         return file_size_;
     }
 
+    /// limits the size of the content that can be sent - returns the new size.
+    std::uint64_t
+    limit_size(std::uint64_t limit, error_code & ec)
+    {
+        std::uint64_t actual_size = file_.size(ec);
```
malixi94, line 140: By this point in the state of the file object, doesn't the member `file_size_` already have the proper size info? If so why do you need to calculate it again with this automatic variable? If my hypothesis is correct, this method could be a one liner:
`return file_size_ = (std::min)(file_size_, limit);`

klemens-morgenstern: This is so you can write only a limited amount of the file, i.e. an incomplete chunk

klemens-morgenstern:
@vinniefalco I think this is ready. It'll allow partial transmission of files, but I think it'll need your approval.

include/boost/beast/core/detail/config.hpp
```diff
@@ -113,4 +113,8 @@ namespace net = boost::asio;
 
 #endif
 
+#ifndef BOOST_BEAST_FILE_BUFFER_SIZE
+#define BOOST_BEAST_FILE_BUFFER_SIZE 4096
```
vinniefalco, line 117: This is more ugly than my ex but I guess its better than the alternative, which is nothing. That number should probably be higher, no?

include/boost/beast/http/basic_file_body.hpp
```diff
@@ -210,8 +223,29 @@ reset(File&& file, error_code& ec)
...
+basic_file_body<File>::
+value_type::
+seek(std::uint64_t offset, error_code& ec)
+{
+    file_.seek(offset, ec);
+    // Cache the size
+    if (!ec)
+        file_size_ = file_.size(ec);
+
+    // Consider the offset
```
vinniefalco, line 256: These comments add no value... lol

include/boost/beast/http/basic_file_body.hpp
vinniefalco: To write a range, ok

vinniefalco: This is weird interface though 'limit size'.. Why not just a `set_range` for both the position and length?

include/boost/beast/core/detail/config.hpp
klemens-morgenstern: 4096 is the previous default and it's asio's default, too.

doc/qbk/03_core/8_conf_macros.qbk
```diff
@@ -35,6 +35,13 @@ failure to do so may result in violations of ODR (One Definition Rule).
         Enables the use of deprecated APIs within Beast.
     ]
 ]
+[
+    [
+        BOOST_BEAST_FILE_BUFFER_SIZE
```
vinniefalco, line 40: This needs the backticks to make it use the monospaced font?

include/boost/beast/http/basic_file_body.hpp
klemens-morgenstern: Because you might set the limit when you've already read parts of it, i.e. moved the cursor. So you limit the size to 2MB after 1MB is read already. the set_range would reset this.

vinniefalco: The precondition for `set_range` should be that no reading has occurred yet

vinniefalco:
This is better yes

PR #2610, opened 2023-01-03
Fix compatibility with -fno-exceptions

bc-lee:
When I'm trying to compile a simple C++ code using boost/beast and clang with `-fno-exceptions` I got the following error. This patch is intended to fix these compilation errors.
```
/path/to/include/boost/beast/core/impl/basic_stream.hpp:152:5: error: cannot use 'try' with exceptions disabled
    try

/path/to/include/boost/beast/include/boost/beast/core/impl/buffers_cat.hpp:104:13: error: cannot use 'throw' with exceptions disabled
            throw std::logic_error("");
```

### DIFF
--- a/include/boost/beast/core/impl/basic_stream.hpp
+++ b/include/boost/beast/core/impl/basic_stream.hpp
@@ -149,13 +149,8 @@ close() noexcept
         error_code ec;
         socket.close(ec);
     }
-    try
-    {
-        timer.cancel();
-    }
-    catch(...)
-    {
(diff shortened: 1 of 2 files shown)


bc-lee:
Oh, lots of CI failures. Sorry to hear that boost/beast requires `BOOST_ASIO_NO_DEPRECATED`. If that macro exists, there is no way to capture the error in a world without C++ exceptions. Feel free to close this PR.

PR #2643, opened 2023-02-20
1.82  release

klemens-morgenstern:


vinniefalco:
This is fine. Usually I merge to master periodically instead of waiting till the end.

PR #2651, opened 2023-03-15
The changes in stream_impl.hpp fix a problem where idle_counter != 0 …

johannesparty:
The change in stream_impl.hpp is fixes a bug that occurs when a stream that timed out is reused.  This would often result in the first assert in update_timer (near line 422) getting called erroneously.  This fix is just resetting idle_counter to 0 on open and reset.  This looks (to me) like a very straightforward and completely safe change.

The change in read.hpp is to fix the ping timeout.  The code currently doesn't reset the timeout when a regular (non-ping) message is received. This is  because the coroutine for operator() near line 81 seldom executes the update near line 97 but instead jumps back into the main loop somewhere, bypassing the call to update_timer.  

While the fix I included works (tested), I don't claim it is the best possible fix.  Someone more familiar with the internals may have a better solution.  I'm happy as long as the problem is fixed.  If am happy to test an alternative fix, but may not have time in the short term to provide a test to others (I will do my best, just trying to be honest).

With this fix, the ping timeout behaves as documented, however it is still not ideal when one side cannot send ping (for example Browser).  Filed Issue #2652  with details and proposed solution (but no code).

### DIFF
--- a/include/boost/beast/websocket/impl/read.hpp
+++ b/include/boost/beast/websocket/impl/read.hpp
@@ -196,6 +196,7 @@ public:
                     }
                     BOOST_ASSERT(impl.rd_block.is_locked(this));
                     impl.rd_buf.commit(bytes_transferred);
+                    if (bytes_transferred) impl.update_timer(this->get_executor());
                     if(impl.check_stop_now(ec))
                         goto upcall;
                     impl.reset_idle();
--- a/include/boost/beast/websocket/impl/stream_impl.hpp
+++ b/include/boost/beast/websocket/impl/stream_impl.hpp
@@ -155,6 +155,7 @@ struct stream<NextLayer, deflateSupported>::impl_type
     {
         // VFALCO TODO analyze and remove dupe code in reset()
         timer.expires_at(never());
+        idle_counter = 0;
         timed_out = false;
         cr.code = close_code::none;
         role = role_;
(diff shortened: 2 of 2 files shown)


klemens-morgenstern:
I don't think you can reuse a stream that timed out to begin with. Are you using it with a websocket or just for raw tcp?

johannesparty:
I am successfully reusing a stream with websocket (with the very minor changes suggest in the PR).  I might be reusing it mostly in cases where it is closed independently of timeout.  I call async_connect again on beast::websocket::stream<tcp_socket>.next_layer(), then async_handshake.  I've tested with 1.80.0 and 1.81.0.

klemens-morgenstern:
Well just because it works, doesn't mean it'll always work.

I probably need to create some example of a reusable stream with a timeout, because I don't think the tcp_stream is cutting it.

johannesparty:
Ok I appreciate your help.

It is interesting that variables including timed_out in stream_impl are set in the constructor, in open(), and in reset() which did lead me to believe the websocket stream could be reused.  

Of course, I could create a unique_ptr to the stream or something if I have to, but I would need to understand the rules clearly.  With that said, the changes I made in the PR do seem to work...

klemens-morgenstern:
have you looked into per-op cancellation? I personally feel the tcp_stream timeouts are an obsolete design since asio introduced that.

klemens-morgenstern:
@johannesparty where are we at with this pr?

PR #2676, opened 2023-05-01
field.ipp: IWYU for <cstdint>

ednolan:
Previously, a file that included only boost/beast/http/fields.hpp would fail to compile on GCC 13.1.0 with the following error:

```
In file included from /boost/boost/beast/http/field.hpp:411,
                 from /boost/boost/beast/http/fields.hpp:16,
                 from /repro/boost_repro.cpp:1:
/boost/boost/beast/http/impl/field.ipp:30:10: error: 'uint32_t' in namespace 'std' does not name a type; did you mean 'wint_t'?
   30 |     std::uint32_t
      |          ^~~~~~~~
      |          wint_t
```

This was because boost/beast/http/impl/field.ipp relied on a transitive include that is no longer present in libstdc++ 13.

This commit addresses the issue by adding a `<cstdint>` include to boost/beast/http/impl/field.ipp.

### DIFF
--- a/include/boost/beast/http/impl/field.ipp
+++ b/include/boost/beast/http/impl/field.ipp
@@ -14,6 +14,7 @@
 #include <boost/assert.hpp>
 #include <algorithm>
 #include <array>
+#include <cstdint>
 #include <cstring>
 #include <ostream>
 


vinniefalco:
That should have been caught with this unit test: https://github.com/boostorg/beast/blob/b0996f099ac4cb024675b0b320f07e2297c3130c/test/beast/http/field.cpp#L11

ednolan:
My guess would be that the unit tests were never run with GCC 13. I don't see that compiler in the list of CI checks

vinniefalco:
of COURSE... the ONE configuration that we don't test... is the one that has the error :)

ednolan:
To be fair, GCC 13 only came out 5 days ago.

vinniefalco:
Oh! I thought it was older.. thanks

royjacobson:
Just had the same problem with `beast/core/file_base.hpp`. It might be a wider problem :/

vinniefalco:
we need to add gcc13 to the matrix and fix all the errors

klemens-morgenstern:
Can confirm from local install (fedora 38)

klemens-morgenstern:
Completed in #2682

Issue #2716, opened 2023-08-10
inactivity_timeout is triggered regardless of having received pings


xim:
### Version of Beast

347

### Steps necessary to reproduce the problem

I'm using plain, async `boost::beast::websocket` for a server and a client.

I have a server that's sending ping packets at a 15s interval and keeping the connection alive correctly. I have verified that the client receives these and responds with pongs.

Since I have these ping packets going, I've added a timeout in the client to try to detect if the connection goes down:

```cpp
ws_.set_option(boost::beast::websocket::stream_base::timeout{
    60s,
    30s,
    false
});
```

I had assumed that this would be kept alive by the ping packets sent from the server, but it isn't. The client disconnects with the message `The socket was closed due to a timeout`.

Setting the final flag (keep_alive_pings) to `true` makes the client keep the connection alive by doing the same as the server's already doing. Given the server set to send ping frames at 15s intervals, this setting on the client:

```cpp
ws_.set_option(boost::beast::websocket::stream_base::timeout{
    60s,
    80s,
    true
});
```

Means I'll get this behaviour:

```cpp
00:00 Client connects
00:15 Server pings as expected
00:15 Client pongs as required
00:30 Server pings as expected
00:30 Client pongs as required
00:40 Client pings despite having seen activity in the form of a ping
```

I feel is the desired behaviour would be to count incoming ping frames as activity. If that can't be made the default behaviour, could we try to add it as a config on `boost::beast::websocket::stream_base::timeout`?

xim:
I made a somewhat-minimal reproduction at https://gist.github.com/xim/06d60c455363fbb764c14f21d7705c40

Compiled and run by running each in a separate terminal:

```
g++ -g -std=gnu++17 websocket_server_async.cpp && echo running && ./a.out 127.0.0.1 8080 1
g++ -g -std=gnu++17 websocket_client_async.cpp && ./a.out 127.0.0.1 8080 'test'
```

klemens-morgenstern:
What are you using as the tcp stream? Are you calling async_read perpetually?

xim:
As shown in the gist, I made a somewhat-minimal reproduction with the example websocket_client_async.cpp as a base.

It's a `websocket::stream<beast::tcp_stream>` and I re-call `async_read` in the `on_read` handler

xim:
Full code instead of patch at https://gist.github.com/xim/97874f6d7e27402f26228a0da6448236

[...]

PR #2718, opened 2023-08-11
websockets: Count pings from server as activity for idle_timeout

xim:
If the stream is receiving control packets like ping, don't count it as idle. This means you can enable `timeout_opt.keep_alive_ping` on only one side to get heartbeat.

Addresses issue #2716


klemens-morgenstern:
Can't this be configured already?

```cpp
beast::websocket::stream_base::timeout to;
my_ws.get_option(to);
to.keep_alive_pings = true;
my_ws.set_option(to);
```

xim:
This PR changes the behavior when `keep_alive_ping` is `false`.

Without the change: receiving control frames does not count as "activity"

With the change: receiving control frames counts as activity

I could make that behaviour change another config member of the timeout struct if wanted.

klemens-morgenstern:
I understand that - but why? Isn't setting the option achieving the same thing? What's the difference?

xim:
Ah. The setting makes you send ping frames at an even interval. This change counts incoming control frames as activity regardless of whether or not you are sending pings yourself.

klemens-morgenstern:
Receiving control frames is not activity. 

So we would need another option to allow the pings to be counted as such. This all seems off to me, why don't you just send pings or increase the timeout?

xim:
I first reported this because I was confused by the `idle_timeout` / `keep_alive_pings` config options; Since a pong is activity when you've sent a ping, shouldn't a ping be activity as well?

If not, timely disconnect detection needs ping and pong to be sent in both directions, doubling the number of packets sent/received for no additional benefit.

Not a big issue, but I feel counting ping as activity is the more intuitive behaviour. Adding this as a `bool incoming_ping_keepalives = false` option also makes sense if keeping the old behaviour as default is a priority.

xim:
Updated the PR with tests, @ashtum

I feel using a second wall clock time may be excessive, but seems to be somewhat in line with what other tests have done

xim:
Can/should I add/change documentation somehow? Is it sufficient that I update the inline documentation in `stream_base.hpp`?

ashtum:
Yes, please update the docs [here](https://github.com/boostorg/beast/tree/develop/doc/qbk/06_websocket) as well (if there is anything related).

ashtum:
@xim Anything I can help with?

xim:
Sorry, @ashtum, this just disappeared in the Big Pile of Stuff. I'll try to update docs on Monday

xim:
Apologies for forgetting about the documentation update. I rebased on develop, added documentation and pushed again.

xim:
I'm happy with this now. If you want me to squash all commits, or fixup my fixup, or anything, just give the word =)

klemens-morgenstern:
This can be squashed on our side, this looks like a single commit.

From what I can tell @vinniefalco approves the behavioural change, so everything looks good to me.

xim:
Looks like one of the tests has a timer that's 500ms and expects it to have been triggered after 600ms. I'll make it more permissive.

xim:
Made that 750ms, effectively bumping the slack from 100ms to 250ms

xim:
Apologies for the noise. Tweaking these values is hard.

ashtum:
@xim Why did you alter the timing values in the tests? Did you encounter any failures as a result of the tight timings?

xim:
Yes. I had to increase/change margins many times before all the different CI builders were green at the same time. The original timing worked once across the board, but not on the next run. The current timing should be stable.

I'd love to see these tests run with stubbed time, for stability and execution speed. But don't know if that's readily available.

ashtum:
Hey @xim, I wanted to let you know that I've decided to review and merge this after the 1.84 release. I reckon it might be risky to do it for 1.84.

xim:
I have, for now, implemented most of the desired behaviour on top of the control callback. Looking forward to cleaning up that code ^_^

vinniefalco:
Why shouldn't control frames count as activity?

ashtum:
> Why shouldn't control frames count as activity?

Did you mean, "Why should control frames count as activity?" If so, we had a discussion in https://github.com/boostorg/beast/issues/2716. The idea is that receiving any form of frame from the other side can be an indication that it is still active.

vinniefalco:
> Did you mean, "Why should control frames count as activity?"

No. I am asking what the rationale would be, if any, to NOT count control frames as activity. I cannot think of one. But that doesn't mean it does not exist (even if the rationale is weak).

xim:
> I am asking what the rationale would be, if any, to NOT count control frames as activity.

Maybe if you want to throw out sessions that "aren't doing anything" at an application level?

Not arguing for this being sensible, just answering the hypothetical =)

ashtum:
Based on the source code, we have been treating the control frames as activity from the beginning (by setting idle_counter to zero). It is this branch in the code that has caused the timeout: https://github.com/boostorg/beast/blob/d0dd9c50694bcbb8836775e780eb046f66b3e1d7/include/boost/beast/websocket/impl/stream_impl.hpp#L572
I don't know whether this behavior has been intentional or accidental.

xim:
> Based on the source code, we have been treating the control frames as activity from the beginning

Yes, but only when the option `keep_alive_pings` is set. So pings and pongs only count as activity if you're *sending* pings.

ashtum:
> Yes, but only when the option `keep_alive_pings` is set. So pings and pongs only count as activity if you're _sending_ pings.

I mean we have been setting `idle_counter` to zero unconditionally.

xim:
> I mean we have been setting `idle_counter` to zero unconditionally.

Yes, but without `keep_alive_pings` the value of `idle_counter` is ignored and we go straight to timeout.

ashtum:
> Yes, but without `keep_alive_pings` the value of `idle_counter` is ignored and we go straight to timeout.

Yeah, my point was that this might have been an unintentional behavior because we wanted to only send the ping message when `keep_alive_pings` is set and accidentally ignored the situation when `idle_counter` is zero and `keep_alive_pings` is unset.

Besides that, other parts of the code show that we have been counting control frames as activity (by unconditionally setting idle_counter to zero).

ashtum:
@xim, could you please squash these commits into a single commit?

xim:
Yeah, the timing there is apparently night-impossible to get correct in a way that always passes on all platforms. In that one case, a ping hasn't been processed within 400ms. I can increase everything 25% again so it's 500ms. But is anyone has a way of running these tests with stubbed time, that would be preferable...

xim:
The failures appear unrelated to my commit

test/beast/websocket/ping.cpp
```diff
@@ -88,6 +88,79 @@ class ping_test : public websocket_test_suite
...
+            es.async_ping();
+            test::run_for(ioc_, std::chrono::milliseconds(500));
+            // The server sent a ping at 1.25s mark, and we're now at 1.75s mark.
+            // We haven't hit the idle timer yet (happens at 1s, 2s, 3s)
+            BEAST_EXPECT(ws.impl_->idle_counter == 0);
+            BEAST_EXPECT(!got_timeout);
+
+            test::run_for(ioc_, std::chrono::milliseconds(750));
+            // At 2.5s total; should have triggered the idle timer
+            BEAST_EXPECT(ws.impl_->idle_counter == 1);
```
ashtum, line 131: I believe the tight timing is a consequence of asserting `idle_counter` in multiple places. There is a good chance that these timings could lead to non-deterministic test results. Therefore, I suggest not checking for `idle_counter` and instead only verifying the side effect (inactivity timeout doesn't occur when receiving pings).

ashtum:
@xim, I've removed the `idle_counter` assertions in the hope that this will reduce the chance of getting non-deterministic test results. Let's see if we can run CI multiple times with all tests passing.

include/boost/beast/websocket/impl/read.hpp
```diff
@@ -227,8 +227,6 @@ class stream<NextLayer, deflateSupported>::read_some_op
                     // Handle ping frame
                     if(impl.rd_fh.op == detail::opcode::ping)
                     {
-                        impl.update_timer(this->get_executor());
```
ashtum: Updating the timer is necessary; otherwise, it will attempt to ping the other side even after receiving a ping.

sdarwin:
Testing out a new docs servers, there may be another "automated preview" shortly on this pull request.

PR #2760, opened 2023-10-27
Replace internal uses of beast::bind_front_handler with asio::prepend

ashtum:
We cannot deprecate `beast::bind_front_handler` in favor of `asio::prepend` or `asio::append` because it has a [special overload](https://github.com/boostorg/beast/blob/667e9ea251e862fa7b62a3ee94054faef201702e/include/boost/beast/core/detail/bind_handler.hpp#L240) for cases when the handler is a member function pointer which enables code like this to work:

```C++
ws_.async_accept(
    req,
    beast::bind_front_handler(
        &websocket_session::on_accept,
        shared_from_this()));
```
While this pattern is only employed in examples and tests, all instances within the library itself can be replaces with `asio::prepend`.

### DIFF
--- a/include/boost/beast/_experimental/test/impl/stream.hpp
+++ b/include/boost/beast/_experimental/test/impl/stream.hpp
@@ -10,14 +10,15 @@
 #ifndef BOOST_BEAST_TEST_IMPL_STREAM_HPP
 #define BOOST_BEAST_TEST_IMPL_STREAM_HPP
 
-#include <boost/beast/core/bind_handler.hpp>
 #include <boost/beast/core/buffer_traits.hpp>
 #include <boost/beast/core/detail/service_base.hpp>
 #include <boost/beast/core/detail/is_invocable.hpp>
+#include <boost/asio/associated_cancellation_slot.hpp>
 #include <boost/asio/any_io_executor.hpp>
 #include <boost/asio/associated_cancellation_slot.hpp>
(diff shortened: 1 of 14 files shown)


test/beast/websocket/close.cpp
```diff
@@ -775,7 +775,7 @@ class close_test : public websocket_test_suite
 
             BEAST_EXPECT(ioc.run() == 0);
             BEAST_EXPECT(count == 3);
-            BEAST_EXPECT(ic == 6);
+            BEAST_EXPECT(ic == 3);
```
ashtum, line 778: @klemens-morgenstern Do you have any idea about the reason?

fpelliccioni (commented):
Most of the changes look good and make sense, what concerns me is the change in the test. Although if I think about it better maybe the test values now even make more sense. Maybe it would be better to compare by greater or equal instead of equal. Let's see what @klemens-morgenstern  thinks.

klemens-morgenstern:
Not this release guys. I think deprecating the public `bind_front` and replacing the usage within beast is a good idea. But you'll need to check for things like internal posts in the handler and make sure we're not breaking any thing by introducing recursion. The test change tells me we're missing something, especially since you don't know what causes it.

test/beast/websocket/close.cpp
ashtum: Resolved in: https://github.com/boostorg/beast/pull/2782

PR #2779, opened 2023-12-23
fix: Unhandled Connection Closing in websocker_server_awaitable.cpp

### LINKED ISSUE #2778: WebSocket Server Awaitable Example Error
### Version of Beast

Beast Version: 351

- Compile the [client](https://github.com/boostorg/beast/blob/master/example/websocket/client/awaitable/websocket_client_awaitable.cpp) and [server](https://github.com/boostorg/beast/blob/master/example/websocket/server/awaitable/websocket_server_awaitable.cpp) examples.
- Run the server on any port, with 1 thread. Run the client on the same port as the server.
- Notice: The client code executes as expected but the server returns an error
```bash
Error in session: Operation canceled [system:125 at /usr/include/boost/beast/websocket/impl/stream_impl.hpp:355:13 in function 'bool boost::beast::websocket::stream< <template-parameter-1-1>, <anonymous> >::impl_type::check_stop_now(boost::beast::error_code&)']
```

### All relevant compiler information

Compiler: GCC 11.4.0
Boost Version: 1.84.0

The more information you provide the sooner your issue
can get resolved!

Anefu:
This PR fixes #2778

### DIFF
--- a/example/websocket/server/awaitable/websocket_server_awaitable.cpp
+++ b/example/websocket/server/awaitable/websocket_server_awaitable.cpp
@@ -76,11 +76,18 @@ do_session(stream ws)
             co_await ws.async_write(buffer.data());
 
         }
-        catch(boost::system::system_error & se)
+        catch (const boost::system::system_error& se)
         {
-            if (se.code() != websocket::error::closed)
-                throw;
-
+            if (se.code() == beast::websocket::error::closed || se.code() == boost::asio::error::operation_aborted)
+            {
+                std::cout << "Connection closed by client." << std::endl;
+                co_return;
+            }
+            else
+            {
+                std::cerr << "Error: " << se.what() << std::endl;
+                co_return;
+            }
         }
 }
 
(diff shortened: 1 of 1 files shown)


ashtum:
Thank you for the PR,
The problem seems to be in this line: https://github.com/boostorg/beast/blob/192b7d5a43a7a95dd29df8b08cff635fb14ad9a6/example/websocket/server/awaitable/websocket_server_awaitable.cpp#L81
It continues to loop if `ec == websocket::error::closed` which leads to an operation_aborted error. I would suggest put the  whole loop inside try catch block so it will break out of loop if an exception throws, something like:
```C++
    try
    {
        for(;;)
        {
            // This buffer will hold the incoming message
            beast::flat_buffer buffer;

            // Read a message
            co_await ws.async_read(buffer);

            // Echo the message back
            ws.text(ws.got_text());
            co_await ws.async_write(buffer.data());
        }
    }
    catch(const boost::system::system_error & se)
    {
        if (se.code() != websocket::error::closed)
            throw;
    }
```
And check if exception_ptr is not null (as your current code does).

Anefu:
Thanks for the suggestion.

ashtum (approved):

PR #2784, opened 2024-01-01
Fix write loop in advanced server examples

### LINKED ISSUE #2739: In example: Error and note
### Version of Beast
At Boost 1_83.

I have some comments about this example.
https://www.boost.org/doc/libs/1_83_0/libs/beast/example/advanced/server/advanced_server.cpp

1. Note.

In function
void queue_write(http::message_generator response)
You do push_back first.
You then ensure that no work is done by checking size()==1 for the vector.

Abstracting from the container type, the empty() operation takes the same amount of time as size() or faster. It is more profitable to first check the vector for empty() and store the result in a variable. Then do push_back(). Then, depending on the value of the variable, execute do_write();

2. Error.
In function
bool do_write()

After checking !response_queue_.empty(), you retrieve the element from the vector and start the asynchronous operation.

If you look at the body of the queue_write() function, you can see that it controls the absence of an asynchronous operation by comparing the queue size with 1 (after adding an element).
Which is incorrect.

Example:

Start. queue size=0
1. queue_write(). queue size=1
2. call do_write()
starting an asynchronous operation. queue size=0
3. queue_write(). queue size=1
4. call do_write()
starting an asynchronous operation. queue size=0

We did not wait for the previous asynchronous operation to complete and started a new one.

ashtum:
Fixes #2739

### DIFF
--- a/example/advanced/server-flex-awaitable/advanced_server_flex_awaitable.cpp
+++ b/example/advanced/server-flex-awaitable/advanced_server_flex_awaitable.cpp
@@ -38,8 +38,9 @@
 #include <cstdlib>
 #include <functional>
 #include <iostream>
-#include <memory>
 #include <list>
+#include <memory>
+#include <queue>
 #include <string>
 #include <thread>
 #include <vector>
@@ -499,7 +500,7 @@ class http_session
     }
 
     static constexpr std::size_t queue_limit = 8; // max responses
-    std::vector<http::message_generator> response_queue_;
+    std::queue<http::message_generator> response_queue_;
 
     // The parser is stored in an optional container so we can
     // construct it from scratch it at the beginning of each new message.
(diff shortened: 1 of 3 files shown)


example/advanced/server-flex-awaitable/advanced_server_flex_awaitable.cpp
```diff
@@ -36,6 +36,7 @@
 #include <boost/optional.hpp>
 #include <algorithm>
 #include <cstdlib>
+#include <queue>
```
fpelliccioni, line 39: NIT: alphabetical order or includes

example/advanced/server-flex/advanced_server_flex.cpp
```diff
@@ -28,6 +28,7 @@
 #include <boost/optional.hpp>
 #include <algorithm>
 #include <cstdlib>
+#include <queue>
```
fpelliccioni, line 31: NIT: alphabetical order or includes

fpelliccioni (approved):

fpelliccioni:
🚀

PR #2789, opened 2024-01-05
Improve documentation for `http::message::prepare_payload()`

### LINKED ISSUE #603: prepare_payload doc tidy
> Many reference pages could use some prosaic explanation. For example, my uncertainty concerning message.prepare_payload() might have been alleviated, were it more clear what the function is intended for and what the effect of omitting it would be.

ashtum:
Fixes #603


vinniefalco:
Seems good!

PR #2793, opened 2024-01-10
Fix the issue of paragraph separation in the documentation

### LINKED ISSUE #2792: `zlib::deflate_stream::write` documentation wall of text
### Version of Beast

Boost 1.84

Browse to https://www.boost.org/doc/libs/1_84_0/libs/beast/doc/html/beast/ref/boost__beast__zlib__deflate_stream/write.html

### Details

The paragraph separations in the `zlib::deflate_stream::write` [Doxygen documentation](https://github.com/boostorg/beast/blob/a0a80359fe8b7153c3985ccbbfb0ccf458bd634e/include/boost/beast/zlib/deflate_stream.hpp#L183) have been lost in the conversion to the online Boost documentation, resulting in a giant, unreadable wall of text. 

I've tried with Firefox and Chromium browsers.

image

ashtum:
Fixes #2792

### DIFF
--- a/doc/xsl/custom-overrides.xsl
+++ b/doc/xsl/custom-overrides.xsl
@@ -15,6 +15,8 @@
   <xsl:template mode="convenience-header" match="@file[contains(., 'boost/beast/zlib')]"     >zlib.hpp</xsl:template>
   <xsl:template mode="convenience-header" match="@file"/>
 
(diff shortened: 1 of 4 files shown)


ashtum:
This breaks some of the new comment blocks because the authors have been relying on the fact that paragraphs in the comment block will not separate in the generated HTML pages.

Like:
Screenshot from 2024-01-10 18-42-23

I'll fix them as part of this PR.

PR #2803, opened 2024-01-19
Fix unreachable code warning in `buffers_cat.hpp`

### LINKED ISSUE #2678: unreachable code in asio/buffer.hpp with optimizations
See References #1582 

@vm2mv wrote:

> boost 1.81 and 1.82
> Visual Studio 2022 17.4.3
> 
>     Disabling inline function expansion (/Ob0) moves the warning from asio to beast:
> 
> boost\beast\core\impl\buffers_cat.hpp(206): warning C4702: unreachable code 
> 
>         template<class I>
>         reference operator()(I)
>         {
> ->          return *self.it_.template get<I::value>();
>         }
> 

I'm using boost-beast in combination with VCPKG.

ashtum:
It seems we can't simply assume that the code for `past_end::operator*()` will never compile, as the following code clearly needs to handle that case:
```C++
return mp11::mp_with_index<
    sizeof...(Bn) + 2>(
        it_.index(), // Considering the index is a runtime value, the compiler needs to generate code for when it points to `past_end`
        dereference{*this});
```

Fixes #2678

### DIFF
--- a/include/boost/beast/core/impl/buffers_cat.hpp
+++ b/include/boost/beast/core/impl/buffers_cat.hpp
@@ -52,11 +52,8 @@ public:
 
 #if defined(_MSC_VER) && ! defined(__clang__)
 # define BOOST_BEAST_UNREACHABLE() __assume(false)
-# define BOOST_BEAST_UNREACHABLE_RETURN(v) return v
 #else
 # define BOOST_BEAST_UNREACHABLE() __builtin_unreachable()
-# define BOOST_BEAST_UNREACHABLE_RETURN(v) \
-    do { __builtin_unreachable(); return v; } while(false)
 #endif
 
 #ifdef BOOST_BEAST_TESTS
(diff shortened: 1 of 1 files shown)


vinniefalco:
Interesting, is this why we keep getting that damn warning !?

ashtum:
> Interesting, is this why we keep getting that damn warning !?

I believe so. The reason we are getting those warnings is that the returned `net::mutable_buffer` from the `past_end::operator*` function (which is marked unreachable) implicitly converts to `net::const_buffer`. For example, the following generates the warning:
https://godbolt.org/z/83G5f4Wx7
```C++
auto buf = buffers_cat(net::const_buffer{}, net::const_buffer{});
auto it  = buf.begin();
*it;
```
However, this doesn't:
https://godbolt.org/z/WePad646W
```C++
auto buf = buffers_cat(net::mutable_buffer{}, net::mutable_buffer{});
auto it  = buf.begin();
*it;
```

This is quite comparable to this example: 
https://godbolt.org/z/8fMroWh54
```C++
[[noreturn]] const char * f1()
{
  __assume(false);
}

const char * f2()
{
  return f1(); // Fine
}

std::string f3()
{
  return f1(); // warning C4702: unreachable code
}
``` 

MSVC is correctly warning us that we are doing something after unreachable code (in this case, the implicit conversion is what we are doing).

PR #2861, opened 2024-05-08
fix asserts, it's possible that we reach code only with 3 symbols

ruz:
This happens when:
* chunked encoding
* body is empty
* chunk has no extensions
* and put calls are really weird

### DIFF
--- a/include/boost/beast/http/impl/basic_parser.ipp
+++ b/include/boost/beast/http/impl/basic_parser.ipp
@@ -701,9 +701,12 @@ parse_chunk_header(char const*& p0,
     }
     else
     {
-        BOOST_ASSERT(n >= 5);
+        BOOST_ASSERT(n >= 3);
         if(f_ & flagExpectCRLF)
+        {
+            BOOST_ASSERT(n >= 5);
             BOOST_VERIFY(parse_crlf(p));
+        }
         std::uint64_t size;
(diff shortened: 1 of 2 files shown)


ashtum:
The minimum valid chunked body is `0\r\n\r\n`, do you what to treat `0\r\n` as a valid body?

ruz:
No, I don't want to. This is a streamed parser, imagine put calls with the following sequence: '0\r\n', '\r' and '\n'. The current code will fail with the assert on the second put as n will be equal to 4.

What I want to do is to correct the assert and make it valid. If flagExpectCRLF is set then n should be >= 5, otherwise 3 is possible and ok. Code that goes after, searches for CRLFCRLF (eom) sequence and correctly returns need_more.

ashtum:
> What I want to do is to correct the assert and make it valid. If flagExpectCRLF is set then n should be >= 5, otherwise 3 is possible and ok. Code that goes after, searches for CRLFCRLF (eom) sequence and correctly returns need_more.

You are correct; the current code returns the error_code `error::need_more`, which expects more data in the next call, which can be as little as one byte. Thank you very much for reporting this and for your PR; I'll add the related test cases and merge this.

@vinniefalco

vinniefalco:
LGTM

ruz:
thanks

PR #2866, opened 2024-05-13
Handle expired timers in basic_stream transfer_op

ashtum:
Previously, the code only handled expired timers when the buffer was empty.

### DIFF
--- a/include/boost/beast/core/impl/basic_stream.hpp
+++ b/include/boost/beast/core/impl/basic_stream.hpp
@@ -313,6 +313,26 @@ public:
     {
         BOOST_ASIO_CORO_REENTER(*this)
         {
+            // apply the timeout manually, otherwise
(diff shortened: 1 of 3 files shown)


include/boost/beast/core/impl/basic_stream.hpp
```diff
@@ -326,13 +346,6 @@ class transfer_op
 
                     async_perform(0, is_read{});
```
ashtum, line 347: @vinniefalco I believe the use of `async_perform(0, is_read{})` here has been to reduce the number of template instantiations. If that's the case, I can replace the use of `net::dispatch` with `async_perform(0, is_read{})` for the same reason.

vinniefalco: seems OK I guess...I think you know more about this than me now.

PR #2875, opened 2024-05-30
Deprecate beast::flat_stream and beast::ssl_stream

ashtum:


vinniefalco:
Commit messages should start with nouns:

std::bind is superfluous in some examples

ssl_stream does not use flat_stream

net::ssl::stream is canonical in examples

net::ssl::stream is canonical in snippets

ssl_stream and flat_stream are deprecated

PR #2878, opened 2024-06-02
Add 418 "I'm a teapot" HTTP status

alex65536:
Though this HTTP status is just a joke, it is supported by numerous HTTP implementations, including (but not limited to):

* [Python](https://github.com/python/cpython/blob/f79ffc879b919604ed5de22ece83825006cf9a17/Lib/http/__init__.py#L130)
* [POCO](https://github.com/pocoproject/poco/blob/bc29f30e5cf035e860b3d2c1fc5df570f01513ea/Net/include/Poco/Net/HTTPResponse.h#L90)
* [Golang](https://pkg.go.dev/net/http#StatusTeapot)
* [Rust `actix-web` crate](https://docs.rs/actix-web/4.6.0/actix_web/http/struct.StatusCode.html#associatedconstant.IM_A_TEAPOT)

So, I think it's also a good idea to support this HTTP status code in Beast.

Also, C++ is occasionally used for embedded programming, such as smart teapots, and I think it also justifies adding this status code :) :)

### DIFF
--- a/include/boost/beast/http/impl/status.ipp
+++ b/include/boost/beast/http/impl/status.ipp
@@ -72,6 +72,7 @@ int_to_status(unsigned v)
     case status::unsupported_media_type:
     case status::range_not_satisfiable:
     case status::expectation_failed:
+    case status::i_am_a_teapot:
     case status::misdirected_request:
     case status::unprocessable_entity:
     case status::locked:
@@ -178,6 +179,7 @@ obsolete_reason(status v)
     case status::unsupported_media_type:                return "Unsupported Media Type";
     case status::range_not_satisfiable:                 return "Range Not Satisfiable";
     case status::expectation_failed:                    return "Expectation Failed";
+    case status::i_am_a_teapot:                         return "I'm a teapot";
     case status::misdirected_request:                   return "Misdirected Request";
     case status::unprocessable_entity:                  return "Unprocessable Entity";
     case status::locked:                                return "Locked";
--- a/include/boost/beast/http/status.hpp
+++ b/include/boost/beast/http/status.hpp
@@ -80,6 +80,7 @@ enum class status : unsigned
     unsupported_media_type              = 415,
     range_not_satisfiable               = 416,
     expectation_failed                  = 417,
+    i_am_a_teapot                       = 418,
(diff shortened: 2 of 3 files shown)


ashtum:
We have recently updated the status code list to conform with the boring IANA registry (https://github.com/boostorg/beast/pull/2811/commits/180d9977011203fe53128441638c64528f1c67f8). So, I'm not sure if it's a good idea to break it.

alex65536:
> I'm not sure if it's a good idea to break it.

Well, I think that having more status codes is better than having less of them. Also, status code 418 is reserved, so it's quite unlikely that it will receive any other meaning:

> Therefore, the 418 status code is reserved in the IANA HTTP Status Code Registry. This indicates that the status code cannot be assigned to other applications currently.

ashtum:
> [[RFC2324](https://www.rfc-editor.org/rfc/rfc9110.html#RFC2324)] was an April 1 RFC that lampooned the various ways HTTP was abused; one such abuse was the definition of an application-specific 418 status code, which has been deployed as a joke often enough for the code to be unusable for any future use.
> Therefore, the 418 status code is reserved in the IANA HTTP Status Code Registry. This indicates that the status code cannot be assigned to other applications currently. If future circumstances require its use (e.g., exhaustion of 4NN status codes), it can be re-assigned to another use.

I guess we can add it then :)

> Well, I think that having more status codes is better than having less of them

Users can use the corresponding integral value for the status codes that are not in the list. The problem with supporting non-standard status codes is that they overlap, and we can't add all of them.

alex65536:
> The problem with supporting non-standard status codes is that they overlap, and we can't add all of them.

Well, given that this code is reserved to prevent it being reused, the overlap will be highly unlikely.

vinniefalco:
> I think that having more status codes is better than having less of them.

Interesting, because I think the exact opposite. Fewer choices means less confusion, especially if the extra choices add no value.

ashtum:
> Well, given that this code is reserved to prevent it from being reused, the chance of overlap is highly unlikely.

I didn't mean 418; I was referring to the reason why we don't add non-standard status codes in general, like the ones Nginx uses, for example.

PR #2881, opened 2024-06-05
read_size_hint does not exceed read_message_max

ashtum:
Fixes #2879


vinniefalco:
Good commit message :)

PR #2886, opened 2024-06-12
`const`-qualify trivial getters

sameer-here:
None of these methods modify `serializer` members, so they ought to be `const`.  This undoes https://github.com/boostorg/beast/commit/8a3df73ffc7b2a1bf89f624fdd9a26c9dcf1fb5c

### DIFF
--- a/include/boost/beast/http/serializer.hpp
+++ b/include/boost/beast/http/serializer.hpp
@@ -248,7 +248,7 @@ public:
 
     /// Returns the serialized buffer size limit
     std::size_t
-    limit()
+    limit() const
     {
         return limit_;
     }
@@ -274,7 +274,7 @@ public:
     /** Returns `true` if we will pause after writing the complete header.
     */
     bool
-    split()
+    split() const
     {
         return split_;
     }
@@ -298,7 +298,7 @@ public:
         serialized header octets have been retrieved.
     */
     bool
-    is_header_done()
+    is_header_done() const
     {
         return header_done_;
     }


ashtum:
I believe the changes in https://github.com/boostorg/beast/commit/8a3df73ffc7b2a1bf89f624fdd9a26c9dcf1fb5c have been intentional because these functions might return different values depending on the state of the serializer object during write operations, making them logically non-const. @vinniefalco, could you please shed some light on this?

sameer-here:
> I believe the changes in https://github.com/boostorg/beast/commit/8a3df73ffc7b2a1bf89f624fdd9a26c9dcf1fb5c were intentional because these functions might return different values depending on the state of the serializer object during write operations, making them logically non-const.

If these methods are "logically non-const" I'd prefer it be explained by documentation rather dropping the `const`, since the methods don't modify the members of the class.

vinniefalco:
> None of these methods modify serializer members, so they ought to be const

That may be true, but you as a user are not supposed to know or care about that. All you need to know is that the functions may not be called concurrently, this is by design. Labeling them const would imply that they are thread-safe. While it is true that the current implementation is thread-safe, that is an implementation detail. The library reserves the right to change the implementation.

As a user, you should not care whether these members are const or not. If there is a specific problem that your patch solves, then please show us the example code or describe the scenario.

sameer-here:
> Labeling them const would imply that they are thread-safe.

My impression is that `const` is orthogonal to thread-safe.  I would not expect thread safety from a method just because it is marked `const`.  I would expect that it is safe to call the `const` method from multiple threads, but to call a non-const method and a `const` method from different threads would cause a data race, unless the documentation specified that the class was thread safe.  If my understanding of `const` is wrong, I would be happy to learn!

> As a user, you should not care whether these members are const or not. If there is a specific problem that your patch solves, then please show us the example code or describe the scenario.

I tend to agree that there's no massive issue caused by these being `non-const`.  I did have some issues when `is_done` was not marked `const`, but `is_header_done()`, `split()`, and `limit()` aren't problematic.  This is just a minor quality-of-life patch.

vinniefalco:
https://stackoverflow.com/questions/14127379/does-const-mean-thread-safe-in-c11

sameer-here:
> [17.6.5.9/3] A C++ standard library function shall not directly or indirectly modify objects (1.10) accessible by threads other than the current thread unless the objects are accessed directly or indirectly via the function’s non-const arguments, including this

I think that stackoverflow answer above puts its pretty concisely, that

> This means that the Standard Library won't introduce a data race as long as operations on const objects of your own types either
> 1. Consist entirely of reads --that is, there are no writes--; or
> 2. Internally synchronizes writes.

`const`-qualifying a getter would fall under the first point.  The `const` doesn't imply the non-`const` methods of the class are thread-safe, merely that concurrently calling the `const` operations is safe.

vinniefalco:
> The `const` doesn't imply the non-`const` methods of the class are thread-safe, merely that concurrently calling the `const` operations is safe.

Yes, exactly. This is precisely why I do not want them to be marked `const`: I do not want to promise to the user that calling those members concurrently is safe. It is true that the current implementation makes it safe today, but the library should have the freedom to change this in the future.

The principle here is that the API should make as few promises as possible while achieving the intended purpose. Instances of a parser or serializer are never intended to be used concurrently, and there is no need for such. Thus we avoid offering the user the guarantee of safe concurrency even for trivial functions whose current implementations cannot possibly race.

PR #2913, opened 2024-08-01
Parser.consume

klemens-morgenstern:
I want to write a stream API for http. 

I will also need a buffer to reset a previously returned buffer from a body-writer.

### DIFF
--- a/include/boost/beast/http/write.hpp
+++ b/include/boost/beast/http/write.hpp
@@ -56,7 +56,7 @@ namespace http {
 
     @param sr The serializer to use.
 
-    @return The number of bytes written to the stream.
+    @return The number of bytes read from the body.
 
     @throws system_error Thrown on failure.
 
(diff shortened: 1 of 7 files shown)


include/boost/beast/http/write.hpp
```diff
@@ -100,7 +100,7 @@ write_some(
 
     @param ec Set to indicate what error occurred, if any.
 
-    @return The number of bytes written to the stream.
+    @return The number of bytes read from the body.
```
vinniefalco, line 103: This is a breaking API change...

klemens-morgenstern: Yes, and one that is overdue.

vinniefalco: The meaning of the return value used to be different from what it is now. We changed it because of user requests. If we change it again, then we are breaking those users.

ashtum: > The meaning of the return value used to be different from what it is now. We changed it because of user requests. If we change it again, then we are breaking those users.

Wasn't that for `async_read_some`? https://github.com/boostorg/beast/issues/1942

vinniefalco: Maybe. I think you should decide if this is a good change. And if not, what else we might do (if anything).

PR #2919, opened 2024-08-14
Performance improvement

BrianWeed:
By-value sink params can be moved into their final destination


ashtum:
Hi, `std::uint64_t` is a trivial type, so moving it is not more efficient than copying it.

BrianWeed:
Ah, yes, I forgot that optional uses inline storage and not dynamically allocated memory.

Issue #2085, opened 2020-09-10
string_param API change results in crash passing literal 0


ecatmur:
Since #1956 1.74.0:
```
#include <boost/beast.hpp>
int main() {
    boost::beast::http::response<boost::beast::http::string_body> res;
    res.set(boost::beast::http::field::content_length, 0);
}
```

Expected: fails to compile
Actual: compiles, crashes with segmentation fault (null dereference).

https://godbolt.org/z/7G1qqv (middle pane is Boost 1.73, right pane is Boost 1.74).

For us, this only broke a test, but other people might be passing literal 0 as a field value in production code. I'd suggest adding
```
void ${function-name}(field name, std::nullptr_t) = delete;
```
alongside each place `string_param` was changed to `string_view` in 14c3d50b570b1d7bf32453f38be77c759007fdb0.

madmongo1:
The road to Hell is paved with good intentions. Thanks for letting us know.

vinniefalco:
> void ${function-name}(field name, std::nullptr_t) = delete;
Hmm... good idea.

PR #2920, opened 2024-08-17
http::parser reports basic_fields errors

ashtum:


include/boost/beast/http/fields.hpp
```diff
@@ -435,12 +444,14 @@ class basic_fields
         @param name The field name.
 
         @param value The value of the field, as a @ref boost::beast::string_view
+
+        @throws boost::system::system_error Thrown if an error occurs:
```
vinniefalco, line 448: I thought the throws comes before the params?

include/boost/beast/http/fields.hpp
```diff
@@ -435,12 +444,14 @@ class basic_fields
...
+        - If the size of @c value exceeds @ref max_value_size, the
+            error code will be @ref error::header_field_value_too_large.
     */
     void
     insert(field name, string_view const& value);
```
vinniefalco, line 453: string views are typically passed by value

include/boost/beast/http/fields.hpp
```diff
@@ -454,12 +465,14 @@ class basic_fields
         @param name The field name. It is interpreted as a case-insensitive string.
 
         @param value The value of the field, as a @ref boost::beast::string_view
+
+        @throws boost::system::system_error Thrown if an error occurs:
+        - If the size of @c value exceeds @ref max_value_size, the
```
vinniefalco, line 470: Why the dash

include/boost/beast/http/fields.hpp
```diff
@@ -478,6 +491,12 @@ class basic_fields
         comparison, otherwise the behavior is undefined.
 
         @param value The value of the field, as a @ref boost::beast::string_view
+
+        @throws boost::system::system_error Thrown if an error occurs:
+        - If the size of @c name_string exceeds @ref max_name_size, the
```
vinniefalco, line 496: should this use @li or is there a docca bug preventing it?

include/boost/beast/http/fields.hpp
```diff
@@ -486,6 +505,35 @@ class basic_fields
...
+            error code will be @ref error::header_field_name_too_large.
+        - If the size of @c value exceeds @ref max_value_size, the
+            error code will be @ref error::header_field_value_too_large.
+    */
+    void
+    insert(field name, string_view name_string,
+           string_view const& value, error_code& ec);
+
+    void
+    insert(field, string_view, std::nullptr_t, error_code& ec) = delete;
```
vinniefalco, line 535: Why is there a `std::nullptr_t` param? And where's the doc for this?

include/boost/beast/http/impl/fields.hpp
```diff
@@ -548,9 +548,24 @@ void
...
+        string_to_field(sname), sname, value);
+}
+
+template<class Allocator>
+void
+basic_fields<Allocator>::
+insert(
+    field name,
+    string_view sname,
+    string_view const& value,
```
vinniefalco, line 561: this is weird, why is this passed by reference and the other passed by value?

include/boost/beast/http/fields.hpp
```diff
@@ -744,9 +799,21 @@ class basic_fields
...
+    new_element(
+        field name,
+        string_view sname,
+        string_view value,
+        error_code& ec);
+
     element&
-    new_element(field name,
-        string_view sname, string_view value);
+    new_element(
```
vinniefalco, line 810: oh wow.. two functions with the same name, one returning a pointer and the other returning a reference? lol... don't you think they should both return pointers?

include/boost/beast/http/impl/fields.hpp
```diff
@@ -590,8 +584,8 @@ basic_fields<Allocator>::
 set(field name, string_view const& value)
 {
     BOOST_ASSERT(name != field::unknown);
-    set_element(new_element(name, to_string(name),
-        static_cast<string_view>(value)));
+    set_element(
```
vinniefalco, line 587: `set_element` should receive a pointer to keep things consistent

include/boost/beast/http/impl/fields.hpp
```diff
@@ -953,18 +947,22 @@ set_keep_alive_impl(
...
-            (std::numeric_limits<off_t>::max)())
-        BOOST_THROW_EXCEPTION(std::length_error{
-            "field name too large"});
-    if(value.size() + 2 >
-            (std::numeric_limits<off_t>::max)())
-        BOOST_THROW_EXCEPTION(std::length_error{
-            "field value too large"});
+new_element(
+    field name,
+    string_view sname,
```
vinniefalco, line 952: `field name` was a mistake on my part. It should have been `field id`. And then `sname` could just be `name` and we wouldn't have two confusingly similar variables.

include/boost/beast/http/impl/fields.hpp
```diff
@@ -953,18 +947,22 @@ set_keep_alive_impl(
...
+    if(sname.size() > max_name_size)
+    {
+        BOOST_BEAST_ASSIGN_EC(ec, error::header_field_name_too_large);
+        return nullptr;
+    }
+    if(value.size() > max_value_size)
+    {
+        BOOST_BEAST_ASSIGN_EC(ec, error::header_field_value_too_large);
+        return nullptr;
+    }
```
vinniefalco, line 965: Do we also need a limit on the sum of `sname.size()` and `value.size()` (maybe plus delimiters)?

include/boost/beast/http/fields.hpp
ashtum: prevent passing passing the literal 0: https://github.com/boostorg/beast/issues/2085

include/boost/beast/http/impl/fields.hpp
ashtum: I didn't modify the existing signatures. Yes, they are strange, and we need to address them all.

include/boost/beast/http/impl/fields.hpp
ashtum: technically no, because we use an `unit16_t` for storing the size of value. and another `unit16_t` for its offset. so, name and value can have a size of `std::numeric_limits<std::uint16_t>::max() - 2`.

include/boost/beast/http/impl/fields.hpp
ashtum: `set_element`, `insert_element`, and `delete_element` receive references because they don't handle `nullptr`. I believe changing `new_element` to return pointers isn't a good idea either because it never returns `nullptr`. I think changing the name of the other `new_element` overload would be a better option, something like `try_create_new_element` or `try_new_element`.

include/boost/beast/http/fields.hpp
vinniefalco: Having both styles in the same parameter list is weird... I think we should change them in this file.

include/boost/beast/http/fields.hpp
ashtum: I replaced it with @li and it works fine.

include/boost/beast/http/impl/basic_parser.ipp
```diff
@@ -112,26 +112,9 @@ loop:
...
-                }
-                if(p + 3 <= p1)
-                    skip_ = static_cast<
-                        std::size_t>(p1 - p - 3);
-            }
-            goto done;
-        }
         BOOST_ASSERT(! is_done());
         n = static_cast<std::size_t>(p1 - p);
         if(p >= p1)
```
ashtum, line 120: @vinniefalco I couldn't find any scenario that makes `p > p1` possible. Do you remember the rationale behind this line? If the only possibility is `p == p1`, we can completely remove this branch and allow `parse_fields` to be called with `n = 0`

include/boost/beast/http/impl/fields.hpp
vinniefalco: `*_element` are private functions so we can do whatever we want. The rules are relaxed. We should do what is easiest for maintenance. My opinion, pointers are better than having new names and different return types. But it is up to you.

include/boost/beast/http/impl/basic_parser.ipp
vinniefalco: When I write these types of conditions, I make them inequalities on purpose, for no particular reason.

include/boost/beast/http/impl/basic_parser.ipp
```diff
@@ -625,102 +581,53 @@ parse_chunk_header(char const*& p0,
...
-            f_ |= flagExpectCRLF;
-            state_ = state::chunk_body;
-            return;
-        }
-
-        f_ |= flagFinalChunk;
     }
-    else
+
+    auto const eol = find_eol(p + skip_, pend, ec);
```
ashtum, line 604: @vinniefalco Now that we've removed `maybe_need_more`, I'm wondering if it's still worth keeping `skip_`. It's only used in this one place.

vinniefalco: hard for me to say. you will need to decide that

ashtum: It's pessimization when the chunk header can be read in one go, which is indeed the case most of the time because the chunk header is much smaller than the read buffer. I'd prefer to remove it and simplify the code flow and state.

Issue #2928, opened 2024-09-11
Websocket continuation frames dropped if there are leftovers in the buffer from the previous read


pyrist:
Beast version used: 1.85

I'm dealing with an issue where websocket messages over 8192 bytes aren't read completely.

For example, for a websocket message split into 3 continuation frames, the first two arrive and are read. On the third, the first part of the frame is read, and then the rest is ignored.

In websocket/read.hpp line 172, it checks for the frame header. At the beginning of the first two frames, this works as intended: No frame header is there yet in the buffer, so async_read_some is called in line 191, cont is set to true, and reading continues happily along.

However, at the beginning of the third frame, parse_fh checks if there's anything in the buffer yet in line 706 in websocket/implstream_impl.hpp. The buffer it checks (impl.rd_buf) still has some stuff in it from the previous read, so it gets the frame header and doesn't call async_read_some. Because of this, cont is false and the loop is broken. The rest of the websocket message is then not processed. I'm using weird proprietary sockets so I sadly cannot provide code to help reproduce the problem, since the problem doesn't happen with Windows native sockets. I believe the issue is with my weird sockets then sending a piece of the next frame as well in the previous packet.

A possible solution:
Would it not make sense to have a cont = true after line 473 in websocket/read.hpp to catch if someone is using weird sockets like me where you still have a piece of the next frame from the previous async_read_some? Or is there some logic I'm not seeing? As far as I can understand, cont would always be true at 473 anyway, but I might be wrong. Thanks!

pyrist:
Helped by @ashtum to understand the problem isn't in boost beast

PR #2929, opened 2024-09-11
Websocket continuation frames dropped if there are leftovers in the buffer from the previous read

pyrist:
Fix for #2928

### DIFF
--- a/include/boost/beast/websocket/impl/read.hpp
+++ b/include/boost/beast/websocket/impl/read.hpp
@@ -471,6 +471,7 @@ public:
                         bytes_written_ += bytes_transferred;
                         impl.rd_size += bytes_transferred;
                         impl.rd_buf.consume(bytes_transferred);
+                        cont = true;
                     }
                     else
                     {
@@ -514,8 +515,12 @@ public:
                     }
                 }
                 BOOST_ASSERT( ! impl.rd_done );
-                if( impl.rd_remain == 0 && impl.rd_fh.fin )
+                if (impl.rd_remain == 0 && impl.rd_fh.fin) {
                     impl.rd_done = true;
+                }
+                else {
+                    cont = true;
+                }
             }
             else
             {
--- a/include/boost/beast/websocket/impl/read.hpp
+++ b/include/boost/beast/websocket/impl/read.hpp
@@ -471,7 +471,6 @@ public:
                         bytes_written_ += bytes_transferred;
                         impl.rd_size += bytes_transferred;
                         impl.rd_buf.consume(bytes_transferred);
-                        cont = true;
                     }
                     else
                     {


vinniefalco:
Is this a real bug?

pyrist:
> Is this a real bug?

I believe so. It wouldn't make sense to me that the read loop should end after reading part of a header frame when the message and frame aren't finished.

ashtum:
@pyrist, I couldn't reproduce the situation you described with the following code, which tests various message sizes and different points for chopping the input bytes.
```C++
for(std::size_t msg_size = 1 ; msg_size <= 10000; msg_size += 7)
{
    for(std::size_t i = 0 ;; ++i)
    {
        net::io_context ioc;
        stream<test::stream> wsc{ioc};
        stream<test::stream> wss{ioc};
        wsc.next_layer().connect(wss.next_layer());
        wsc.async_handshake(
            "localhost", "/", [](error_code){});
        wss.async_accept([](error_code){});
        ioc.run();
        ioc.restart();
        BEAST_EXPECT(wsc.is_open());
        BEAST_EXPECT(wss.is_open());
        auto sent_msg = std::string(msg_size, 'a');
        wss.write(net::buffer(sent_msg));

        if(i > wsc.next_layer().str().size())
            break; // we have tested all possible chunk sizes

        // steal the input buffer in order to write it in two steps
        std::string bytes = wsc.next_layer().str();
        wsc.next_layer().buffer().clear();

        // first part
        wss.next_layer().write_some(
            net::buffer(bytes.substr(0, i)));

        flat_buffer received_msg;
        wsc.async_read(
            received_msg,
            [](error_code, std::size_t){});

        ioc.poll();

        // second part
        wss.next_layer().write_some(
            net::buffer(bytes.substr(i)));

        ioc.run();
        BEAST_EXPECT(received_msg.size() == msg_size);
    }
}
```

pyrist:
@ashtum Thanks for looking at it. I will look tomorrow to compare between what your code is doing and what is going wrong when I use it and see if I can make something reproducible.

include/boost/beast/websocket/impl/read.hpp
```diff
@@ -514,8 +514,12 @@ class stream<NextLayer, deflateSupported>::read_some_op
                     }
                 }
                 BOOST_ASSERT( ! impl.rd_done );
-                if( impl.rd_remain == 0 && impl.rd_fh.fin )
+                if (impl.rd_remain == 0 && impl.rd_fh.fin) {
                     impl.rd_done = true;
+                }
+                else {
+                    cont = true;
```
ashtum, line 521: `cont` is unrelated to the logic of the read operation itself; it determines whether the completion handler should be scheduled on an executor or executed directly. I suspect there might be an issue in your custom stream type, where an additional dispatch can make it work.

pyrist:
@ashtum Thanks so much for the help. I used your example to compare what was going on with our sockets. You're right, our socket failing to call an additional dispatch is what's actually going on here, and setting cont to true just forces beast to do the socket's job for it.

pyrist:
@ashtum Thanks so much for the help. I used your example to compare what was going on with our sockets. You're right, our socket failing to call an additional dispatch is what's actually going on here, and setting cont to true just forces beast to do the socket's job for it.

PR #2940, opened 2024-10-15
Add forward declaration headers

ashtum:


ashtum:
@vinniefalco, this PR replaces `message.hpp` with `message_fwd.hpp` in all body types, eliminating the need for a separate forward declaration header for them. This change could break user code that includes only a body header like `string_body.hpp` and attempts to construct a message. However, this is unlikely since users typically include `parser.hpp`, `serializer.hpp`, or perform read/write operations, where `message.hpp` is already included in those headers, preventing issues. An alternative would be to add a forward declaration header for each body type, but that approach would be messy and cumbersome to use.
What do you think?

PR #2980, opened 2025-02-08
websockets: Count pings from server as activity for idle_timeout 

ashtum:
@xim, I merged your commit here because I couldn't reopen #2718.


xim:
# 🥳

PR #2992, opened 2025-03-04
Fix link to async_teardown in teardown document

enolan-maystreet:
Previously, this link incorrectly pointed to teardown instead of async_teardown.


enolan-maystreet:
The documentation preview shows that the link to `async_teardown` is fixed: https://2992.beast.prtest.cppalliance.org/libs/beast/doc/html/beast/using_websocket/teardown.html

PR #3020, opened 2025-07-23
Add missing "return"

DockedFerret800:
Hi. Aren't we supposed to return in case of an error? Similarly to what we do in other examples, for example in `websocket_server_async_local.cpp`.

### DIFF
--- a/example/websocket/server/async-ssl/websocket_server_async_ssl.cpp
+++ b/example/websocket/server/async-ssl/websocket_server_async_ssl.cpp
@@ -151,7 +151,7 @@ public:
             return;
 
         if(ec)
-            fail(ec, "read");
+            return fail(ec, "read");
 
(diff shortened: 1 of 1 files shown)


ashtum:
@DockedFerret800, thank you for the PR. There are similar issues in a few other examples as well. Could you please go ahead and fix those as part of this commit?

DockedFerret800:
> @DockedFerret800, thank you for the PR. There are similar issues in a few other examples as well. Could you please go ahead and fix those as part of this commit?

Sure

ashtum:
Thanks, please squash them into a single commit if possible.

DockedFerret800:
> Commit messages are even better when they start with nouns instead of verbs:
> 
> "examples use return statements where needed"

I'll remember that for the future. Thank you!

PR #3027, opened 2025-08-30
Annotate facade header with IWYU export annotation

hzeller:
Without that annotation, tools such as `clang-tidy` or the `clangd` language server (as well as many other tools) will complain about headers not directly providing a symbol if users include `beast.hpp`; With this annotation, they know.

Documentation IWYU
https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-begin_exportsend_exports

Documentation llvm include cleaner/clang-tidy/clangd https://clangd.llvm.org/design/include-cleaner#iwyu-pragmas

### DIFF
--- a/include/boost/beast.hpp
+++ b/include/boost/beast.hpp
@@ -10,6 +10,8 @@
 #ifndef BOOST_BEAST_HPP
 #define BOOST_BEAST_HPP
 
+// IWYU pragma: begin_exports
+
 #include <boost/beast/core/detail/config.hpp> // must come first
 
 #include <boost/beast/core.hpp>
@@ -17,5 +19,7 @@
 #include <boost/beast/version.hpp>
 #include <boost/beast/websocket.hpp>
 #include <boost/beast/zlib.hpp>
+
+// IWYU pragma: end_exports
 
 #endif


sehe:
Does IWYU not honour -isystem includes (e.g. via CMake SYSTEM includes)?

hzeller:
`-isystem` does not make a difference here.

Often, warnings for `-isystem`-provided headers would be suppressed.

But here we have a different situation, and want to guide warnings in the project itself: IWYU marks includes relevant or not given if they provide a symbol directly in a particular header (and by default it does not follow further includes). That makes sure that symbols don't come from transitive dependencies (accidental visibility of symbols due to the textual concatenation of C headers, the very thing that IWYU wants to help prevent).

So if we have a header that is indeed meant to provide a symbol one nest-level deeper, we can annotate it like what I've done here.

vinniefalco:
This looks benign

sehe:
> But here we have a different situation

Thank you. It's more clear to me now that this is not "silencing warnings" but annotating the _facade header(s)_ specifically.

ashtum:
@hzeller, I'm using clangd-20 and clang-tidy-20 with the `-misc-include-cleaner` check, but I wasn't able to reproduce the warning you mentioned. Could you please provide more details on how to reproduce the issue?

ashtum:
I’m closing this for now as I wasn’t able to reproduce the reported issue. I’d appreciate it if the OP could provide a bit more information so we can investigate further.

PR #3042, opened 2025-10-09
Parser only stores standard trailer fields

ashtum:


edtanous:
This patch appears to have dropped a few fields that broke code I maintain in a similar way to what https://github.com/gibbz00/simple_http/commit/50d887bf38ed5d48ac5a45fdfa89992ea75c495c had to fix.  Is there a reason HTTP2-Settings was removed?  It's listed on IANA as a valid field.  The commit message doesn't specify well why it was removed.  The commit message on db31a880525fe84b0e17b80015049363106c5b61 doesn't really specify why.

ashtum:
> The commit message on [db31a88](https://github.com/boostorg/beast/commit/db31a880525fe84b0e17b80015049363106c5b61) doesn't really specify why.

The previous field list included many unrelated entries (primarily from the SMTP protocol). The commit cleaned up the list by referencing two sources:

- https://www.iana.org/assignments/http-fields/http-fields.xhtml
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers

As a result, the number of fields has been reduced from 357 to 129.

> This patch appears to have dropped a few fields that broke code I maintain in a similar way to what [gibbz00/simple_http@50d887b](https://github.com/gibbz00/simple_http/commit/50d887bf38ed5d48ac5a45fdfa89992ea75c495c) had to fix. Is there a reason HTTP2-Settings was removed? It's listed on IANA as a valid field. The commit message doesn't specify well why it was removed.

Sorry for the inconvenience. Unfortunately, `HTTP2-Settings` is not listed on mozilla.org and is marked as obsolete in the IANA registry, which is why it was removed. This appears to have been a mistake. I’ve added `HTTP2-Settings` back in a new commit, and it will be included in the upcoming release.

Just for the record, the code in https://github.com/gibbz00/simple_http/commit/50d887bf38ed5d48ac5a45fdfa89992ea75c495c can be simplified (even without the enumerator) to:

```cpp
h2_setting = headers["HTTP2-Settings"];
```

This works because `http::header::operator[]` returns an empty `string_view` if the field does not exist.

edtanous:
I appreciate you looking into it in this commit.  The deprecation was the reason I asked.  It wasn't clear what criteria was used.  This commit was what we did to workaround , which was more or less what you suggested. https://github.com/openbmc/bmcweb/commit/240f5555992f58755eed8bb9c2c1e67d7539778b

We also were broken by content_transfer_encoding being removed, but that appears to have never been valid for http, so we're cleaning that up as well.

I appreciate the work you do.

PR #3053, opened 2025-11-05
Fix cancellation slot retrieval in saved_handler::emplace.

biljazovic:
Bug introduced in a3d00a6d0c551026dc7e763f0297f974d3b062a4

### DIFF
--- a/include/boost/beast/core/impl/saved_handler.hpp
+++ b/include/boost/beast/core/impl/saved_handler.hpp
@@ -153,7 +153,7 @@ emplace(Handler&& handler, Allocator const& alloc,
     auto tmp = boost::exchange(s.p, nullptr);
     p_ = tmp;
 
-    auto c_slot = net::get_associated_cancellation_slot(handler);
+    auto c_slot = net::get_associated_cancellation_slot(tmp->v_.h);
     if (c_slot.is_connected())
     {
         class cancel_op


ashtum:
Thank you. I'm wondering how our CI didn't catch this at all. Did you find it in a sanitizer build?

biljazovic:
UBSan found it here https://github.com/biljazovic/async-mqtt5/actions/runs/19098896699/job/54565719299

websocket::stream::async_close is in parallel group with timer wait, see here https://github.com/boostorg/mqtt5/blob/develop/include/boost/mqtt5/impl/shutdown_op.hpp and here https://github.com/boostorg/mqtt5/blob/develop/include/boost/mqtt5/websocket.hpp

ashtum:
Interesting, Our Calng-18 Ubsan build was cool about it: https://drone.cpp.al/boostorg/beast/1283/36/2
Probably because most of the handlers in the tests don't have an associated cancellation slot to trigger it.

ashtum:
Looks like there are enough tests to cover that: https://github.com/boostorg/beast/blob/5d6edfabfa9e2532d7a5e4432c05763f12547483/test/beast/core/saved_handler.cpp#L141
I believe the difference is that in a regular handler, the cancellation slot type remains valid when handler is moved, whereas in asio's `parallel_group_op_handler`, the cancellation slot [is retrieved from the signal on demand](https://github.com/boostorg/asio/blob/f23e04f1ef1e0eae025ecfaa6ddfc6d5f76709f6/include/boost/asio/experimental/impl/parallel_group.hpp#L198-L201), which is what triggered the sanitizer. I should say we were really lucky :)

PR #3056, opened 2025-11-12
Clear error_code on entry in win32_unicode_path constructor

### LINKED ISSUE #3035: win32_unicode_path fails to clear ec.
If there is no failure ec remains unchanged (not cleared).

```cpp
    win32_unicode_path(const char* utf8_path, error_code& ec) {
        int ret = mb2wide(utf8_path, static_buf_.data(),
            static_buf_.size());
        if (ret == 0)
        {
            int sz = mb2wide(utf8_path, nullptr, 0);
            if (sz == 0)
            {
                ec.assign(boost::winapi::GetLastError(),
                    system_category());
                return;
            }
            dynamic_buf_.resize(sz);
            int ret2 = mb2wide(utf8_path,
                dynamic_buf_.data(),
                dynamic_buf_.size());
            if (ret2 == 0)
            {
                ec.assign(boost::winapi::GetLastError(),
                    system_category());
                return;
            }
        }
    }
```
https://github.com/boostorg/beast/blob/develop/include/boost/beast/core/detail/win32_unicode_path.hpp#L31-L54

ssam18:
Fixes #3035

### Issue

The `win32_unicode_path` constructor accepts an `error_code&` parameter but wasn't clearing it on successful path conversion. This led to callers retaining stale error values even when operations succeeded.

### Problem

As reported in #3035, when using `file.open()` with a pre-set error code (e.g., initialized to `error::not_found` as a default), the error would persist even after successful file operations. This is inconsistent with std::filesystem conventions where error_code parameters are cleared on success.

### Solution  

Added `ec = {};` at the beginning of the constructor to ensure the error_code is always in a known state - either cleared for success or properly set for failure.

This aligns with:
- The std::filesystem API convention (e.g., `std::filesystem::create_directory`)
- The recommendation from @vinniefalco to clear `ec` immediately upon entry to avoid these kinds of issues
- Beast's own `file_win32::open` which already clears ec on line 222

### Testing

Windows-specific change. The fix is minimal and follows established error handling patterns in the codebase.

### DIFF
--- a/include/boost/beast/core/detail/win32_unicode_path.hpp
+++ b/include/boost/beast/core/detail/win32_unicode_path.hpp
@@ -29,6 +29,7 @@ class win32_unicode_path
 
 public:
     win32_unicode_path(const char* utf8_path, error_code& ec) {
+        ec = {};
         int ret = mb2wide(utf8_path, static_buf_.data(),
             static_buf_.size());
         if (ret == 0)
--- a/include/boost/beast/core/impl/file_posix.ipp
+++ b/include/boost/beast/core/impl/file_posix.ipp
@@ -288,12 +288,13 @@ read(void* buffer, std::size_t n, error_code& ec) const
         if(result == 0)
         {
             // short read
-            return nread;
+            break;
         }
         n -= result;
         nread += result;
         buffer = static_cast<char*>(buffer) + result;
     }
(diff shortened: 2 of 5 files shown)


ashtum:
Addressed in #3103.

PR #3066, opened 2026-01-07
Avoid tyecast warnings in `inflate_with_eb`

pps83:


ashtum:
The logic is correct, `rd_eb_consumed` can never be greater than 4:
```C++
        const std::uint8_t eb[4] = { 0x00, 0x00, 0xff, 0xff };
        zs.next_in = eb + pmd_->rd_eb_consumed;
        zs.avail_in = sizeof(eb) - pmd_->rd_eb_consumed;
        inflate(zs, ec);
        pmd_->rd_eb_consumed += zs.total_in;
        BOOST_ASSERT(pmd_->rd_eb_consumed <= sizeof(eb));
        if(ec == zlib::error::need_buffers)
            ec.clear();
```
Changing `std::uint8_t` to `std::size_t` would silence the warning, but it does not appear to provide additional safety. If the value were ever to exceed 4, the behavior would already be undefined. For that reason, adding a `static_cast` on the following line seems like a more appropriate approach:
```C++
        pmd_->rd_eb_consumed += zs.total_in;
```

PR #3068, opened 2026-01-16
Add C++20 module support

mikomikotaishi:
This PR adds support for C++20 modules through CMake, as `boost.beast`. The exported classes are those listed in the [reference documentation](https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/quickref.html).

This is hopefully in line with other Boost libraries which have begun offering modules:
* [`boost.any`](https://www.boost.org/libs/any)
* [`boost.pfr`](https://www.boost.org/libs/pfr)
* [`boost.regex`](https://www.boost.org/libs/regex)
* [`boost.type_index`](https://www.boost.org/libs/type_index)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ if (BOOST_BEAST_IS_ROOT)
 endif ()
 option(BOOST_BEAST_BUILD_TESTS "Build boost::beast tests" ${BUILD_TESTING})
 option(BOOST_BEAST_BUILD_FUZZERS "Build boost::beast fuzzers" OFF)
+option(BOOST_BEAST_EXPERIMENTAL_MODULE "Build experimental boost.beast C++20 module" OFF)
 option(BOOST_BEAST_BUILD_EXAMPLES "Build boost::beast examples" ${BOOST_BEAST_IS_ROOT})
 option(BOOST_BEAST_WARNINGS_AS_ERRORS "Treat warnings as errors" OFF)
 
@@ -161,6 +162,15 @@ endif ()
 #-------------------------------------------------
 if (BOOST_BEAST_BUILD_TESTS)
     add_subdirectory(test)
+endif ()
+
+#-------------------------------------------------
+#
+# Module
+#
+#-------------------------------------------------
+if (BOOST_BEAST_EXPERIMENTAL_MODULE)
+    add_subdirectory(modules)
 endif ()
 
 #-------------------------------------------------
--- a/modules/boost_beast.cppm
+++ b/modules/boost_beast.cppm
@@ -1,0 +1,250 @@
+//
+// Copyright (c) 2016-2026 Vinnie Falco (vinnie dot falco at gmail dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/beast
+//
+
+module;
+
+#include <boost/beast.hpp>
+#include <boost/beast/core/buffer_traits.hpp>
+#include <boost/beast/_experimental/http/icy_stream.hpp>
+#include <boost/beast/_experimental/test/error.hpp>
+#include <boost/beast/_experimental/test/fail_count.hpp>
+#include <boost/beast/_experimental/test/handler.hpp>
+#include <boost/beast/_experimental/test/stream.hpp>
+
+export module boost.beast;
+
+export namespace boost::beast {
+    using boost::beast::async_base;
+    using boost::beast::basic_stream;
+    using boost::beast::file;
+    using boost::beast::file_mode;
+    #if BOOST_BEAST_USE_POSIX_FILE
+    using boost::beast::file_posix;
+    #endif
+    using boost::beast::file_stdio;
+    #if BOOST_BEAST_USE_WIN32_FILE
+    using boost::beast::file_win32;
+    #endif
+    using boost::beast::flat_stream;
+    using boost::beast::iequal;
+    using boost::beast::iless;
+    using boost::beast::rate_policy_access;
+    using boost::beast::saved_handler;
+    using boost::beast::simple_rate_policy;
+    using boost::beast::static_string;
+    using boost::beast::stable_async_base;
+    using boost::beast::string_view;
+    using boost::beast::tcp_stream;
+    using boost::beast::unlimited_rate_policy;
+
+    using boost::beast::condition;
+    using boost::beast::error;
+    using boost::beast::file_mode;
+    using boost::beast::role_type;
+
+    using boost::beast::allocate_stable;
+    using boost::beast::async_detect_ssl;
+    using boost::beast::beast_close_socket;
+    using boost::beast::bind_front_handler;
+    using boost::beast::bind_handler;
+
+    template <class Socket>
+    inline void close_socket(Socket& sock) {
+        boost::beast::close_socket(sock);
+    }
+
+    using boost::beast::detect_ssl;
+    using boost::beast::generic_category;
+    using boost::beast::get_lowest_layer;
+    using boost::beast::iequals;
+    using boost::beast::to_static_string;
+
+    using boost::beast::executor_type;
+    using boost::beast::has_get_executor;
+    using boost::beast::is_async_read_stream;
+    using boost::beast::is_async_stream;
+    using boost::beast::is_async_write_stream;
+    
+    using boost::beast::is_file;
+    using boost::beast::is_sync_read_stream;
+    using boost::beast::is_sync_stream;
+    using boost::beast::is_sync_write_stream;
+    // using boost::beast::lowest_layer_type;
+
+    // using boost::beast::ssl_stream;
+
+    namespace errc = boost::beast::errc;
+    using boost::beast::error_category;
+    using boost::beast::error_code;
+    using boost::beast::error_condition;
+    using boost::beast::system_category;
+    using boost::beast::system_error;
+
+    using boost::beast::basic_flat_buffer;
+    using boost::beast::basic_multi_buffer;
+    // using boost::beast::buffer_ref;
+    using boost::beast::buffered_read_stream;
+    using boost::beast::buffers_adaptor;
+    using boost::beast::buffers_cat_view;
+    using boost::beast::buffers_prefix_view;
+    using boost::beast::buffers_prefix_view;
+    using boost::beast::buffers_suffix;
+    using boost::beast::flat_buffer;
+    using boost::beast::flat_static_buffer;
+    using boost::beast::flat_static_buffer_base;
+    using boost::beast::multi_buffer;
+    using boost::beast::static_buffer;
+    using boost::beast::static_buffer_base;
+
+    using boost::beast::async_write;
+
+    template <class BufferSequence>
+    [[nodiscard]]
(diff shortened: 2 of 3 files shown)


ashtum:
Sorry for the late reply, and thank you @mikomikotaishi for your PR.

There is ongoing research into making Boost consumable as C++20 modules, led by Rubén Pérez. The latest updates are available in his blog post: https://anarthal.github.io/cppblog/modules3

Your work does not appear to be based on that effort, and if I’m not mistaken, it technically couldn’t be, since it requires modularizing libraries in dependency order. Given that, are you proposing this PR as a temporary solution? If so, as a Beast user, what specific problem does this solve for you? Is this something that could instead be done at the application level as a workaround?

mikomikotaishi:
Hi,

While I haven't been familiar with that blog post prior nor did I explicitly copy its formula to build this PR, I do think that the contents of my PR resemble the following module interface that was described:
```cpp


// File: boost_mp11.cppm
module; // Global module fragment
#define BOOST_MP11_INTERFACE_UNIT // We want headers to actually declare entities
                                  // No BOOST_CONFIG_SKIP_IMPORT_STD: import std is fine in the global module fragment
#include <cassert>        // Some standard library headers need to be included for their macros
#include <boost/mp11.hpp> // All entities are attached to the global module.

export module boost.mp11;

// List all symbols we want to export
export namespace boost::mp11 {
using mp11::list;
}
```
This technique has pros and cons, and these can be ironed out as more users adopt or give feedback on the feature. Nonetheless, when you say "as a temporary solution", I think that this does provide some structure or means to use the library as a module, and this can be developed or made to conform more with the blog post's design goals, so I'd be happy to see this provided for the time being.

A lot of my projects have been using modules as opposed to headers, and I find that once build system configuring and tooling is correctly configured it is nicer to use than traditional headers. While many things could be done at the application level (I've done this a lot, for example with asio/boost::asio, nlohmann::json, and other libraries, etc.), this feature does seem to have interest from others, such as building a better environment for module adoption where popular libraries can be easily supported with modules, and thus by just providing it with first-class support we would be able to make a lot of lives easier.

mikomikotaishi:
I wouldn't mind working towards number 2, but I'm not very skilled with CI manipulation and have worked with it very seldom.

import std is very difficult to actually get working, so much so that I have my own "fake" std module just because I have never managed to get the official one to work on CMake.

anarthal:
I feel you with `import std`. I don't know the exact internals of Beast CIs, but you might get some inspiration from the ones I'm writing for mp11 (I don't have gcc working yet though): https://github.com/anarthal/mp11/blob/feature/cxx20-modules/.github/workflows/ci.yml

mikomikotaishi:
cc. @ashtum

module/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.28)
+
+add_library(boost_beast_module)
+
+target_sources(boost_beast_module
+    PUBLIC
+        FILE_SET CXX_MODULES FILES
+            boost.cppm
```
ChuanqiXu9: beast.cppm ?

CMakeLists.txt
```diff
@@ -163,6 +164,15 @@ if (BOOST_BEAST_BUILD_TESTS)
     add_subdirectory(test)
 endif ()
 
+#-------------------------------------------------
+#
+# Modules
+#
+#-------------------------------------------------
+if (BOOST_BEAST_BUILD_MODULES)
+    add_subdirectory(modules)
```
ChuanqiXu9: module

ChuanqiXu9 (commented):
I am trying to use boost/beast as a module in personal projects. I find the PR and I want to try it locally... but it seems like it can't be used directly...

module/beast.cppm
```diff
@@ -0,0 +1,254 @@
...
+    using boost::beast::get_lowest_layer;
+    using boost::beast::iequals;
+    using boost::beast::to_static_string;
+
+    using boost::beast::executor_type;
+    using boost::beast::has_get_executor;
+    using boost::beast::is_async_read_stream;
+    template <class T>
+    using is_async_stream = std::integral_constant<bool,
+        is_async_read_stream<T>::value && is_async_write_stream<T>::value>;
```
ChuanqiXu9: This definition is duplicated with "include/boost/beast/core/stream_traits.hpp:422".

module/beast.cppm
```diff
@@ -0,0 +1,254 @@
...
+    using boost::beast::is_async_write_stream;
+    
+    using boost::beast::is_file;
+    using boost::beast::is_sync_read_stream;
+    template <class T>
+    using is_sync_stream = std::integral_constant<bool,
+        is_sync_read_stream<T>::value && is_sync_write_stream<T>::value>;
```
ChuanqiXu9: ditto

module/beast.cppm
ChuanqiXu9: > This definition is duplicated with "include/boost/beast/core/stream_traits.hpp:422".

Why not just using it? I tried locally and it compiles.

ChuanqiXu9:
Although there are some problems, I feel your patch is pretty easy to use after some simple changes.

mikomikotaishi:
I went ahead and applied the fixes you suggested, thanks

module/beast.cppm
mikomikotaishi: Fixed

module/beast.cppm
mikomikotaishi: Fixed

CMakeLists.txt
mikomikotaishi: Fixed

module/CMakeLists.txt
mikomikotaishi: Fixed

vinniefalco:
@ashtum can this be merged? should it?

mikomikotaishi:
@anarthal Do you still have any thoughts on integrating `import std;` with this library? As far as I'm concerned, the need to do so lies more so with building the library itself than it does with building the module, so that could be done in a follow-up PR. CMake is certainly a lot more competent with handling it now, and both GCC and Clang have great support for it (I am sure MSVC does as well, but I could not personally testify towards this).

ashtum:
> @ashtum can this be merged? should it?

This is an independent effort from the bottom-up modularization of Boost, which is currently postponed due to the lack of suitable tooling. Given the current state of the tooling, that effort may still be several releases away. Once the bottom-up work resumes, the module introduced here would be replaced, which would be a breaking change for anyone who has adopted it.

Considering the current state of the CMake files added in this PR, it won't be usable as-is by users who consume Beast through a Boost CMake release. Some additional modifications would be needed to make it work in that setup.

@mikomikotaishi, how are you currently consuming Beast as a dependency? Are you cloning the superproject with submodules, or downloading a Boost CMake release? It would be helpful to understand how this PR, in its current form, would benefit your codebase.

I think the PR could be merged once we fixed the CMake files and have tests that exercise the public API through `import boost.beast` and CI coverage that runs those tests.

mikomikotaishi:
I consume it by retrieving it through CMake's FetchContent, so yes; I am essentially cloning it. Could you point me in the direction of what would need to be done to support direct releases?

ashtum:
> I consume it by retrieving it through CMake's FetchContent, so yes; I am essentially cloning it. Could you point me in the direction of what would need to be done to support direct releases?

I would suggest just adding `beast.cppm` to your project for now and waiting until the bottom-up approach for modularizing Boost begins. Anything we add using this `export-using` approach would be experimental and could have issues that your codebase doesn't currently exercise. We should also consider that this approach would ultimately be removed from the repository.

module/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/../boost/type_traits/include>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/../boost/utility/include>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/../boost/winapi/include>
+    $<INSTALL_INTERFACE:include>
+)
+
+add_library(Boost::beast::module ALIAS boost_beast_module)
+
+# Installation
+install(TARGETS boost_beast_module
```
ashtum: We shouldn't install anything directly, that's BoostInstall job.

module/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+target_sources(boost_beast_module
+    PUBLIC
+        FILE_SET CXX_MODULES FILES
+            beast.cppm
+)
+
+target_compile_features(boost_beast_module PUBLIC cxx_std_20)
+
+target_include_directories(boost_beast_module PUBLIC
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
```
ashtum: These are unnecessary because we don't need to call install directly. The whole file can be reduced to:
```
add_library(boost_beast_module)
target_sources(boost_beast_module PUBLIC
    FILE_SET CXX_MODULES FILES beast.cppm)
target_compile_features(boost_beast_module PUBLIC cxx_std_20)
target_link_libraries(boost_beast_module PUBLIC Boost::beast)
add_library(Boost::beast::module ALIAS boost_beast_module)
```

CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ if (BOOST_BEAST_IS_ROOT)
 endif ()
 option(BOOST_BEAST_BUILD_TESTS "Build boost::beast tests" ${BUILD_TESTING})
 option(BOOST_BEAST_BUILD_FUZZERS "Build boost::beast fuzzers" OFF)
+option(BOOST_BEAST_BUILD_MODULES "Build boost::beast modules (requires C++20)" OFF)
```
ashtum:
```
option(BOOST_BEAST_EXPERIMENTAL_MODULE "Build experimental boost.beast C++20 module" OFF)
```

tools/get-boost.sh
```diff
@@ -91,5 +93,34 @@ git submodule update --init --depth 20 --jobs 4 \
 
 echo Submodule update complete
 
+# A library that is checked out while a library it names is not makes b2 fail
```
ashtum: I believe this is already fixed in `develop`, so you just need to rebase. The generic script is unnecessary, the idiomatic way to do this would be to use `boostdep`.

ashtum (commented):
After these changes your CMake file should look like:
```
set(BOOST_INCLUDE_LIBRARIES beast)
set(BOOST_BEAST_EXPERIMENTAL_MODULE ON)
add_subdirectory(boost) # or FetchContent_MakeAvailable
target_link_libraries(app PRIVATE Boost::beast::module)
```
Note: I kept `Boost::beast::module` because that's what you are currently using. It should be replaced with `Boost::beast_module` in the final version.

modules/boost_beast.cppm
```diff
@@ -0,0 +1,250 @@
+//
```
ashtum, line 1: The file should be renamed to `boost_beast.cppm` to be consistent with the rest of the libraries. The directory should also be renamed to `modules`.

anarthal:
As of today (cmake 4.4.3), import std is still experimental unfortunately. There was an attempt to make it stable in 4.3 but was rolled back due to problems (for instance, I found out a x10 compilation slowdown in Boost.Mp11). Hopefully the first stable version won't suffer from that. I agree with @ashtum proposed path.

mikomikotaishi:
Would you like me to try and rework the module's design to follow the prescribed format? In the meantime I can address all the comments that were left.

Issue #3062, opened 2025-12-12
io-context overwhelmed by small 16KB TLS reads limits high-throughput scenarios


hofst:
We use boost beast to read from S3 in high-throughput scenarios: files with hundreds of GBs on machines with 192 cores and 100Gb/s network interfaces to S3 from ec2.
The io-context is limited to ~500k operations per second and saturates at ~8 threads on our machines. More threads will just burn through CPU cycles fighting for locks.
500k operations per second should be plenty to achieve 8+GB/s throughput but it isn't: Every read_some operation only reads a single TLS fragment (max size is 16KB) even when all buffers (socket and user-space buffers) are decently sized. This leads to a limit of 3GB/sec while still beeing inefficient with the io-context busy with tiny read operations.
I tried a hacky boost modification to read more than one TLS fragment per read_some operation which promptly increased throughput by 3x and is close to the theoretical maximum of the machine. All while requiring fewer operations (and CPU utilization) in the io-context.

I don't see any way to control this without modifying boost itself - am I missing something or is this scenario currently not covered?

vinniefalco:
which version of Boost are you using?

hofst:
@vinniefalco boost 1.85 - but this has been troubling us for at least 3 years

ashtum:
> 500k operations per second should be plenty to achieve 8+GB/s throughput but it isn't: Every read_some operation only reads a single TLS fragment (max size is 16KB) even when all buffers (socket and user-space buffers) are decently sized. This leads to a limit of 3GB/sec while still beeing inefficient with the io-context busy with tiny read operations.

Using Asio's immediate executor should significantly reduce the scheduling cost, you can use [bind_immediate_executor](https://www.boost.org/doc/libs/develop/doc/html/boost_asio/reference/bind_immediate_executor.html) to bind an [inline_executor](https://www.boost.org/doc/libs/develop/doc/html/boost_asio/reference/inline_executor.html) to your handler so it just reads the available data and execute your handler immediately. To avoid stack exhaustion you should avoid binding a immediate executor every N operation (N could be like 16).

Because your load is I/O heavy you might also try running one io_context per core to see if it makes a difference. With that you would be able to pass BOOST_ASIO_CONCURRENCY_HINT_UNSAFE_IO to each io_context and reduce contention.

The tips described in this issue might be useful too: https://github.com/boostorg/beast/issues/2947

hofst:
[...]

PR #3080, opened 2026-02-06
HTTP: Increase max read buffer size to 128KB

hofst:
Increase the max read buffer size in read_size from 64KB to 128KB. This change aligns with another optimizations in Asio/SSL to reduce io_context contention and CPU utilization by allowing larger reads per operation: https://github.com/chriskohlhoff/asio/pull/1712

### DIFF
--- a/include/boost/beast/http/impl/read.hpp
+++ b/include/boost/beast/http/impl/read.hpp
@@ -206,7 +206,7 @@ public:
                 {
                     cont_ = true;
                     // VFALCO This was read_size_or_throw
-                    auto const size = read_size(b_, 65536);
+                    auto const size = read_size(b_, 128*1024);
                     if(size == 0)
                     {
                         BOOST_BEAST_ASSIGN_EC(ec, error::buffer_overflow);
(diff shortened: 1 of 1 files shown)


hofst:
@vinniefalco Opened the PR as discussed in the issue here: https://github.com/boostorg/beast/issues/3062

ashtum:
@hofst, it looks like Chris has responded to your issue with this commit: https://github.com/chriskohlhoff/asio/commit/873cb0987522ea34d93a544912683a590c44babd.

For the Beast part, you can overload [beast::read_size_helper](https://github.com/boostorg/beast/blob/7c1e061f91e2ef542217b76286c314d006c0c8fc/include/boost/beast/core/impl/read_size.hpp#L24) for beast's buffer types, or introduce a custom buffer type and `http::read_some` and `http::read` will pick up that overload, for example:

```cpp
struct tuned_buffer : beast::flat_buffer
{
    using beast::flat_buffer::flat_buffer;
};

std::size_t
read_size_helper(tuned_buffer& b, std::size_t)
{
    std::size_t const remaining = b.max_size() - b.size();
    return (std::min)(std::size_t{131072}, remaining);
}
```
Here is a complete example: https://godbolt.org/z/xTM7sY5MK.

hofst:
@ashtum That's great, thank you! Now the main part that is missing is the multi-TLS decode per task. The encode-side (upload path) has the same problem by the way and needs a multi-encode implementation to utilize large machines efficiently and fully.

ashtum:
@hofst, could `read_size_helper` help you? Is there anything else that could be done on the Beast side?

hofst:
@ashtum yes the new `read_size_helper` will be helpful and allows us to set the required read granularity, thank you!
Unfortuantely, the reamining parts are outside of boost::beast's control and need my proposed improvements in asio/ssl

PR #3086, opened 2026-03-11
Potential Vulnerability in Cloned Code

pr-hung:
## Summary
This PR fixes a potential security vulnerability in cloned code that appears to have missed an upstream security patch.

## Details
- **Affected file**: `test/bench/parser/nodejs-parser/http_parser.c`
- **Upstream fix commit**: https://github.com/nodejs/http-parser/commit/7d5c99d09f6743b055d53fc3f642746d9801479b
- **Clone similarity score**: 0.998532

## What this PR does
- Applies the upstream security fix logic to the cloned implementation in this repository.

## References
- Upstream patch commit: https://github.com/nodejs/http-parser/commit/7d5c99d09f6743b055d53fc3f642746d9801479b

Please review and merge this PR to ensure your repository is protected against this potential vulnerability. 
Thank you for your time !

### DIFF
--- a/test/bench/parser/nodejs-parser/http_parser.c
+++ b/test/bench/parser/nodejs-parser/http_parser.c
@@ -377,7 +377,10 @@ enum header_states
   , h_transfer_encoding
   , h_upgrade
 
+  , h_matching_transfer_encoding_token_start
   , h_matching_transfer_encoding_chunked
+  , h_matching_transfer_encoding_token
+
   , h_matching_connection_token_start
   , h_matching_connection_keep_alive
   , h_matching_connection_close
@@ -1384,6 +1387,7 @@ reexecute:
                 parser->header_state = h_general;
               } else if (parser->index == sizeof(TRANSFER_ENCODING)-2) {
                 parser->header_state = h_transfer_encoding;
+                parser->flags |= F_TRANSFER_ENCODING;
               }
               break;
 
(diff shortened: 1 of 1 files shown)


sehe:
This code, as the names suggest, is part of a comparative benchmark only. As such, the code was cloned at the specific time the comparison was done, and represents an "archive copy" for reproducing the results. Of course, we might remove the archive copy and replace it with a revision link to the source repository, or at least clearly mark the files as old and having security issues.

But in short, unless users willingly do *not* use Boost Beast, but instead use some benchmark code that is *not* part of Boost Beast, there is no security issue here.

Interestingly, Beast might have corresponding issues with multiple Transfer-Encodings, but that's not the topic of this issue.

ashtum:
Sorry for the late reply.

> Interestingly, Beast might have corresponding issues with multiple Transfer-Encodings, but that's not the topic of this issue.

I think there are two specific issues the Node.js commit addressed:

The presence of both `Transfer-Encoding` and `Content-Length`, which is properly handled in Beast here:
https://github.com/boostorg/beast/blob/a74967f06abe755a166de888db80c751acfb92e2/include/boost/beast/http/impl/basic_parser.ipp#L742-L744

https://github.com/boostorg/beast/blob/a74967f06abe755a166de888db80c751acfb92e2/include/boost/beast/http/impl/basic_parser.ipp#L793-L798

Multiple `Transfer-Encoding` values with `chunked` not in last position, which is also handled here:
https://github.com/boostorg/beast/blob/a74967f06abe755a166de888db80c751acfb92e2/include/boost/beast/http/impl/basic_parser.ipp#L801-L810

When `chunked` is not the last encoding, neither `flagChunked` nor `flagContentLength` is set, so Beast silently falls back to read-until-EOF mode, which is correct per RFC 7230 §3.3.3.

PR #3089, opened 2026-04-01
doc: fix json_body example entry in examples table

ssam18:
The json_body example row in the HTTP clients table was missing a link to `json_body.hpp`, which is the actual file that defines the custom body type and is the main point of the example. Also added the missing empty SSL column placeholder `[]` to match the structure of every other row in the table.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -23,3 +23,5 @@ ignore:
   - extra/**/*
   - test/*
   - test/**/*
+  - tools/*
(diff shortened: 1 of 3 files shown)


ssam18:
Can someone take a look into this PR?

ashtum:
Addressed in https://github.com/boostorg/beast/pull/3091/changes/591d05fd8dd8f882b127c2318b025895359bdcc3.

PR #3094, opened 2026-06-05
basic_parser ignores connection and framing fields in trailers

sahvx655-wq:
Noticed keep_alive() changing value after is_header_done() had already returned true whilst going through the chunked trailer path.

1. inner_parse_fields runs do_field() over the trailer section as well, so a Connection, Upgrade, Content-Length or Transfer-Encoding field sat in a chunked trailer gets applied to the message.
2. a trailer Connection: close then clears keep-alive and a trailer Upgrade sets upgrade(), both settled after the header section is finished, which rfc7230 4.1.2 does not allow.

Gated do_field() to the header section so trailer fields only reach on_trailer_field_impl. Left unfixed a peer can steer connection reuse and upgrade state from bytes that arrive behind the headers, where a front end that drops trailers would not agree. Regression added covering the keep-alive and upgrade cases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,20 +76,13 @@ jobs:
           - toolset: gcc-15
             cxxstd: 11,14,17,20,23,2c
             os: ubuntu-latest
-            container: ubuntu:25.04
+            container: ubuntu:26.04
             install: g++-15
-          - toolset: clang
-            compiler: clang++-3.5
-            cxxstd: 11
-            os: ubuntu-24.04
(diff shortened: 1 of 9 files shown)


ashtum:
@sahvx655-wq, thank you for the PR.

This change introduces a new bug. When `http::parser::merge_all_trailers` is enabled, protocol headers such as `Content-Length` and `Transfer-Encoding` can be merged with trailer headers. Previously, this condition would have triggered a parser error, but with this change it may instead be accepted and merged.

I've added a follow-up commit to address this issue.

PR #3097, opened 2026-06-13
reject requests with more than one Host header field

sahvx655-wq:
basic_parser rejects a duplicate Content-Length but quietly accepts two Host header fields, which RFC 7230 5.4 forbids because a front-end and the back-end can then resolve the request to different authorities, a routing and cache-poisoning vector.

1. do_field acts on Connection, Content-Length, Transfer-Encoding and Upgrade but never looked at Host, so a second Host was merged like any ordinary field.
2. track it with a flagHost bit and fail the request with error::multiple_host on the second occurrence, request-side only so response parsing is unchanged.

### DIFF
--- a/include/boost/beast/http/basic_parser.hpp
+++ b/include/boost/beast/http/basic_parser.hpp
@@ -110,6 +110,7 @@ class basic_parser
     static unsigned constexpr flagContentLength         = 1<< 10;
     static unsigned constexpr flagChunked               = 1<< 11;
     static unsigned constexpr flagUpgrade               = 1<< 12;
+    static unsigned constexpr flagHost                  = 1<< 13;
 
     static constexpr
     std::uint64_t
--- a/include/boost/beast/http/error.hpp
+++ b/include/boost/beast/http/error.hpp
@@ -170,7 +170,10 @@ enum class error
     header_field_name_too_large,
 
     /// Header field value exceeds @ref basic_fields::max_value_size.
-    header_field_value_too_large
+    header_field_value_too_large,
+
+    /// The request contains more than one Host header field.
+    multiple_host
 };
 
 } // http
--- a/include/boost/beast/http/impl/basic_parser.ipp
+++ b/include/boost/beast/http/impl/basic_parser.ipp
@@ -835,6 +835,21 @@ do_field(field f,
         return;
     }
 
+    // Host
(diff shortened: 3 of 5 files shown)


ashtum:
@sahvx655-wq, thanks for the great work here.

My concern is that this probably doesn't belong in `basic_parser` and arguably not in Beast at all. It's a request-routing concern rather than message framing, and Beast deliberately owns framing while leaving HTTP semantics and policy to the user. Even if we merged it, it would be an incomplete, inconsistent slice of RFC 7230 §5.4: of the three Host requirements (missing, duplicate, and invalid), the parser can only enforce the duplicate case the other two need method/version context and URI-authority validation.

sahvx655-wq:
Fair point, and I'll grant the §5.4 framing in the commit message oversells what this actually does. I put it in `do_field` because it sits a few lines below the existing `multiple_content_length` rejection and follows the same shape: one request carrying two values for a field whose duplication lets a front-end and Beast resolve the message to different authorities. That's the framing-ambiguity class the parser already owns rather than a routing policy a user layers on top, which is why it felt at home next to the Content-Length and Transfer-Encoding handling.

You're right about the incomplete slice though. Missing Host could be derived from the version context, but invalid Host means comparing the request-target authority against the field value, which is real URI work Beast doesn't do at this layer, so enforcing one of three would read as a half-measure. I scoped it to the duplicate case on purpose rather than as a stab at full §5.4, since that's the one that's pure framing ambiguity. If you'd still rather keep Host semantics out of the parser entirely I'm happy to drop it, your call.

ashtum:
> You're right about the incomplete slice though. Missing Host could be derived from the version context, but invalid Host means comparing the request-target authority against the field value, which is real URI work Beast doesn't do at this layer, so enforcing one of three would read as a half-measure. I scoped it to the duplicate case on purpose rather than as a stab at full §5.4, since that's the one that's pure framing ambiguity. If you'd still rather keep Host semantics out of the parser entirely I'm happy to drop it, your call.

I believe Beast's parser responsibilities should remain clear. Delegating Host validation entirely to the user is preferable to validating it partially within the parser.

sahvx655-wq:
Makes sense, I'd rather the parser's remit stay unambiguous than half-enforce §5.4. I'll leave Host policy to the application layer and close this out. Thanks for the look.

PR #3105, opened 2026-07-03
reject CR and LF in basic_fields field name and value

sahvx655-wq:
basic_fields::try_create_new_element assembles the on-wire field as name, ": ", value, then a terminating CRLF, but it only checks the name and value lengths, never their contents. A CR or LF sitting inside a value is copied straight into that buffer, so a header built from attacker-influenced data such as a redirect target or a reflected parameter splits the field and starts new header lines. Setting Location to "/next\r\nSet-Cookie: sid=attacker" and serialising the response puts Set-Cookie on its own line, which is classic response splitting.

The fix rejects a name or value containing CR or LF in try_create_new_element, alongside the existing size checks, reusing the bad_field and bad_value codes the parser already returns for those octets when reading. All of the set and insert overloads, including the error_code variants, route through this one function, so the check covers them together and stops the malformed field at the point it is created instead of relying on every caller to pre-scan its input.

### DIFF
--- a/include/boost/beast/http/impl/fields.hpp
+++ b/include/boost/beast/http/impl/fields.hpp
@@ -981,6 +981,20 @@ try_create_new_element(
         BOOST_BEAST_ASSIGN_EC(ec, error::header_field_value_too_large);
         return nullptr;
     }
+    // A CR or LF in the name or value is written verbatim into the
+    // serialized header block, which is terminated by CRLF; either one
+    // splits the field and injects additional header lines (rfc7230
+    // 3.2 forbids them in field-name and field-value).
+    if(sname.find_first_of("\r\n") != string_view::npos)
+    {
+        BOOST_BEAST_ASSIGN_EC(ec, error::bad_field);
+        return nullptr;
+    }
+    if(value.find_first_of("\r\n") != string_view::npos)
+    {
+        BOOST_BEAST_ASSIGN_EC(ec, error::bad_value);
+        return nullptr;
+    }
     value = detail::trim(value);
     std::uint16_t const off =
         static_cast<off_t>(sname.size() + 2);
--- a/include/boost/beast/http/impl/fields.hpp
+++ b/include/boost/beast/http/impl/fields.hpp
@@ -775,6 +775,17 @@ keep_alive_impl(
     beast::detail::temporary_buffer& s, string_view value,
     unsigned version, bool keep_alive);
 
+// Two single-octet finds lower to memchr, which is
+// considerably faster than find_first_of("\r\n").
+inline
+bool
+contains_crlf(string_view s)
+{
+    return
+        s.find('\r') != string_view::npos ||
+        s.find('\n') != string_view::npos;
+}
+
 } // detail
 
 //------------------------------------------------------------------------------
@@ -862,6 +873,11 @@ void
 basic_fields<Allocator>::
 set_method_impl(string_view s)
 {
+    // A CR or LF here splits the serialized start
+    // line, just as it does in a field (rfc7230 3.1.1).
+    if(detail::contains_crlf(s))
+        BOOST_THROW_EXCEPTION(
+            system_error{error::bad_method});
     realloc_string(method_, s);
 }
 
@@ -871,6 +887,9 @@ void
 basic_fields<Allocator>::
 set_target_impl(string_view s)
 {
+    if(detail::contains_crlf(s))
+        BOOST_THROW_EXCEPTION(
+            system_error{error::bad_target});
     realloc_target(
         target_or_reason_, s);
 }
@@ -881,6 +900,9 @@ void
 basic_fields<Allocator>::
 set_reason_impl(string_view s)
 {
+    if(detail::contains_crlf(s))
+        BOOST_THROW_EXCEPTION(
+            system_error{error::bad_reason});
     realloc_string(
         target_or_reason_, s);
 }
@@ -985,12 +1007,12 @@ try_create_new_element(
     // serialized header block, which is terminated by CRLF; either one
     // splits the field and injects additional header lines (rfc7230
     // 3.2 forbids them in field-name and field-value).
-    if(sname.find_first_of("\r\n") != string_view::npos)
+    if(detail::contains_crlf(sname))
     {
         BOOST_BEAST_ASSIGN_EC(ec, error::bad_field);
         return nullptr;
     }
-    if(value.find_first_of("\r\n") != string_view::npos)
+    if(detail::contains_crlf(value))
     {
         BOOST_BEAST_ASSIGN_EC(ec, error::bad_value);
         return nullptr;
--- a/test/beast/http/fields.cpp
+++ b/test/beast/http/fields.cpp
@@ -477,6 +477,26 @@ public:
             BEAST_THROWS(f.set("", big_value),                boost::system::system_error);
         }
 
+        // reject CR or LF in a field name or value (rfc7230 3.2)
+        {
+            fields f;
+            error_code ec;
+
+            f.insert(field::age, "a\r\nb", "1", ec);
+            BEAST_EXPECT(ec == error::bad_field);
+            f.insert(field::age, "age", "1\r\nInjected: x", ec);
+            BEAST_EXPECT(ec == error::bad_value);
+            f.insert(field::age, "age", "1\ntwo", ec);
+            BEAST_EXPECT(ec == error::bad_value);
+            f.insert(field::age, "age", "1\rtwo", ec);
+            BEAST_EXPECT(ec == error::bad_value);
+
+            BEAST_THROWS(f.set(field::location, "/\r\nSet-Cookie: x"),
+                boost::system::system_error);
+            BEAST_THROWS(f.set("X\r\nY", "1"),  boost::system::system_error);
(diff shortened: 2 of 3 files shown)


ashtum:
Hi, thanks for the PR. A few caveats worth flagging:

- Given the layer Boost.Beast's serializer operates at, checks like this are usually left to a higher layer, unless doing them here proves cheaper than delegating.
- The request method, request target, and response reason-phrase aren't covered by this check and likely need the same sanitization.
- Checking for CR/LF fixes the vulnerability, but we could also validate names and values against their allowed character sets, which would cover this case as well, if it proves to be worthwhile.

Separately, `core::string_view::find_first_of` is oddly slow for this task. Even a naive form like:

```cpp
    if( sname.find('\r') != string_view::npos ||
        sname.find('\n') != string_view::npos)
    {
        BOOST_BEAST_ASSIGN_EC(ec, error::bad_field);
        return nullptr;
    }
    if( value.find('\r') != string_view::npos ||
        value.find('\n') != string_view::npos)
    {
        BOOST_BEAST_ASSIGN_EC(ec, error::bad_value);
        return nullptr;
    }
```

is ~3x faster for ~30 char inputs and ~10x faster around ~300 chars. Note: I'm not suggesting this exact version, just that `core::string_view::find_first_of` looks like a poor fit here and warrants a closer look.

sahvx655-wq:
Good points, took a proper look at each. On the scan: I benchmarked before touching anything, and on Apple clang with core::string_view two single-octet find() calls measure about 1.4x faster than find_first_of("\r\n") at 30 chars and roughly 5x at 300, with find_first_of only winning marginally below ~10 chars. The root cause is that find_first_of walks the input octet by octet testing membership in the needle set, while find(char) forwards to memchr and gets vectorised. I've switched to the two-find form through a small detail::contains_crlf helper and extended the same rejection to set_method_impl, set_target_impl and set_reason_impl, reusing the parser's bad_method, bad_target and bad_reason codes, with tests in message.cpp. The read path is unaffected since basic_parser never passes those strings through with a bare CR or LF; the only new cost there is one memchr pass per start line. One quirk I noticed while testing: header::method_string assigns its cached verb before calling set_method_impl, so a rejected method leaves the cached verb as unknown while the stored bytes stay intact. I left that ordering alone to keep the diff contained.

On layering, the case for doing it here is that try_create_new_element is the one choke point every set and insert overload funnels through, so the check is written once and costs two memchr passes over bytes the function is about to copy anyway. Delegating means every call site that touches request-derived data has to remember to pre-scan, and a single miss puts a split header on the wire, which is the failure mode that worries me. Full charset validation would cover this too, but it needs a per-octet table lookup rather than memchr, and its main additional effect beyond CR/LF is rejecting stray C0 controls, which are ugly but don't alter framing on a correct peer; CR and LF are the two octets that actually change message structure. If you think the table-driven version is worth that cost I'm fine doing it instead.

sahvx655-wq:
any update?

ashtum:
> Any update?

Regarding the code, I think we could probably do a bit better than:

```cpp
   s.find('\r') != string_view::npos ||
   s.find('\n') != string_view::npos;
```

Especially since there’s an opportunity here to validate names and values against their complete allowed character sets.

That said, I think this is the less important part. It would be good to first settle the questions above:

> * Given the layer Boost.Beast's serializer operates at, checks like this are usually left to a higher layer, unless doing them here proves cheaper than delegating.
> * The request method, request target, and response reason-phrase aren't covered by this check and likely need the same sanitization.
> * Checking for CR/LF fixes the vulnerability, but we could also validate names and values against their allowed character sets, which would cover this case as well, if it proves to be worthwhile.

If we decide to perform the validation, there are a few options:

* If we defer the validation to serialization time, a simple configuration option in the serializer could allow users to skip the validation when they don’t need it, so they don’t pay the cost unnecessarily.
* If we add the validation to the container, we would need an _unchecked variant that skips the validation. This could be used by users who want to avoid the validation cost, and also by the parser itself when inserting parsed field names and values, since the parser has already validated them.
* There’s also the possibility of adding concrete view types for field names and values that guarantee the referenced string is valid. These could then be passed around without incurring validation overhead at each boundary. (Similar to [boost::urls::pct_string_view](https://www.boost.org/doc/libs/latest/doc/antora/url/reference/boost/urls/pct_string_view.html)).

The following is a table of popular HTTP libraries and frameworks, showing whether they perform validation and, if so, whether it happens at the container level or at serialization time (generated using Claude Code):

```text
Library                                  Name     Value    Where
---------------------------------------- -------- -------- -------------
C / C++
  libcurl                                none     none     -
  cpp-httplib                            full s   full s   container
  libevent                               crlf     crlf     container
  libsoup                                crlf s   crlf s   container
  libmicrohttpd                          crlf     crlf     container
  nghttp2 (outbound)                     none     none     -
  POCO                                   none     none     -
  Qt QHttpHeaders (>=6.7)                full     full     container
  Proxygen HTTPHeaders (egress)          none?    none?    -
Go
  net/http client                        full     full     serialization
  net/http server                        full s   crlf s   serialization
  fasthttp                               crlf s   crlf s   container
Rust
  http (hyper/reqwest/axum/actix)        full     full     type
JS
  Node http                              full     full     container
  Fetch Headers (browsers/undici/Deno)   full     crlf     container
Python
  http.client / urllib3~                 crlf     crlf     serialization
  requests                               crlf     crlf     serialization
  httpx (~h11)                           full~    full~    serialization
  h11                                    full     full     container
  aiohttp                                crlf     crlf     serialization
  Tornado                                crlf     crlf     serialization
  Twisted                                crlf s   crlf s   container
  Django                                 crlf     crlf     container
  Werkzeug / Flask                       none     crlf     container
  Starlette / FastAPI (~ASGI server)     ~        ~        ~
  uvicorn (httptools)                    full     full!    serialization
  gunicorn                               full     full     serialization
Java / Kotlin
  Netty                                  full     full     container
  OkHttp                                 full     full     container
  JDK java.net.http                      full     full     container
  JDK HttpURLConnection                  crlf!    crlf!    container
  JDK com.sun.net.httpserver             crlf     crlf     container
  Apache HttpComponents 5                none     crlf s   serialization
  Jetty                                  crlf s   full s   serialization
  Tomcat                                 crlf s   full s   serialization
  Ktor                                   full     full     container
  Spring HttpHeaders (~container)        ~        ~        ~
.NET
  HttpClient HttpHeaders.Add             full     crlf     container
    TryAddWithoutValidation              full     none     container
  WebHeaderCollection                    full     full     container
  Kestrel                                full     full     container
Other
  Ruby Net::HTTP                         none     crlf     container
  Puma                                   full s   full! s  serialization
  PHP header() (Symfony/Laravel~)        crlf     crlf     container
  guzzle/psr7 (PSR-7)                    full     full     container
  Elixir Plug                            none     crlf     container
  Elixir Mint                            full     full     serialization
  SwiftNIO                               full     full     serialization
  Dart dart:io                           full     full     container

full = grammar: name token (or close approximation), value no CTL
       except HT (obs-text handling varies)
crlf = CR/LF caught (some also NUL, ':' or leading WS), nothing else
none = no check
s    = sanitized or silently dropped instead of rejected
!    = bare CR passes (only LF caught)
~    = no check of its own; inherits from the layer in parentheses
?    = not verified
```

Beast is neither a client nor a server library, so there are other vulnerabilities on both the client and server sides that it doesn’t, and can’t, prevent. This wouldn’t be the only case where the higher-level application needs to take responsibility for the appropriate validation.

Given that, I believe delegating this validation to the higher level would be a defensible position.

sahvx655-wq:
I measured the container check instead of arguing it, and the numbers undercut what I said earlier. My claim that the read path was unaffected was wrong: `parser::on_field_impl` inserts every parsed field through `try_create_new_element`, so each one is rescanned after `basic_parser` has already validated it. On an M4 with Apple clang at -O2, parsing a 14-field browser-style GET into `request_parser<empty_body>` goes from about 850 ns to about 980 ns per message on this branch, roughly 15%, and building an 8-field response goes from about 208 ns to 260 ns. That works out to a fixed 7 to 9 ns per field on short strings, and it didn't move when I swapped the memchr pair for an inline loop. I also tried the full grammar in a scratch tree, a token bitmap for the name and a branch-free no-CTL-except-HT loop for the value, checked octet by octet against `is_token_char` and `is_text`, and it landed within noise of the CR/LF version (about 970 ns and 268 ns). So you're right on that point too: if validation goes in at all it should be the complete character sets, because stopping at CR/LF saves nothing.

Given that, of your three options I'd pick the container check with an unchecked insert for the parser. It puts the parse path back at baseline by construction, the error still surfaces at the `set()` or `insert()` call that supplied the bad octets instead of later inside `serializer::next`, and since `parser<>` is already written against `basic_fields<Allocator>` the unchecked entry point doesn't have to touch the Fields concept. Serialisation-time validation rescans relayed messages unless the user knows to switch it off, and the view types are a bigger API change than this PR should carry. The layering call is yours though, and I won't push on it. The risk in delegating is that most of the libraries in your table do catch this, so people arriving from those stacks will assume Beast does as well, and one unsanitised Location or Set-Cookie value becomes a split response. If you'd rather leave it to the application, I can cut this down to a note in the `basic_fields` docs saying names and values go to the wire unvalidated, or close it. I'll hold off changing the branch until you've picked a direction.

vinniefalco:
Curious to what this means for Boost.Http

PR #3108, opened 2026-08-05
Fix the template parameter name in buffers_iterator_type

karpovantonme:
One `@tparam` naming something the template does not take.

```cpp
/** Type alias for the iterator type of a buffer sequence type.
    ...
    @tparam T The buffer sequence type to use.
*/
template <class BufferSequence>
using buffers_iterator_type = __see_below__;
```

The three other `@tparam` tags in this file, on `buffers_type` and its neighbours, all say `BufferSequence`, so this is the odd one out rather than a convention I am changing.

Not touched, in case you want it: `http/message.hpp:284` documents `@param args` on a constructor taking `Arg1&& arg1, ArgN&&... argn`. Describing a pack collectively reads fine to me, and `clang -Wdocumentation` disagrees, so I left it as your call.


ashtum (approved):

karpovantonme:
The two windows clang-win failures look unrelated to a comment-only change, but say the word if you want me to look

PR #3111, opened 2026-09-18
reset stale params in ext_list::const_iterator::increment

sahvx655-wq:
ext_list::const_iterator::increment clears v_.first on each step but leaves v_.second alone, and it returns early when the extension token runs to the end of the string, before the param_list is rebuilt. I found it while tracing how pmd_read walks Sec-WebSocket-Extensions: printing the elements of `ext_list{"a;x=1,b"}` shows b carrying x=1, the parameters of the element before it. Through pmd_read, `x-foo; server_max_window_bits=9, permessage-deflate` is read as a permessage-deflate offer with server_max_window_bits=9, and `x-foo; mode=1, permessage-deflate` gets the offer declined over a parameter that belongs to another extension.

The root cause is stale iterator state, so the reset sits at the top of increment next to the v_.first one and every user of ext_list gets the right answer without special-casing the last element. Left as is, code that decides per extension from a peer-supplied header sees parameters the peer attached to something else, and a websocket endpoint can settle on compression parameters that were never sent for permessage-deflate. The new testExtList cases fail on develop (`'a;i=1,b;i=1'` at rfc7230.cpp(127)) and pass with the change; the http and websocket suites run clean locally.


vinniefalco:
Very good find

